PR #129, opened 2023-03-21
refactor: delete `transferprocess-sftp` as it was not used

paullatzelsperger:
## WHAT
deletes the `transferprocess-sftp-*` modules

## WHY 
they were not used

## Further notes
- added a retry feature to the `helm install` script of the business tests


paullatzelsperger:
builds are broken due to https://github.com/eclipse-edc/GradlePlugins/pull/106

florianrusch-zf:
Do you know the reason why this was implemented in the past?

paullatzelsperger:
> Do you know the reason why this was implemented in the past?

I don't, but I assume it was a first implementation to have _something_ usable, without the need to provision an S3 bucket or an Azure Blobstore

florianrusch-zf:
**Before we merge this**: I would like to speak within the team about it. I would like to understand, why this was added in the past

paullatzelsperger:
> **Before we merge this**: I would like to speak within the team about it. I would like to understand, why this was added in the past

I found no usages of it within the project, and according to @stefan-ettl no outside party is using it either. so we can go ahead with merging this.

bcronin90:
> Do you know the reason why this was implemented in the past?

Support for SFTP data transfer was a requirement in the past. If it's now a requirement not to support it, go ahead and delete it. I would however ask, what our goals are here. They seem to be rather volatile.

florianrusch-zf:
As it was never used in the control-/data-plane I'm fine with dropping it.

paullatzelsperger:
> > Do you know the reason why this was implemented in the past?
> 
> Support for SFTP data transfer was a requirement in the past. If it's now a requirement not to support it, go ahead and delete it. I would however ask, what our goals are here. They seem to be rather volatile.

The goal I'm pursuing here are to not maintain unused/dead code.

bcronin90:
> > > Do you know the reason why this was implemented in the past?
> > 
> > 
> > Support for SFTP data transfer was a requirement in the past. If it's now a requirement not to support it, go ahead and delete it. I would however ask, what our goals are here. They seem to be rather volatile.
> 
> The goal I'm pursuing here are to not maintain unused/dead code.

I had gathered as much. I am referring to requirements from Catena-side. We had a task to create this extension. Thus we created it. If it's now supposed to be removed, that's a shame, but fine by me. But only if the initial requirement no longer exists. I assume this feature was requested with some reason behind it.

Maybe someone with a better perspective in the project like @stefan-ettl can comment on this? What's the plan here? Do we need this or do we not?

paullatzelsperger:
Update from @stefan-ettl: apparently the TransferProcess-SFTP module(s) are still needed by a stakeholder, so I'll remove the deletion from the PR, and update the description accordingly.

PR #143, opened 2023-03-23
docs: add decision record about removing Lombok

paullatzelsperger:
## WHAT
creates a decision record about completely excising Lombok from our code base

## WHY
Lombok uses undocumented and non-public APIs of the JVM and is expected to break in upcoming Java releases.

### DIFF
--- a/docs/development/decision-records/2023-03-23_remove_lombok/README.md
+++ b/docs/development/decision-records/2023-03-23_remove_lombok/README.md
@@ -24,7 +24,7 @@ refactoring (`@ToString(of = <FIELDNAME>)`) and more dangerous (`@SneakyThrows`)
 Fourth and finally bytecode modification could conceivably cause problems in use cases where audited/certified code is
 required. Since the code gets modified during compilation in a way not covered by any spec, technically the runtime code
 could be significantly different from the source code. Although this problem is admittedly theoretical at the moment, we
-should build those obstructions into the code base.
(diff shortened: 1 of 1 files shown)


ndr-brt (approved):
Totally agree, I took a glance at the issue on the lombok repository and I think all this noise just to save some constructors and methods seems to much.
:axe:

tuncaytunc-zf:
To be fair, we should also mention some advantages of Lombok
- Avoid boilerplate code
- Save time
- Clean code
- Open Source with huge community behind
- ...
I have used it also in many other projects can't say that it is a bad tool.

wolf4ood (approved):
:axe:

florianrusch-zf (approved):

SebastianBezold:
Hi @tuncaytunc-zf, 
reading your comment, I cannot really see, if you're against this change. I want to avoid merging PRs, if the team is not aligned. Can you please clarify, if you do support the decision beside your topics from the comment?

tuncaytunc-zf:
> Hi @tuncaytunc-zf, reading your comment, I cannot really see, if you're against this change. I want to avoid merging PRs, if the team is not aligned. Can you please clarify, if you do support the decision beside your topics from the comment?

Hi @SebastianBezold I just wanted to point out that it has also some advantages, but indeed the points in D-R are some potential risks and if most of us are against Lombok then of course it can be removed. So please merge it.

PR #147, opened 2023-03-27
docs: add concept for encrypted transfers

bcronin90:
As per https://jira.catena-x.net/browse/A1IDSC-390 a way to easily transfer encrypted data is required. The requirement specifically goes beyond SSL/TLS during transfer. Rather, assets are to be stored with encryption and access to decryption should be granted to anyone legitimately accessing it.

As per @stefan-ettl @stephanbcbauer a decision record is needed. This is it.

<sub>Brendan Cronin [brendan.cronin@mercedes-benz.com](mailto:brendan.cronin@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>


florianrusch-zf:
Is this really something we need in tractusx-edc, or wouldn't this fit better in upstream?

bcronin90:
> Is this really something we need in tractusx-edc, or wouldn't this fit better in upstream?

@stefan-ettl @stephanbcbauer Should this be moved?

jimmarino:
I don't think this belongs upstream. It is a solution for transferring encrypted data specific to a dataspace. For this to be adopted upstream (since it affects the data exchange protocol), it would need to be included in an IDS protocol specification.

jimmarino:
> I just thought, that this could be beneficial also for other dataspaces not only Catena-X.

It could be, but it should be part of the IDS specs since it affects the protocol. There is a separate issue of how best to achieve encrypted data transfer. I don't think the Contract Agreement or Offer is the correct place to do this, but we can discuss that offline.

florianrusch-zf:
@bcronin90 @jimmarino what is the current state?

jimmarino:
@florianrusch-zf, my bad for not responding earlier. The requirements unclear to me. Specifically:

1.  Why does the provider need to encrypt the data with a private key and then share that private key with one or more clients? That seems to violate the requirement that "Private keys need to remain private"

2. I'm not sure what "The same data can be shared with multiple recipients in a resource-conserving way" entails.

3. Where and when does the data need to be encrypted, e.g. in transit, at rest, both, etc.?

We should o back and clarify those points. Encryption keys should not be part of the contract agreement. As a rough sketch, I would take a different approach:

1. Client and provider negotiate a contract agreement. The offer contains a usage constraint that specifies the encryption requirements.
2. If the data needs to be encrypted in transit beyond what transport level encryption provides, the client should provide a public key in the data address part of the transfer process request.
3.  If the data needs to be encrypted in transit beyond what transport level encryption provides, the provider will use the public key to provision an encrypted version of the data during its provision phase and either sends the encrypted data to the client or notify the client that the data is available
4. The client obtains the data and uses the associated private key to decrypt it.

Note that this approach would not solve the use case where a provider only wants to make data available to a client for a certain period by supplying some type of expiring key.

bcronin90:
Alright, I'm back. Let's get into this.

> @florianrusch-zf, my bad for not responding earlier. The requirements are unclear to me. Specifically:
> 
>     1. Why does the provider need to encrypt the data with a private key and then share that private key with one or more clients? That seems to violate the requirement that "Private keys need to remain private"
I think I need to clarify this. The idea is to encrypt the actual data with a keypair that the recipient generates. The recipient then sends the private key of that keypair to the sender and encrypts that message with the public key of the sender. Basically this flow here: https://en.wikipedia.org/wiki/Pretty_Good_Privacy#/media/File:PGP_diagram.svg
The only private key that is transferred is the one called "random key" in that diagram. I assume that won't be a security issue, right?

>     2. I'm not sure what "The same data can be shared with multiple recipients in a resource-conserving way" entails.
Good point. We had an earlier concept that was rather inefficient. Since this started on JIRA, you can't know that. I'll rephrase this requirement.

>     3. Where and when does the data need to be encrypted, e.g. in transit, at rest, both, etc.?
Both. I know we already encrypt transfers, but there is a (business) requirement to go beyond that. So the next logical step to me was to create a proposal to implement PGP. That way we encrypt the relevant data before a transfer and it remains encrypted at the destination.

> We should go back and clarify those points. Encryption keys should not be part of the contract agreement. As a rough sketch, I would take a different approach:
> 
>     1. Client and provider negotiate a contract agreement. The offer contains a usage constraint that specifies the encryption requirements.
> 
>     2. If the data needs to be encrypted in transit beyond what transport level encryption provides, the client should provide a public key in the data address part of the transfer process request.
> 
>     3. If the data needs to be encrypted in transit beyond what transport level encryption provides, the provider will use the public key to provision an encrypted version of the data during its provision phase and either sends the encrypted data to the client or notifies the client that the data is available
> 
>     4. The client obtains the data and uses the associated private key to decrypt it.
> 
> Note that this approach would not solve the use case where a provider only wants to make data available to a client for a certain period by supplying some type of expiring key.
I agree. I don't think expiration of access is a requirement here, and even than I don't think we could solve it here. Pinning the transferred key to the Data Request makes more sense, I'll adapt that.

So really, we just have this one disagreement, the use of a random key to perform the actual encryption with. And I think it boils down to this:
Pro: Allows for validation of sender (since only the sender can have the encryption key)
Con: More overhead and more keys to keep in vault

So my remaining question is: Is that tradeoff worth it?

jimmarino:
As a first step, I think we need to go back and refine the requirements as they are unclear. Once this is done, we can produce a detailed solution write-up.

For example, the requirements should be similar to the following:

A provider must be able to publish an asset (to the dataspace) with a usage requirement that the asset data will be encrypted in transit (in addition to transport level encryption) and must remain encrypted on the consumer side when the data is at rest. 

Once we have the requirements clearly stated, we can propose a solution. Taking the above example as a baseline, we create a solution similar to the following:

1. A written specification for the encryption usage policy, including an example of the ODRL policy serialization
2. A written specification for transmitting a public key for decryption using the new Datasapce  Protocol messages. This should be based on standards where possible and include a schema using a schema language such as JSON Schema.
 
More details of a proposal include:

1. Private keys should not be transmitted (only public ones should be)
2. Encryption keys should not be shared across consumers
3. The provider should perform encryption for the consumer during the provisioning phase
4. Key transmission should be based on standards, e.g., JWK/JWKS, which allow for different algorithms (as opposed to PGP)

Let's discuss this.

bcronin90:
Thanks for the feedback.I think it would make sense for both parties to be able to require encryption as part of their contract. I'll flesh things out a bit based on that.

JWK seems like a good fit for the key transfer format, thanks for the suggestion.






---- Am Di, 25 Apr 2023 10:40:08 +0200 schrieb Jim Marino ***@***.***> ----





As a first step, I think we need to go back and refine the requirements as they are unclear. Once this is done, we can produce a detailed solution write-up.

For example, the requirements should be similar to the following:

A provider must be able to publish an asset (to the dataspace) with a usage requirement that the asset data will be encrypted in transit (in addition to transport level encryption) and must remain encrypted on the consumer side when the data is at rest.

Once we have the requirements clearly stated, we can propose a solution. Taking the above example as a baseline, we create a solution similar to the following:

A written specification for the encryption usage policy, including an example of the ODRL policy serialization

A written specification for transmitting a public key for decryption using the new Datasapce  Protocol messages. This should be based on standards where possible and include a schema using a schema language such as JSON Schema.


More details of a proposal include:

Private keys should not be transmitted (only public ones should be)

Encryption keys should not be shared across consumers

The provider should perform encryption for the consumer during the provisioning phase

Key transmission should be based on standards, e.g., JWK/JWKS, which allow for different algorithms (as opposed to PGP)


Let's discuss this.

bcronin90:
Any reviews?

florianrusch-zf:
I still have concerns about having this implemented in this project 😕

If we implement this only in tractusx-edc, that would mean our Catena-X EDCs are not compatible with other variants of the upstream Connector repo. Means: no interoperability with other Dataspaces. (For sure, no one checked already, if we are already interoperable - but we shouldn't increase the possible already existing gab) @stefan-ettl what do you think about this point?

jimmarino:
> I still have concerns about having this implemented in this project 😕
> 
> If we implement this only in tractusx-edc, that would mean our Catena-X EDCs are not compatible with other variants of the upstream Connector repo. Means: no interoperability with other Dataspaces. (For sure, no one checked already, if we are already interoperable - but we shouldn't increase the possible already existing gab) @stefan-ettl what do you think about this point?

This is a great start, but I still think this DR needs more detail because it doesn't outline enough of *how* encryption should be architected. Let's discuss that in a separate thread.

This DR should not be upstreamed because it is a specific solution rather than a platform feature. The distinction here is that this DR should cover how to implement a business requirement using features of the EDC rather than providing a set of general features that can address many different business requirements. Specifically, this DR should focus on what formats keys are transmitted, how they are transmitted via DSP messages, and what extension points should be used to perform encryption (e.g., the provisioner). EDC should instead focus on providing a message protocol (DSP) and extension points that invoke user code that implements business requirements. 

Keeping this DR in C-X *does not* mean C-X will not be interoperable with upstream EDC. In a dataspace, there are several aspects to interop:

1.  The wire-level protocol and message flow 
2. Trust model, including trust anchors
3. Policy semantics (not technical implementation)   

EDC handles #1 with its support for DSP. The trust model involves the type of trust system used and the flows it requires, e.g. OAuth2 or some form of SSI . EDC provides extension points and partial implementations of them. However, it leaves the decision on trust anchors to individual dataspaces. Item 3 is left entirely to individual dataspaces since policies are typically tied to specific domains.

All three aspects are required to have dataspace interoperability.

Also, moving extensions to upstream EDC does not impact interoperability. They must still be included in a runtime distribution (or substitutes used). This is easily done using Java dependency resolution tools (Maven, Gradle).

Regarding "inter-dataspace" interoperability, a useful way to frame this (which we did in the DSP working group) is to say: "There is no data exchange outside a dataspace." In other words, for two parties to exchange data in DSP, there must be a dataspace. If two participants from different dataspaces want to exchange data, they must create a third dataspace. This is fine since dataspaces can be lightweight contexts. Once that is done, we're back to the three interop layers above. Turtles all the way down.

bcronin90:
I feel like I'm just blindly guessing here. What kind of detail would you like to see? What are the open questions?

jimmarino:
> I feel like I'm just blindly guessing here. What kind of detail would you like to see? What are the open questions?

I think this is a great start. Give me a few days since I'm knee-deep in a release. Sorry for the delay.

bcronin90:
@stephanbcbauer I added you as reviewer because Jim is occupied and I had hoped you'd be familiar with the requirements. It's been open for so long, we can either merge or delete it at this point.

stephanbcbauer:
> @stephanbcbauer I added you as reviewer because Jim is occupied and I had hoped you'd be familiar with the requirements. It's been open for so long, we can either merge or delete it at this point.

@bcronin90 sorry, I have no idea about these requirements :(

jimmarino:
Does anyone have a written copy of the requirements? Maybe Stefan knows? If there are no requirements or no one is willing to write them down for us, maybe that is an indication this is not a priority?

stephanbcbauer:
@stefan-ettl we need a set of requirements for this enabler. Who is the requester for this? Thx

PR #172, opened 2023-03-30
fix: dockerization

paullatzelsperger:
## WHAT

This PR adds a condition to dockerization to only perform the task if a `Dockerfile` exists.

## WHY
avoid build failures

## ALSO
it adds a comment in the build file for people on Apple M1 (such as me).


tuncaytunc-zf:
I think this might solve the problem why I get a lot of dangling images :-)

PR #176, opened 2023-04-01
fix: remove private backend service

paullatzelsperger:
## WHAT

This PR removes the private service image, that was used until now, from the business-test helm charts. In the business tests it was replaced with a Mock.

It also replaces/amends the corresponding Cucumber test with an equivalent JVM-based test.

## WHY

We should not have code in an Eclipse project, that is not covered by either Eclipse, or an appropriate license, or that we have no control over.

Also, the corresponding Cucumber test (`HttpProxyTransferSteps`) didn't really test anything of value, so going forward, all test maintenance must happen in the `HttpConsumerPullWithProxyTest`. 

_Note: a similar test already exists in upstream EDC, but was replicated here for the sake of not losing test coverage_.

## FURTHER NOTES

This new test verifies, that access to a private backend can be granted through the data plane's HTTP proxy feature, where the consumer pulls the data.

Specifically:
- two connectors, sokrates (consumer) and plato (provider)
- plato offers data through an asset, which is a private backend (simulated by `mockwebserver`)
- sokrates requests access to that asset
- the test executes contract negotiation, data transfer and the request chain through both data planes to extract the data.


edc-tests/cucumber/src/test/java/org/eclipse/tractusx/edc/tests/HttpProxyTransferSteps.java
wolf4ood, line 1: Does it make sense to remove it at this point?

paullatzelsperger: it does - done!

wolf4ood (approved):

stephanbcbauer (approved):
LGTM @SebastianBezold could you please merge this PR? Thx

Issue #167, opened 2023-03-29
0.3.0 Charts to not reflect setting changes of the 0.3.0 image


DominikPinsel:
## Describe the bug

As it is now the charts to not reflect the changes from the migration guide.
https://github.com/catenax-ng/tx-tractusx-edc/blob/develop/docs/migration/Version_0.1.x_0.3.x.md

The affected charts are currently from the product-edc repository. But the idea would be to fix them here.

Affected charts:
```
catenax-ng-product-edc/edc-controlplane --version 0.3.0
catenax-ng-product-edc/edc-dataplane --version 0.3.0
catenax-ng-product-edc/tractusx-connector --version 0.3.0
```

## Context Informations

- Used version: 0.3.0

<sub>Dominik Pinsel <dominik.pinsel@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md) </sub>

florianrusch-zf:
Are all points not taken into account? Or only a few? If only a few, can we collect them here? That would make the work on the fix easier in the end.

DominikPinsel:
> Are all points not taken into account? Or only a few? If only a few, can we collect them here? That would make the work on the fix easier in the end.

I'm not 100% sure which changes are causing issues in which charts.

I've a feeling the `web.http.data` to `web.http.management` change might cause issues in the Control- & DataPlane Chart. And the `edc.receiver.http.endpoint` to `edc.receiver.http.dynamic.endpoint` might cause issues in the TXDC chart. But haven't tested it myself yet and also cannot say anything about the other changes. 

<sub>Dominik Pinsel <dominik.pinsel@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md) </sub>

ndr-brt:
> I would say we shouldn't use Control- & Data Plane charts anymore for version newer then 0.1.7 since they are deprecated anyway. Otherwise we have to refactor them to make it compatible with v0.3.x For the `edc.receiver.http.dynamic.endpoint` we have to add new property in our TXDC chart.

with hindsight they could have been removed in 0.3.0 as breaking changes were expected anyway. I will create an issue

DominikPinsel:
So, to sum it up I would suggest to 

[...]

PR #178, opened 2023-04-03
feat(charts): removes edc-controlplane and edc-dataplane charts

wolf4ood:
This PR removes the following charts

- `edc-controlplane` 
-  `edc-dataplane` 

We are currently using only the `tractus-connector` one


florianrusch-zf (approved):

tuncaytunc-zf:
Maybe removing these charts needs to be aligned with stakeholders, as @DominikPinsel has mentioned here:
https://github.com/eclipse-tractusx/tractusx-edc/issues/167#issuecomment-1490007164

florianrusch-zf:
I think it's not necessary to align this with the "stakeholders" (who ever this is) as we have marked them already as deprecated. Who wants to keep using them needs to take a copy from them and store it somewhere locally 😉

ndr-brt (approved):
:axe:

wolf4ood:
Hi @SebastianBezold could you merge this? Thanks :)

Issue #180, opened 2023-04-05
Object has already been returned to this pool or is invalid


tuncaytunc-zf:
## Describe the bug
In `edc-controlplane-postgresql` an exception occurs during startup

Reported in [Jira](https://jira.catena-x.net/browse/A1IDSC-482)

### To Reproduce
On startup from controlplane when you use the image `edc-controlplane-postgresql`

### Screenshots/Error Messages
```
SEVERE 2023-04-05T08:06:08.375372 StateMachineManager [transfer-process] error caught
java.lang.IllegalStateException: Object has already been returned to this pool or is invalid
    at org.apache.commons.pool2.impl.BaseGenericObjectPool.markReturningState(BaseGenericObjectPool.java:1251)
    at org.apache.commons.pool2.impl.GenericObjectPool.returnObject(GenericObjectPool.java:1029)
    at org.eclipse.edc.sql.pool.commons.CommonsConnectionPool.returnConnection(CommonsConnectionPool.java:76)
    at org.eclipse.edc.sql.datasource.PooledDataSourceConnection.close(PooledDataSourceConnection.java:96)
    at org.eclipse.edc.connector.store.sql.transferprocess.store.SqlTransferProcessStore.lambda$nextForState$1(SqlTransferProcessStore.java:80)
    at org.eclipse.edc.transaction.spi.NoopTransactionContext.execute(NoopTransactionContext.java:34)
    at org.eclipse.edc.connector.store.sql.transferprocess.store.SqlTransferProcessStore.nextForState(SqlTransferProcessStore.java:69)
    at org.eclipse.edc.connector.transfer.process.TransferProcessManagerImpl.lambda$processTransfersInState$21(TransferProcessManagerImpl.java:591)
    at org.eclipse.edc.statemachine.StateProcessorImpl.process(StateProcessorImpl.java:46)
    at java.base/java.util.stream.ReferencePipeline$5$1.accept(Unknown Source)
    at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
    at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
    at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)
[...]

PR #181, opened 2023-04-05
fix: Issue with sql pool

tuncaytunc-zf:
Fixes the issue with sql pool by adding the missing dependencies.

Fixes the issue: #180

### DIFF
--- a/edc-controlplane/edc-controlplane-postgresql/build.gradle.kts
+++ b/edc-controlplane/edc-controlplane-postgresql/build.gradle.kts
@@ -12,9 +12,10 @@ dependencies {
     runtimeOnly(project(":edc-extensions:postgresql-migration"))
     runtimeOnly(edc.azure.vault)
     runtimeOnly(edc.bundles.sqlstores)
+    runtimeOnly(edc.transaction.local)
+    runtimeOnly(edc.sql.pool)
     runtimeOnly(edc.core.controlplane)
     runtimeOnly(edc.dpf.transfer)
-
 }
 
 


wolf4ood (approved):

florianrusch-zf (approved):

florianrusch-zf:
@tuncaytunc-zf do you know, why this wasn't catched by any tests? Is it even possible to fetch this by any kind of test?

matgnt:
> @tuncaytunc-zf do you know, why this wasn't catched by any tests? Is it even possible to fetch this by any kind of test?

Fully agree. I also wanted to create a few black box tests for the image (which can be only additional tests to what you mentioned) and what limits me is that with the image you have to use azure vault and I can not mock it because of hardcoded domain name in AzureVault.java
```
.vaultUrl("https://" + keyVaultName + ".vault.azure.net")
```
If no security reason is against this, I would propose to make this configurable.

tuncaytunc-zf:
> @tuncaytunc-zf do you know, why this wasn't catched by any tests? Is it even possible to fetch this by any kind of test?

Probably this bug was made by Gradle migration. Unfortunately our tests don't cover so much, in our business-test we just test one image "edc-controlplane-postgresql-hashicorp-vault". This is not the only bug we couldn't catch, there were many others too which could be caught when we would have more tests in place.

tuncaytunc-zf:
Of course we need more tests but starting such a fundamental discussion on such a minor bugfix makes no sense for me. 
We should have had such a discussion before we made a lot of fundamental changes in our code where we fixed them afterwards and still fixing, this is not the only one we have fixed. It seems for me we didn't test our fundamental changes so much. I think we could spend more time on making our products more stable instead of refactoring running/stable code were we introduce a lot of new bugs, since we really don't have so much tests. Instead of implementing new tests we have already started to refactor the existing ones :smiley:

stefan-ettl (approved):
Hi all, let's focus on the fix first, as Tuncay said.

PR #194, opened 2023-04-11
docs: decision record about refactoring helm charts

paullatzelsperger:
## WHAT

adds a decision record about refactoring our helm charts as per meeting of March 20th 2023.

## WHY
outlining the reasoning and especially the technical approach behind the decision


ndr-brt (approved):
neat

PR #208, opened 2023-04-12
feature: create Helm chart for a pure in-mem runtime

paullatzelsperger:
## WHAT

this PR adds a new Helm Chart `tractusx-connector-memory`, that uses only in-memory components (save for DAPS). Furthermore, `edc-controlplane-memory` was renamed to `edc-runtime-memory`, and I removed the Azure KeyVault from it. 
I also created a new folder for the test infrastructure, which is a pruned-down version of the "old" `supporting-infrastructure`

## WHY

Could be useful for testing or development purposes. lay the ground work for deploy-testing the production charts (future PRs)

## FURTHER NOTES
- added an in-memory vault, that can be seeded with secrets upon startup. 
- moved the `run-deployment-test` workflow to a composite action
- used `helm test` to ping the runtime's health endpoint
- obviously, more elaborate tests can be added, such as sending `POST` requests against the management api, and verifying that an object was actually created, etc.

## OPEN QUESTIONS
- seed secrets are stored in `.github/workflows/deploy-test-secrets`. is that the proper place?


.github/actions/publish-docker-image/action.yml
florianrusch-zf, line 1: Why are these changes necessary?

.github/actions/run-deployment-test/action.yml
```diff
@@ -0,0 +1,104 @@
...
+    - name: Setup Kubectl
+      uses: azure/setup-kubectl@v3.2
+
+    - name: Create k8s Kind Cluster
+      uses: helm/kind-action@v1.5.0
+
+    - name: Load images into KinD
+      shell: bash
+      run: |
+        kind get clusters | xargs -n1 kind load docker-image ${{ inputs.imagename}}:${{ inputs.image_tag}} --name
```
florianrusch-zf, line 80:
```suggestion
        kind get clusters | xargs -n1 kind load docker-image ${{ inputs.imagename }}:${{ inputs.image_tag }} --name
```

.github/actions/run-deployment-test/action.yml
```diff
@@ -0,0 +1,104 @@
...
+
+    #################
+    ### Tear Down ###
+    #################
+    - name: Destroy the kind cluster
+      if: always()
+      shell: bash
+      run: >-
+        kind get clusters | xargs -n1 kind delete cluster --name
+
```
florianrusch-zf, line 104:
```suggestion

```

.github/actions/run-deployment-test/action.yml
```diff
@@ -0,0 +1,104 @@
...
+      shell: bash
+      run: |-
+        helm install infra edc-tests/deployment/src/main/resources/helm/test-infrastructure \
+        --wait-for-jobs --timeout=30s --dependency-update
+
+    - name: Install In-Mem Runtime
+      shell: bash
+      run: ${{ inputs.helm_command }}
+
+
```
florianrusch-zf, line 95:
```suggestion
```

.github/dependabot.yml
```diff
@@ -2,8 +2,7 @@
 version: 2
 updates:
   # Maven
-  -
-    package-ecosystem: "maven"
+  - package-ecosystem: "maven"
```
florianrusch-zf, line 5: This needs to be replaced with gradle, right? Do you know if we have an issue for that already?

settings.gradle.kts
```diff
@@ -3,6 +3,7 @@ rootProject.name = "tractusx-edc"
 include(":edc-extensions:business-partner-validation")
 include(":edc-extensions:control-plane-adapter")
 include(":edc-extensions:cx-oauth2")
+include(":edc-extensions:vault-memory")
```
florianrusch-zf: Yes, I would highly agree to move it into the `edc-runtime-memory` module. If we keep it as an extension we would also publish this as a jar which mean people will maybe us it, which again means we need to maintain/support/update in a good shape. Of course we also need to maintain/support/update the `edc-runtime-memory` but I think it's less work for us

settings.gradle.kts
```diff
@@ -138,37 +139,37 @@ dependencyResolutionManagement {
             library("micrometer-jetty", "org.eclipse.edc", "jetty-micrometer").versionRef("edc")
             library("monitor-jdklogger", "org.eclipse.edc", "monitor-jdk-logger").versionRef("edc")
             library(
-                "transfer.dynamicreceiver",
-                "org.eclipse.edc",
-                "transfer-pull-http-dynamic-receiver"
+                    "transfer.dynamicreceiver",
```
florianrusch-zf, line 141: The indention looks weird for me. 8 spaces? Is this correct?

edc-tests/deployment/src/main/resources/helm/test-infrastructure/values.yaml
```diff
@@ -0,0 +1,185 @@
...
+        LXRWZjHwexDqtgMGTCN/7aZlMA8GA1UdEwEB/wQFMAMBAf8wDQYJKoZIhvcNAQEL
+        BQADggEBAD2a5kuIdICNXfYLpSe7AIONwZVucaArYtpXBxHEy5lMJsTEJgjZzypd
+        iIMU7onEQGVbii6yVNpWfIpJYM4e8ytVdJuk5evclVKZs/lZ2IshLyWFVj+ITh2E
+        28X4C/Hnmt4MPBCNowQf71nMp4LEziBgXp54qFV9C+qSTEVdrherRE0PU/zKyX10
+        S/P5o42weTHnAO/pBN/8AmL3AymynKVgcPaW46IjjRAuc6kfZWCrYQ0M4+/7Ws5r
+        uM55Zae/L+C82OTNNaaK324ogsCkORPeQ23OCrRD8rZJmQ9bpoOGglPminfwEOhB
+        UHtyKgmvqCyOV3G/4G93W/xsLV0kxLA=
+        -----END CERTIFICATE-----
+        EOF
+        }
```
florianrusch-zf, line 185:
```suggestion
        }

```

edc-tests/deployment/src/main/resources/helm/test-infrastructure/README.md
```diff
@@ -0,0 +1,95 @@
...
+> risk! ([source](https://github.com/Fraunhofer-AISEC/omejdn-server))
+
+### HashiCorp Vault
+
+The Control- and Data Plane persist confidential in the vault and persist and communicate using only the secret
+names. Hence, it is not possible to run a connector without an instance of a vault.
+
+### Backend Application
+
+After a Data Transfer is successfully prepared the control plane will contact the a configurable endpoint with the
```
florianrusch-zf:
```suggestion
After a Data Transfer is successfully prepared the control plane will contact the configurable endpoint with the
```

florianrusch-zf (changes requested):
just a few comments already - hard to review if you are still pushing things 😅 Will try to review again at the weekend.

edc-tests/deployment/src/main/resources/helm/test-infrastructure/README.md
paullatzelsperger: in the meantime i removed this section

settings.gradle.kts
paullatzelsperger: yep, already done

.github/actions/publish-docker-image/action.yml
paullatzelsperger: i just came across them and fixed them. a composite action should not care about the trigger event or execution context

.github/dependabot.yml
paullatzelsperger: > just a few comments already - hard to review if you are still pushing things 😅 Will try to review again at the weekend.

yeah, i'm sorry, just went through the code, and noticed a couple of things, which i pushed right away ... :/

paullatzelsperger: i don't think we haven issue, will create one

settings.gradle.kts
paullatzelsperger: nope, thx

README.md
```diff
@@ -16,24 +16,28 @@ Please also refer to:
...
+The eclipse data space connector is split up into Control-Plane and Data-Plane, whereas the Control-Plane functions as
+administration layer
 and has responsibility of resource management, contract negotiation and administer data transfer.
 The Data-Plane does the heavy lifting of transferring and receiving data streams.
 
 Depending on your environment there are different derivatives of the control-plane prepared:
 
-- [edc-controlplane-memory](edc-controlplane/edc-controlplane-memory) with dependency onto
+- [edc-runtime-memory](edc-controlplane/edc-runtime-memory) with dependency onto
   - [Azure Key Vault](https://azure.microsoft.com/en-us/services/key-vault/#product-overview)
```
wolf4ood, line 33: This deps is not valid anymore right?

docs/README.md
```diff
@@ -13,7 +15,7 @@ When running a EDC connector from the Tractus-X EDC repository there are three s
 The three supported setups are.
 
 - Setup 1: In Memory & Azure Vault
-  - [Control Plane](../edc-controlplane/edc-controlplane-memory/README.md)
+  - [Control Plane](../edc-controlplane/edc-runtime-memory/README.md)
     - [IDS DAPS Extensions](https://github.com/eclipse-edc/Connector/tree/main/extensions/common/iam/oauth2/daps)
     - In Memory Persistence done by using no extension
     - [Azure Key Vault Extension](https://github.com/eclipse-edc/Connector/tree/main/extensions/common/vault/azure-vault)
```
wolf4ood, line 21: Also this one probably the dep to the vault  it's not valid anymore

edc-controlplane/edc-runtime-memory/src/main/java/org/eclipse/tractusx/edc/vault/memory/VaultMemoryExtension.java
```diff
@@ -0,0 +1,53 @@
...
+@Extension(value = "In-memory vault extension", categories = {"vault", "security"})
+public class VaultMemoryExtension implements ServiceExtension {
+
+    @Setting(value = "Secrets with which the vault gets initially populated. Specify as comma-separated list of key:secret pairs.")
+    public static final String VAULT_MEMORY_SECRETS_PROPERTY = "edc.vault.secrets";
+
+    @Override
+    public String name() {
+        return "In-Memory Vault Extension";
+    }
```
wolf4ood, line 36:
```suggestion
@Extension(value = VaultMemoryExtension.NAME, categories = {"vault", "security"})
public class VaultMemoryExtension implements ServiceExtension {

    public static final String NAME = "In-Memory Vault Extension";

    @Setting(value = "Secrets with which the vault gets initially populated. Specify as comma-separated list of key:secret pairs.")
    public static final String VAULT_MEMORY_SECRETS_PROPERTY = "edc.vault.secrets";

    @Override
    public String name() {
        return NAME;
    }
```

wolf4ood (commented):
Just a nit I would store the name on a static var

paullatzelsperger:
had to close, will re-open

.github/dependabot.yml
paullatzelsperger: [edit] #216

PR #211, opened 2023-04-13
fix: Adapt Helm Chart for version 0.3.x

tuncaytunc-zf:
WHAT
Fixes some Helm Chart issues reported in #167 

WHY
The Helm Chart was not compatible with the version 0.3.x

### DIFF
--- a/edc-dataplane/edc-dataplane-base/build.gradle.kts
+++ b/edc-dataplane/edc-dataplane-base/build.gradle.kts
@@ -4,19 +4,18 @@ plugins {
 }
 
 dependencies {
-    implementation(edc.config.filesystem)
-    implementation(edc.dpf.awss3)
-    implementation(edc.dpf.oauth2)
-    implementation(edc.dpf.http)
-
-    implementation(edc.dpf.framework)
-    implementation(edc.dpf.api)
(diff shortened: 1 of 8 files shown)


edc-dataplane/edc-dataplane-base/build.gradle.kts
```diff
@@ -15,7 +15,8 @@ dependencies {
     implementation(edc.core.connector)
     implementation(edc.boot)
 
-
     implementation(edc.bundles.monitoring)
     implementation(edc.ext.http)
+
+    runtimeOnly(project(":edc-extensions:observability-api-customization"))
```
paullatzelsperger, line 21: our custom extension is a drop-in replacement, so `implementation(edc.api.observability)` (couple of lines above) should be deleted. And while you're there: could you replace all `implementation` dependencies with `runtimeOnly` dependencies?

also: nit - project-dependencies should be the first in the list.

paullatzelsperger (changes requested):
`edc.api.observability` should be removed, as it is replaced by "our" observability API

paullatzelsperger (approved):

ndr-brt (approved):

PR #212, opened 2023-04-13
chore: update to temurin 17

paullatzelsperger:
## WHAT
updates all Dockerfiles and CI jobs to use `temurin-17`

## WHY
it is the latest LTS version

### DIFF
--- a/.github/actions/publish-docker-image/action.yml
+++ b/.github/actions/publish-docker-image/action.yml
@@ -60,7 +60,7 @@ runs:
     - name: Set up JDK 11
       uses: actions/setup-java@v3.11.0
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'temurin'
         cache: 'gradle'
     - name: Build Controlplane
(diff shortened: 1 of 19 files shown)


florianrusch-zf (changes requested):
We should pin the parent images to a specific version in the Dockerfiles. Otherwise our builds wouldn't be reproducible.

paullatzelsperger:
> We should pin the parent images to a specific version in the Dockerfiles. Otherwise our builds wouldn't be reproducible.

`17.0.6_12-jre-alpine` seems to be the latest one

florianrusch-zf (approved):

florianrusch-zf (changes requested):
17.0.6.12 doesn't exist. Only the 17.0.6.10 exists on hub.docker.com https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17.0.6

paullatzelsperger:
> 17.0.6.12 doesn't exist. Only the 17.0.6.10 exists on hub.docker.com https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17.0.6

don't know how that ended up in my clipboard...

paullatzelsperger:
merging, as previously approved.

PR #251, opened 2023-04-19
feature: add explicit docker image creation during release process

### LINKED ISSUE #250: Bug: Docker images are not tagged with release version
## Describe the bug

When creating a new release, for example `4.2.0`, Docker images should be tagged with `...:4.2.0` (amongst others). 
For some reason this does not happen. 

### To Reproduce

Create a new release version (DON'T DO THIS IN THE UPSTREAM!)

### Expected behavior

the version tag is created

_If applicable, add screenshots and/or error messages to help explain your problem._

## Context Informations

first observed during the `0.3.3` release on April 19, 2023

## Possible Implementation

My initial suspicion was, that either the `docker/metadata-action` does not pick up the newly created tag, or creating a tag does not trigger the `build.yaml` workflow. 

On closer inspection, the reason turned out to be, that in order for workflows to be triggered from workflow runs, we apparently cannot use a `GITHUB_TOKEN`, but must use a PAT, according to the [documentation](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow). That means, that the workflow that _creates_ the tag (i.e. `publish_new_release.yaml`) **cannot** run using a `GITHUB_TOKEN`, it must use a PAT, otherwise the other job, that publishes the image (`build.yaml`) won't execute. The solution could be as easy as to remove the `permissions: ` block from the `github-release` job.

I tried it out by manually creating a Git tag, which did indeed trigger the workflow, and caused the correct Docker tag to be created.

paullatzelsperger:
## WHAT

This PR adds a step to the `publish-new-release.yaml` workflow, that explicitly builds and pushes all Docker images with the release version as Docker tag.

For example, if we'd just released version `0.6.9`, it would create `edc-runtime-memory:0.6.9`, etc.

## WHY

Due to a security restriction, workflow runs that use `GITHUB_TOKEN` cannot trigger other workflows. 
Here, this is a problem because the `publish-new-release.yaml` flow creates a Git tag, which would normally trigger another `build.yaml` flow, which would in turn build docker images with the correct version tag. Since we're using `GITHUB_TOKEN`s instead of PATs this doesn't happen.

## FURTHER NOTES

I also simplified the use of `matrix` builds in the other flows, effectively reducing the number of jobs that are launched.

Closes #250

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -38,51 +38,29 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  create-docker-image-controlplane:
+  create-docker-image:
     name: "Create Docker Images for the ControlPlane"
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        name:
-          - edc-runtime-memory
-          - edc-controlplane-memory-hashicorp-vault
-          - edc-controlplane-postgresql
-          - edc-controlplane-postgresql-hashicorp-vault
+        variant: [ { dir: edc-controlplane, img: edc-runtime-memory },
+                   { dir: edc-controlplane, img: edc-controlplane-memory-hashicorp-vault },
+                   { dir: edc-controlplane, img: edc-controlplane-postgresql-hashicorp-vault },
+                   { dir: edc-controlplane, img: edc-controlplane-postgresql },
+                   { dir: edc-dataplane,    img: edc-dataplane-azure-vault },
+                   { dir: edc-dataplane,    img: edc-dataplane-hashicorp-vault } ]
     permissions:
       contents: write
       packages: write
     steps:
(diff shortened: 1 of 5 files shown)


paullatzelsperger:
markdown lint errors will be fixed once #252 is merged

.github/workflows/publish-docker.yaml
```diff
@@ -38,51 +38,29 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  create-docker-image-controlplane:
+  create-docker-image:
     name: "Create Docker Images for the ControlPlane"
```
wolf4ood, line 42:
```suggestion
    name: "Create Docker Images"
```

wolf4ood (changes requested):
Just a nit

wolf4ood (approved):

PR #261, opened 2023-04-20
Fix wrong helm install command

florianrusch-zf:
## WHAT

In the README.md of the memory helm chart is the wrong chart referenced.

### DIFF
--- a/charts/tractusx-connector-memory/README.md.gotmpl
+++ b/charts/tractusx-connector-memory/README.md.gotmpl
@@ -12,7 +12,7 @@
 
 ```shell
 helm repo add tractusx-edc https://eclipse-tractusx.github.io/charts/dev
-helm install my-release tractusx-edc/tractusx-connector --version {{ .Version }}
+helm install my-release tractusx-edc/tractusx-connector-memory --version {{ .Version }}
 ```
 
 {{ template "chart.maintainersSection" . }}


florianrusch-zf:
@paullatzelsperger do I need to do something to generate the README.md file? Or will this been done automatically somehow?

paullatzelsperger:
> @paullatzelsperger do I need to do something to generate the README.md file? Or will this been done automatically somehow?

@florianrusch-zf in the `charts/` directory, simply execute the following:
```bash
docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:v1.10.0
```

paullatzelsperger:
> @paullatzelsperger so we need to do this manually and there is no automatism for this at the moment. okay, good to know!

i haven't seen any. the question is, at which point we would insert such an automation. intuitively, i would do this during the release process.

florianrusch-zf:
> What we could do, is insert checks that the README's are up-to-date. In EDC we do a similar thing with the [API doc](https://github.com/eclipse-edc/Connector/blob/main/.github/workflows/apidoc.yaml?rgh-link-date=2023-04-20T12%3A48%3A17Z#L18) 

Thats a great idea! I will create an issue for that ;-)

paullatzelsperger (approved):

PR #259, opened 2023-04-20
build(deps): add constraints to avoid vulnerable transitive dependencies

ndr-brt:
## WHAT

constraints: 
- `net.minidev:json-smart` to version `2.4.10`
- `org.yaml:snakeyaml` to version `2.0`

## WHY

to avoid hi-level vulnerabilities detected by veracode

## FURTHER NOTES

Closes #258

PR #273, opened 2023-04-24
build(deps): Move Gradle dependencies constrains into root build.gradle.kts

tuncaytunc-zf:
## WHAT

The following constrains were centralized to apply to all modules:

net.minidev:json-smart to version 2.4.10
org.yaml:snakeyaml to version 2.0

## WHY

To avoid hi-level vulnerabilities in all modules detected by Veracode


build.gradle.kts
```diff
@@ -159,3 +159,14 @@ subprojects {
...
+dependencies {
+    constraints {
+        implementation("org.yaml:snakeyaml:2.0") {
+            because("version 1.33 has vulnerabilities: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-1471.")
+        }
+        implementation("net.minidev:json-smart:2.4.10") {
+            because("version 2.4.8 has vulnerabilities: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-1370.")
+        }
+    }
+}
```
florianrusch-zf, line 172:
```suggestion
}

```

florianrusch-zf (changes requested):

florianrusch-zf:
In the past I already asked, if it wouldn't make more sense to have it in the root `build.gradle.kts` and the answer from @ndr-brt  was (PR #259):

> having them underneath the "parent" dependency that caused the vulnerability will make them more visible when it will be possible to remove them (because of upgrades).

Soo, how do we want to handle this now?

tuncaytunc-zf:
> In the past I already asked, if it wouldn't make more sense to have it in the root `build.gradle.kts` and the answer from @ndr-brt was (PR #259):
> 
> > having them underneath the "parent" dependency that caused the vulnerability will make them more visible when it will be possible to remove them (because of upgrades).
> 
> Soo, how do we want to handle this now?

I don't know how else we can solve such a transitive dependencies better then putting it into global build file. Otherwise we have to do it in every single module, which would lead to a lot of inconsistency in terms of upgrading.  @ndr-brt do you have a better idea?

ndr-brt (approved):

PR #297, opened 2023-04-28
docs: Adapt local setup documentations

### LINKED ISSUE #286: Update used image version in documentation
### What

- [ ] adapt the referenced used version for the image
- [ ] analyze if the file itself is still needed/used. Cause right now it seems we have several of these manuals.

### Why

the referenced image version is 0.2 which is not correct in this [file](https://github.com/eclipse-tractusx/tractusx-edc/blob/main/docs/samples/Local%20TXDC%20Setup.md)

tuncaytunc-zf:
## WHAT

Adapt local setup documentation. 

## WHY

Some values in some local setup document were outdated. 

## FURTHER NOTES

Closes #286

### DIFF
--- a/docs/development/Run-business-tests-local.md
+++ b/docs/development/Run-business-tests-local.md
@@ -14,7 +14,7 @@ Prerequisites:
 ## 2. Install the all-in-one supporting infrastructure environment (Daps, Vault, PostgreSql, Minio, Backend-Service)
 
 ```shel
-helm install infrastructure edc-tests/src/main/resources/deployment/helm/supporting-infrastructure -n business-tests --dependency-update --create-namespace
+helm install infrastructure edc-tests/cucumber/src/main/resources/deployment/helm/supporting-infrastructure -n business-tests --dependency-update --create-namespace
 ```
 
 To access the PostgreSql databases you could use following kubectl port forwardings:
@@ -32,7 +32,7 @@ Please use the same ports later for your environment variables.
 helm install plato charts/tractusx-connector -n business-tests --create-namespace \
   --set fullnameOverride=plato \
   --set controlplane.service.type=NodePort \
-  --set controlplane.endpoints.data.authKey=password \
+  --set controlplane.endpoints.management.authKey=password \
   --set controlplane.image.pullPolicy=Never \
   --set controlplane.image.tag=latest \
   --set controlplane.image.repository=edc-controlplane-postgresql-hashicorp-vault \
@@ -45,7 +45,6 @@ helm install plato charts/tractusx-connector -n business-tests --create-namespac
   --set postgresql.username=user \
   --set postgresql.password=password \
   --set postgresql.jdbcUrl=jdbc:postgresql://plato-postgresql:5432/edc \
-  --set vault.hashicorp.enabled=true \
   --set vault.hashicorp.url=http://vault:8200 \
   --set vault.hashicorp.token=root \
   --set vault.secretNames.transferProxyTokenSignerPublicKey=plato/daps/my-plato-daps-crt \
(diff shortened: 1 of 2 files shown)


stephanbcbauer (approved):
LGTM

docs/samples/Local TXDC Setup.md
```diff
@@ -97,10 +96,9 @@ helm install sokrates charts/tractusx-connector  \
     --namespace cx \
     --create-namespace \
     --set fullnameOverride=sokrates \
-    --set controlplane.image.tag=0.2.0 \
+    --set controlplane.image.tag=latest \
```
paullatzelsperger, line 99: using `latest` is dangerous, because the `latest` tag gets overwritten on every build we do on tractusx-edc

paullatzelsperger (changes requested):

docs/samples/Local TXDC Setup.md
```diff
@@ -64,10 +64,9 @@ helm install plato charts/tractusx-connector \
     --namespace cx \
     --create-namespace \
     --set fullnameOverride=plato \
-    --set controlplane.image.tag=0.2.0 \
+    --set controlplane.image.tag=latest \
```
paullatzelsperger, line 67: using `latest` is dangerous, because the `latest` tag gets overwritten on every build we do on tractusx-edc

docs/development/Run-business-tests-local.md
```diff
@@ -14,7 +14,7 @@ Prerequisites:
 ## 2. Install the all-in-one supporting infrastructure environment (Daps, Vault, PostgreSql, Minio, Backend-Service)
 
 ```shel
```
sekikn, line 16: Should we fix the format type as "sh" or "shell"?

docs/samples/Local TXDC Setup.md
```diff
@@ -33,7 +33,7 @@ Before the connectors can be setup, the Supporting Infrastructure must be in pla
 to run two connectors independently.
 
 For this local test scenario,
-the [Supporting Infrastructure](../../edc-tests/src/main/resources/deployment/helm/supporting-infrastructure/README.md)
+the [Supporting Infrastructure](../../edc-tests/cucumber/src/main/resources/deployment/helm/supporting-infrastructure/README.md)
```
sekikn, line 36: Just like this fix, should we also insert "cucumber" into the following two paths?

docs/samples/Local TXDC Setup.md
sekikn, line 67: I tried the fixed command, but ran into the following error unfortunately.

```
Error: INSTALLATION FAILED: execution error at (tractusx-connector/templates/deployment-controlplane.yaml:191:54): .Values.postgresql.username is required
```

Considering the context that PostgreSQL is disabled in the preceding infrastructure setup, how about adding the following option?

```
--set controlplane.image.repository=edc-controlplane-memory-hashicorp-vault
```

In addition, even if specifying the option above, `postgresql.username`, `postgresql.password`, and `postgresql.jdbcUrl` still seem to be required, though any value is OK for them. According to the document, [there is a `postgresql.enabled` option](https://github.com/eclipse-tractusx/tractusx-edc/blob/57da0cb124a3efea54d7fe5059d2b33b816167bf/charts/tractusx-connector/README.md?plain=1#L216), but it doesn't seem to be used in any Helm chart.

```
$ git grep postgresql.enabled
.github/workflows/business-tests.yaml:              --set postgresql.enabled=true \
.github/workflows/business-tests.yaml:              --set postgresql.enabled=true \
charts/tractusx-connector-app/README.md:| runtime.postgresql.enabled | bool | `false` |  |
charts/tractusx-connector/README.md:| postgresql.enabled | bool | `false` |  |
docs/development/Run-business-tests-local.md:  --set postgresql.enabled=true \
docs/development/Run-business-tests-local.md:  --set postgresql.enabled=true \
```

Is this a documentation bug, and if so, should we fix it in this PR?

PR #299, opened 2023-04-28
refactor(chart): move test infrastructure into the test chart

paullatzelsperger:
## WHAT

This PR removes the `tractusx-connector-app` chart and moves the dependencies into the `test-infrastructure` chart

## WHY

the `tractusx-connector-app` chart:
- contained mostly test code
- was not publishable, due to a relative filesystem link
- is not actually a deliverable of TractusX-EDC

## FURTHER NOTES

- this was actually a bug, because it prevented the helm-release action from succeeding, and it was discovered during a dry-run of the release
- **we need this for the 0.3.4 release!**


wolf4ood (approved):
LGTM

PR #320, opened 2023-05-09
chore(build): use composite action for GPG import

### LINKED ISSUE #166: Setup and implement publishing Maven artefacts to MavenCentral and OSSRH
## WHAT

In order to have tractusx-edc consumable also on code-level we need to publish all artefacts to MavenCentral. In Eclipse projects that happens through an EF-managed Jenkins instance ("JIPP") that contains signing keys for MavenCentral.

## WHY
MavenCentral requires all artefacts to be signed. That is done in Jenkins.

## Implementation proposal
- remove the `publishAllPublicationsToGithubPackagesRepository` task
- ~~GH Actions triggers Jenkins (through [this action](https://github.com/toptal/jenkins-job-trigger-action))~~
- ~~on releases: the `version` parameter does not contain the `-SNAPSHOT` suffix, thus causing it to sync to MavenCentral~~
- use the credentials provided by [this EF helpdesk ticket](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2976) to sign the artefacts, and to publish to OSSRH/MavenCentral.
- **No Jenkins instance is required, we can directly push to OSSRH!**
- comment on [this OSSRH issue](https://issues.sonatype.org/browse/OSSRH-91206) once our first release is published successfully.

## OPEN QUESTIONS
- what's our publishing schedule? Snapshots on every push to `develop`? Nightlies?
- ~~do we already have a publishing environment set up?~~
- if not: who will open helpdesk issues with EF and OSSRH? -> done [here](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2976)
- ~~should we host the declarative pipelines in a GitHub repo? (in EDC we do that)~~

[edit]: updated description based on not having to use Jenkins.

paullatzelsperger:
## WHAT

Refactors importing the GPG key into a composite action.

## WHY

avoid code duplication

## FURTHER NOTES

- the `list keys` step is necessary as it creates the necessary directories and files
- we'll see after the merge whether it actually works, because the secrets are not present on PRs, and the job only runs on `main` and `releases`

Closes #166


.github/actions/import-gpg-key/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Import GPG Private Key
+      shell: bash
+      run: |
+        echo "use-agent" >> ~/.gnupg/gpg.conf
+        echo "pinentry-mode loopback" >> ~/.gnupg/gpg.conf
+        echo -e "${{ inputs.gpg-private-key }}" | gpg --import --batch
+        for fpr in $(gpg --list-keys --with-colons | awk -F: '/fpr:/ {print $10}' | sort -u);
+        do
+          echo -e "5\\ny\\n" |  gpg --batch --command-fd 0 --expert --edit-key $fpr trust;
+        done
```
florianrusch-zf, line 46:
```suggestion
        done

```

florianrusch-zf (approved):

PR #332, opened 2023-05-10
feat(build): delete test keys from repo

### LINKED ISSUE #278: delete keys from repository
## Describe the bug

GitGuardien mentioned some issues with our dummy keys in our repository. They are only used for testing purpose and are not valid in the outside world, but there is a need to get rid of them

### Expected behavior

The keys should be stored in the repository.

image

## Possible Implementation (already discussed in teams)

- if we just delete them, some of our GH test actions would fail
- i see two ways forward: a) convince whoever mandated their deletion, that test keys are not really harmful, or b) go Florian Rusch (Guest)'s route an regenerate them every test run. 

b) is cleaner, but more effort

paullatzelsperger:
## WHAT

This PR adds scripts and tools to generate certificates and keys, so that we do not have them checked in the code base.

## WHY

security guidelines prohibit checking in even test keys

## FURTHER NOTES

- scripts should be used as much as possible to increase portability and user-friendlyness
- unless stated otherwise, a POSIX-compliant shell (e.g. bash) is assumed
- the following tools are presumed to be present:
    - `openssl` to generate certs/keys
    - `yq` to manipulate YAML files
    - `base64` to encode/decode base64 strings

Closes #278


paullatzelsperger:
I don't know why Gitguardian still complains, I did actually _delete_ the hard-coded secret.

PR #342, opened 2023-05-14
feat: replace using hard-coded certs with dynamically generated ones

### LINKED ISSUE #278: delete keys from repository

paullatzelsperger:
## WHAT

Deletes the test keys from the repository

## WHY

required by DevSecOps 

## FURTHER NOTES

- don't know why GitGuardian still complains, the secret was actually _removed_ from the code base

Closes #278


wolf4ood (approved):

paullatzelsperger:
Will merge, even though GitGuardian still complains about a secret, that has - in fact - been _deleted_. Sounds like a bug in GitGuardian.

PR #377, opened 2023-05-18
chore: remove old control-plane adapter

wolf4ood:
## WHAT

Removes the old Control Plane Adapter (CPA)

## WHY

The old CPA was not compatible with the new DSP protocol of EDC, and it will be replaced with a new way of negotiating EDR and requesting data in a simplified way.

DSP/AAS integration is described [here](https://github.com/Metaform/edc-aas).


paullatzelsperger (approved):
LGTM

PR #378, opened 2023-05-18
feature: switch to Dataspace Protocol

paullatzelsperger:
## WHAT

This PR switches Tractus-X EDC over to the new Dataspace Protocol. Alongside the protocol-level changes, there are a few side effects:
- the existing management API is now deprecated and will be replaced with a JSON-LD-aware variant soon. Currently, the "new" mgmt API is available at the `.../v2/...` path
- removed `business-tests`, as they only tested things that are already tested otherwise (deployment tests, jvm-runner, TCK,...) and it was not compatible to DSP
- removed the `control-plane-adapter`: it was not compatible with DSP, and it has been replaced with a light-weight implementation 
- uses EDC `0.0.1-milestone-9`
- a check was added to verify that all relevant files have a license header

## WHY

The legacy IDS Multipart protocol is deprecated and will be maintained henceforn and is expected not to work anymore with EDC.

## FURTHER NOTES

.


paullatzelsperger:
\cc @jimmarino

wolf4ood (approved):

PR #382, opened 2023-05-18
feat(tests): adds pg implementations for E2E tests

### LINKED ISSUE #379: E2E tests add Postgresql variant
## Feature Request

For each e2e tests we should have two variants that run:

- In memory
- With Postgresql

This will cover the connector running in test and prod-like environment

wolf4ood:
## WHAT

Adds PG variant for E2E tests

## WHY

To cover the connector running in test and prod-like environment

Closes #379


edc-tests/runtime/build.gradle.kts
```diff
@@ -1,53 +1,32 @@
 /*
```
ndr-brt, line 1: is this module needed?

ndr-brt (approved):

.github/workflows/verify.yaml
```diff
@@ -127,3 +127,22 @@ jobs:
...
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: password
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-java
+
+      - name: Run Postgresql E2E tests
+        run: ./gradlew :edc-tests:runtime:build  test -DincludeTags="PostgresqlIntegrationTest"
```
paullatzelsperger, line 148: the `...:build` task is obsolete, since you don't scope `test` to the same project, so it will compile everything anyway.

.github/workflows/verify.yaml
```diff
@@ -127,3 +127,22 @@ jobs:
 
       - name: Run E2E tests
         run: ./gradlew :edc-tests:runtime:build test -DincludeTags="EndToEndTest"
+
+  postgres-end-to-end-tests:
```
paullatzelsperger, line 131: nit: rename to `postgres-tests`. How would you feel about running this in the `integration-test` job instead as a separate job?

edc-tests/runtime/runtime-postgresql/build.gradle.kts
```diff
@@ -0,0 +1,57 @@
...
+    implementation(libs.edc.core.controlplane)
+    // for the controller
+    implementation(libs.jakarta.rsApi)
+}
+
+application {
+    mainClass.set("org.eclipse.edc.boot.system.runtime.BaseRuntime")
+}
+
+tasks.withType<com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar> {
```
paullatzelsperger, line 49: doesn't need to be "shadow-jar-able"

paullatzelsperger (changes requested):

edc-tests/runtime/build.gradle.kts
wolf4ood: btw removed

.github/workflows/verify.yaml
wolf4ood: removed

PR #477, opened 2023-06-13
feat: Initial implementation of Catena-X policies

### LINKED ISSUE #476: feat: Implement the new Catena-X Policies
We need to implement support for the 3.1 Catena-X policies according to these specifications:

- [Summary Policy](https://github.com/eclipse-tractusx/ssi-docu/blob/main/docs/credentials/summary/summary.vc.md)
- [Catena-X Policies](https://github.com/eclipse-tractusx/ssi-docu/blob/main/docs/architecture/cx-3-2/edc/policy.definitions.md)

These will be done by introducing a new CX policy extension.

jimmarino:
## WHAT

This PR provides an initial implementation of SSI policies per this [specification](https://github.com/eclipse-tractusx/ssi-docu/blob/main/docs/architecture/cx-3-2/edc/policy.definitions.md) and the [Summary Credential](https://github.com/eclipse-tractusx/ssi-docu/blob/main/docs/credentials/summary/summary.vc.md).

## FURTHER NOTES

The CX policy module contains functions for working with Json-Ld structures. These functions (`JsonLdTypeFunctions` and `JsonLdValueFunctions`) will likely be refactored into a common module and shared with the new SSI-based `IdentityService` in a subsequent PR.

Closes #476

### DIFF
--- a/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/common/AbstractVpConstraintFunction.java
+++ b/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/common/AbstractVpConstraintFunction.java
@@ -1,0 +1,135 @@
+/*
+ *  Copyright (c) 2023 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  SPDX-License-Identifier: Apache-2.0
+ *
+ *  Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.tractusx.edc.policy.cx.common;
+
+import jakarta.json.JsonObject;
+import jakarta.json.JsonValue;
+import org.eclipse.edc.policy.engine.spi.AtomicConstraintFunction;
+import org.eclipse.edc.policy.engine.spi.PolicyContext;
+import org.eclipse.edc.policy.model.Operator;
+import org.eclipse.edc.policy.model.Permission;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.stream.Collectors;
+
+import static jakarta.json.JsonValue.ValueType.ARRAY;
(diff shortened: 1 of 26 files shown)


edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/common/JsonLdValueFunctions.java
```diff
@@ -0,0 +1,70 @@
...
+        } else {
+            return convertType(root);
+        }
+    }
+
+    /**
+     * Converts the value to a string representation.
+     */
+    @Nullable
+    private static String convertType(JsonValue value) {
```
paullatzelsperger, line 58: eventually we'll likely want to upstream these utilities

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/common/AbstractVpConstraintFunction.java
```diff
@@ -0,0 +1,135 @@
...
+            return null;
+        }
+
+        return subjectValue.asJsonObject();
+    }
+
+    /**
+     * Returns true if the actual operand value is a string literal case-insensitive equal to the expected value.
+     */
+    protected boolean validateRightOperand(String expectedValue, Object actualValue, PolicyContext context) {
```
paullatzelsperger, line 121: I think the expected value should be a vararg, because there could be several allowed ones, same as we have for the `validateOperator`

jimmarino: > I think the expected value should be a vararg, because there could be several allowed ones, same as we have for the `validateOperator`

Yeah, that method is going to actually have to be modified when we switch versions from the left to right operand as described [here](https://github.com/eclipse-tractusx/ssi-docu/pull/9). I'll do that in another PR.

wolf4ood (approved):

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/common/JsonLdValueFunctions.java
jimmarino: Yeah, I thinking we need separate set of Json-Ld utils.

paullatzelsperger (approved):

PR #480, opened 2023-06-14
chore(helm): add hook-delete-policy for Helm chart tests

fty4:
## WHAT

Adding [Hook deletion policies](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies) for the Helm tests.

## WHY

When running tests with Helm the pods will not be tidied up.
When the annotation `"helm.sh/hook-delete-policy": hook-succeeded` is added, after a successful run the pod will be deleted afterwards.
If the pod fails it will still be present.

Also the value `before-hook-creation` was added to be able to run the test again if it failed without manually deleting the pod.

<hr />

### DIFF
--- a/charts/tractusx-connector-azure-vault/templates/tests/test-controlplane-readiness.yaml
+++ b/charts/tractusx-connector-azure-vault/templates/tests/test-controlplane-readiness.yaml
@@ -26,6 +26,7 @@ metadata:
     {{- include "txdc.controlplane.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": test
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   containers:
     - name: wget
--- a/charts/tractusx-connector-azure-vault/templates/tests/test-dataplane-readiness.yaml
+++ b/charts/tractusx-connector-azure-vault/templates/tests/test-dataplane-readiness.yaml
@@ -26,6 +26,7 @@ metadata:
     {{- include "txdc.dataplane.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": test
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   containers:
     - name: wget
--- a/charts/tractusx-connector-memory/templates/tests/test-readiness.yaml
+++ b/charts/tractusx-connector-memory/templates/tests/test-readiness.yaml
@@ -29,6 +29,7 @@ metadata:
     {{- include "txdc.runtime.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": test
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   containers:
     - name: wget
--- a/charts/tractusx-connector/templates/tests/test-controlplane-readiness.yaml
+++ b/charts/tractusx-connector/templates/tests/test-controlplane-readiness.yaml
@@ -26,6 +26,7 @@ metadata:
     {{- include "txdc.controlplane.labels" . | nindent 4 }}
   annotations:
     "helm.sh/hook": test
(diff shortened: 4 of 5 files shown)


fty4:
Test is failing because test-pod is cleaned up as expected when successful and not longer present:

https://github.com/eclipse-tractusx/tractusx-edc/blob/2f81217ddcdb71ebdf3e6df4b27a85a1a5891d9f/.github/workflows/deployment-test.yaml#L155-L156

fty4:
The command should not be the problem - there is a bug that `--logs` will not act as expected:

helm/helm#10603

paullatzelsperger:
> The command should not be the problem - there is a bug that `--logs` will not act as expected:
> 
> [helm/helm#10603](https://github.com/helm/helm/pull/10603)

I tried this on my local machine, and there it works, but if removing the `--logs` flag solves it on CI, that would be fine. can you remove the flag in the workflow file, adding a comment about the helm bug and the link you posted?

fty4:
Sure then i would revert the last commit which adds the `tests.hookDeletePolicy` value in 7a832c7 to keep the value file more clean, right?

I thought removing the `--logs` isn't an option to preserve potential error messages.

paullatzelsperger:
> Sure then i would revert the last commit which adds the `tests.hookDeletePolicy` value in [7a832c7](https://github.com/eclipse-tractusx/tractusx-edc/commit/7a832c7fbe0253de8aebc361b9d5ae04f4f97f6f) to keep the value file more clean, right?
> 
> I thought removing the `--logs` isn't an option to preserve potential error messages.

if i understand this right we have two options: to not delete the test pods, or to not have logs, right?

fty4:
Right - here both options compared:

**1. Option: persist test-pods for test workflow** (current implementation of the PR)
With the last commit the workflow succeeds.
Here the Helm default will be overwritten to delete the test-pod after a successful run.
But because the workflow failed I have overwritten it again in test values back to the Helm defaults:

see: [`edc-tests/deployment/src/main/resources/helm/tractusx-connector-test.yaml`](https://github.com/mercedes-benz/eclipse-tractusx_tractusx-edc/blob/7a832c7fbe0253de8aebc361b9d5ae04f4f97f6f/edc-tests/deployment/src/main/resources/helm/tractusx-connector-test.yaml#L87-L88)

This would work and the pod will not be deleted in the test infrastructure which is no problem (will be deleted anyway).

> \+ Logs of failed pods in test workflow
> \- Additional values required in `values.yaml` and `tractusx-connector-test.yaml`

 
**2. Option: no logs for test workflow**
When removing the `--logs` flag the workflow would also succeed but in case of an error the log is not present in the workflow logs due the mentioned bug.

> \+ No additional values required
> \- No logs of failed pods in test workflow

<hr />

I think I would prefer option 1.

fty4:
> would that mean, if someone forgot to add those values to their `xyz-values.yml`, they'd still end up with zombie test pods?

No the test-pods will be removed because in the [default values.yaml
](https://github.com/mercedes-benz/eclipse-tractusx_tractusx-edc/blob/35931a252e02e440d70e66a066c3f88fee979490/charts/tractusx-connector/values.yaml#L572-L575) file the policy is set to `"helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded` which removes succeeded test-pods.
Failed will still be available - because you want to see the error messages here therefore no deletion.

If you want all to persist test-pods you can adapt to the [`tractusx-connector-test.yaml`](https://github.com/mercedes-benz/eclipse-tractusx_tractusx-edc/blob/7a832c7fbe0253de8aebc361b9d5ae04f4f97f6f/edc-tests/deployment/src/main/resources/helm/tractusx-connector-test.yaml#L87-L88) test values file.

Without this PR the test-pods will not be removed after successful testing.

paullatzelsperger:
> Without this PR the test-pods will not be removed after successful testing.

ok, that sounds reasonable, thanks for the clarification.

paullatzelsperger (approved):

PR #508, opened 2023-06-20
feat(build): only increase patch version on final releases

paullatzelsperger:
## WHAT

This PR improves the release workflow by changing the logic where the new snapshot version is selected.

Assuming `<major>.<minor>.<patch>` versions, previously, we always incremented the `patch` version, regardless of what the input was. So `0.5.0-rc1` would have resulted in a `0.5.1-SNAPSHOT`, which is incorrect.

Now, we only increase the `<patch>` version, if there is no `-rc1` or anything:
- `0.5.0` becomes `0.5.1-SNAPSHOT`
- `0.5.0-rc1` becomes `0.5.0-SNAPSHOT`

## WHY

The existing logic would have handled release candidates etc. incorrectly.

## FURTHER NOTES

### DIFF
--- a/.github/workflows/publish-new-release.yml
+++ b/.github/workflows/publish-new-release.yml
@@ -233,9 +233,18 @@ jobs:
           git checkout main && git merge -X theirs releases --no-commit --no-ff
 
           # Extract release version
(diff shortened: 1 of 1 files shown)


paullatzelsperger:
\cc @jimmarino

jimmarino (approved):

wolf4ood (approved):

PR #544, opened 2023-06-28
feat(SSI): adds E2E test for Catalog request with real MIW

### LINKED ISSUE #543: MIW E2E integration test
## Describe the solution you'd like

It would be nice to have some E2E tests that uses the real MIW 

Since setting up the MIW isn't trivial and seeding the VC requires more setup we could start by doing a CatalogTest.

On dev environment the MIW seed the database with the authority and BpnCredential and SummaryCredential.

We could use this participant to request the catalog to himself by using the MIW flow.

wolf4ood:
## WHAT

Adds an E2E test with real MIW running. The test is a single `CatalogRequest` with one participant asking the catalog to self.  

Due to the non-trivial seed of credentials in MIW for now it's ok just to have only this flow. 
It will still be tested the real communication with MIW, Keycloak and new policy based on credentials.

The used image for MIW is 

```
ghcr.io/catenax-ng/tx-managed-identity-wallets_miw_service:0.0.1-snapshot.2994d69
```

## WHY

Catching regressions 

Closes #543

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -172,7 +172,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Starting MIW, Keycloak and Postgres Servers
         run: |
-          docker-compose -f edc-tests/e2e-tests/src/test/resources/docker-compose.yml up -d 
+          docker compose -f edc-tests/e2e-tests/src/test/resources/docker-compose.yml up -d 
 
       - name: Run MIW E2E tests
         run: ./gradlew test -DincludeTags="MiwIntegrationTest"
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -160,7 +160,7 @@ jobs:
       - name: Run Postgresql E2E tests
         run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest"
 
-  miw-integration-tests:
+  ssi-integration-tests:
     runs-on: ubuntu-latest
     continue-on-error: true
     needs: [ verify-formatting, verify-license-headers ]
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -159,3 +159,20 @@ jobs:
 
       - name: Run Postgresql E2E tests
         run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest"
+
+  miw-integration-tests:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    needs: [ verify-formatting, verify-license-headers ]
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-java
+
+      - uses: actions/checkout@v2
+      - name: Starting MIW, Keycloak and Postgres Servers
+        run: |
+          docker-compose -f edc-tests/e2e-tests/src/test/resources/docker-compose.yml up -d 
+
+      - name: Run MIW E2E tests
+        run: ./gradlew test -DincludeTags="MiwIntegrationTest"
--- a/edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
+++ b/edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
@@ -137,7 +137,7 @@ public class MiwApiClientImpl implements MiwApiClient {
         if (valid) {
             return Result.success();
         } else {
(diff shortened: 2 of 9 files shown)


tuncaytunc-zf (approved):
I could successfully test it against the MIW

paullatzelsperger:
KICS is reporting a myriad of problems/warnings. since this is just used for testing, I think we can safely disregard and ignore them. Also, this solution will only stay in place for one release of C-X.

wolf4ood:
Yeah i saw all the warning but since it's only for testing we can ignore them i guess

edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
```diff
@@ -137,7 +137,7 @@ private Result<Void> handleVerifyResult(Map<String, Object> response) {
         if (valid) {
             return Result.success();
         } else {
-            return Result.failure(format("Verification failed with response: %s", response));
+            return Result.failure("MIW verification failed");
```
jimmarino, line 140: Can we provide more info on what failed?

edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tag/MiwIntegrationTest.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@IntegrationTest
+@Tag("MiwIntegrationTest")
+public @interface MiwIntegrationTest {
```
jimmarino, line 29: Doc?

edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/MiwSsiCatalogTest.java
```diff
@@ -0,0 +1,96 @@
...
+                put("tx.ssi.endpoint.audience", SOKRATES_DSP_CALLBACK);
+            }
+        };
+        var baseConfiguration = sokratesConfiguration();
+        ssiConfiguration.putAll(baseConfiguration);
+        return ssiConfiguration;
+    }
+
+    @Test
+    void requestCatalog_fulfillsPolicy_shouldReturnOffer() {
```
jimmarino, line 68: Doc on what this verifies?

jimmarino (approved):

edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
paullatzelsperger: should we? I mean, rephrasing it to `"VP verification failed"` is about the extent of info we want to emit, no? if the verification of a VP failed, there is a limited set of possibilities what could've gone wrong anyway. And even if the API did return the exact cause (which I _hope_ it doesn't), we should not leak any details on our side.

jimmarino: Ah no you are right, that is returned to the client. We may want to send the failure to the monitor though

.github/workflows/verify.yaml
```diff
@@ -159,3 +159,20 @@ jobs:
 
       - name: Run Postgresql E2E tests
         run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest"
+
+  miw-integration-tests:
```
paullatzelsperger, line 163:
```suggestion
  ssi-integration-tests:
```

paullatzelsperger (approved):
nothing much to add beyond what's already stated :)

edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
wolf4ood: There is no failure details actually, only valid or not. We can only know if it was the audiance verification that failed or the VP verification

ndr-brt (approved):

PR #564, opened 2023-07-05
fix(charts): add temporary emptyDir volume to the controlplane charts

wolf4ood:
## WHAT

It turned out that this https://github.com/eclipse-tractusx/tractusx-edc/discussions/556 was caused by the default security context that prevented the JSON-LD extensions to properly load the cached context from disks.

Logs:
```
WARNING  Failed to register cached json-ld document: Cannot read resource document/credential-v1.jsonld: 
WARNING  Failed to register cached json-ld document: Cannot read resource document/summary-vc-context-v1.jsonld: 
WARNING  Failed to register cached json-ld document: Cannot read resource document/summary-vc-context-v1.jsonld:
```

This can be fixed by allowing the setting the property in the securityContext `readOnlyRootFilesystem:false` so that the process can write in the `/tmp` directory. With this PR it will not be needed anymore since in all charts a temporary volume will be mounted for `/tmp` that will be writable.

A proper fix that avoids the temporary volume could be done in EDC upstream by not requiring a `File` for cached context

### DIFF
--- a/docs/samples/example-dataspace/plato-values.yaml
+++ b/docs/samples/example-dataspace/plato-values.yaml
@@ -54,9 +54,6 @@ dataplane:
     pullPolicy: Never
     tag: "latest"
     repository: "edc-dataplane-hashicorp-vault"
-  securityContext:
-    # avoids some errors in the log: cannot write temp files of large multipart requests when R/O
-    readOnlyRootFilesystem: false
   aws:
     endpointOverride: http://minio:9000
     secretAccessKey: qwerty123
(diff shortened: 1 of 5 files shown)


docs/samples/example-dataspace/plato-values.yaml
```diff
@@ -54,9 +54,6 @@ dataplane:
     pullPolicy: Never
     tag: "latest"
     repository: "edc-dataplane-hashicorp-vault"
-  securityContext:
-    # avoids some errors in the log: cannot write temp files of large multipart requests when R/O
-    readOnlyRootFilesystem: false
```
paullatzelsperger, line 59: can we please leave that in place? otherwise Jersey will vomit exceptions all over the log, because it tries to write a temp file

wolf4ood: That's the fix of temporary volume. It will fix also the Jersey exceptions :)

paullatzelsperger (approved):

PR #574, opened 2023-07-06
feat(MIWClient): adds response body in case errors in MIW response

### LINKED ISSUE #573: Improve error reporting on MIW responses
## Describe alternatives you've considered

As of now when some errors happens in the call to MIW we only report the response code as error, we should also include 
the body of the response if present.

wolf4ood:
## WHAT

Adds response body in case errors in MIW response

## WHY

Better diagnostics when dealing on errors in MIW

Closes #573

### DIFF
--- a/edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
+++ b/edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
@@ -164,17 +164,25 @@ public class MiwApiClientImpl implements MiwApiClient {
             var body = Objects.requireNonNull(response.body()).string();
             return Result.success(mapper.readValue(body, tr));
         } catch (IOException e) {
-            monitor.debug("Failed to parse response from MIW");
+            monitor.severe("Failed to parse response from MIW");
             return Result.failure(e.getMessage());
         }
     }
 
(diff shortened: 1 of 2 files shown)


edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
```diff
@@ -170,12 +170,20 @@ private <R> Result<R> handleSuccess(Response response, TypeReference<R> tr) {
...
+        var body = "";
+        if (response.body() != null) {
+            try {
+                body = response.body().string();
+            } catch (IOException e) {
+                monitor.debug("Failed to read response from MIW");
+                return Result.failure(e.getMessage());
+            }
+        }
+        var msg = format("MIW API returned %s with body: %s", response.code(), body);
```
tuncaytunc-zf, line 182: We should take care to not log any sensitive data in terms MIW returns some in message body.

tuncaytunc-zf (approved):

edc-extensions/ssi/ssi-miw-credential-client/src/main/java/org/eclipse/tractusx/edc/iam/ssi/miw/api/MiwApiClientImpl.java
wolf4ood: It should not i guess/hope 
but in case we can also not forward the body to the Result and just use that on the debug logging

PR #588, opened 2023-07-07
fix(test): converted some MockWebServer to long-running instances

paullatzelsperger:
## WHAT

this PR attempts to fix some of the flaky e2e tests we saw in the past.
The problem there was that the `MockWebServer` could not be started, due to an "Address already in use" error.

We don't exactly know where this is coming from, so this is really more of an "exploratory" PR

It should be noted that this goes somewhat _against_ the recommended use of the `MockWebServer`, which requires it to be re-instantiated on every test.

## WHY

Flaky e2e tests

## FURTHER NOTES

other solution ideas are
- switching to another mockwebserver. In EDC we use the Netty MockServer, which seems to not have these problems.
- adding a `RetryPolicy` to the instantiation and starting of the mock webserver: this would guard against transient errors such as timing issues
- use randomized ports for _every_ instance: would probably the cleanest solution, but might be difficult to implement, since all in-mem runtimes are instantiated only once, and would need to be re-instantiated.

 

wolf4ood (approved):
:crossed_fingers:

PR #619, opened 2023-07-17
feat: generate, merge and publish OpenAPI spec 2

### LINKED ISSUE #295: Setup and implement publishing OpenAPI files to SwaggerHub
## WHAT

Generate and publish OpenAPI specifications for all APIs provided by the Tractus-X code base to SwaggerHub.

## WHY
Documentation

## HOW
- have versioned api docs of upstream EDC (https://github.com/eclipse-edc/Connector/issues/2834)
- download the correct EDC api doc for a Tractus-x release
- generate Tractus-x specs (`./gradlew resolve && ./gradlew mergeApiDoc...`)
- merge both the EDC spec file and the TractusX spec file
- publish that merged file to SwaggerHub

bcronin90:
## WHAT

Generates OpenAPI spec for our extensions, merges it with the upstream docs and publishes it. Hopefully does this automatically and with the correct versions.

## WHY

#295

Closes #295


bcronin90:
This is a successful workflow run:
https://github.com/bcronin90/tractusx-edc/actions/runs/5611012302/jobs/10266762957

I assume the red X-s will go away once DEPENDENCIES are fixed and I rebase. What I cannot test is how the new workflow links up to publish and publish-release. Please review that with care. Thanks.

bcronin90:
Okay. Everything's green. Feel free to review.

.github/workflows/publish.yaml
```diff
@@ -58,8 +58,9 @@ jobs:
           [ ! -z "${{ secrets.ORG_GPG_PASSPHRASE }}" ] && 
           [ ! -z "${{ secrets.ORG_GPG_PRIVATE_KEY }}" ] && 
           [ ! -z "${{ secrets.ORG_OSSRH_USERNAME }}" ] && 
-          [ ! -z "${{ secrets.ORG_OSSRH_PASSWORD }}" ]  && 
-          echo "HAS_OSSRH=true" >> $GITHUB_OUTPUT
+          [ ! -z "${{ secrets.ORG_OSSRH_PASSWORD }}" ]  && echo "HAS_OSSRH=true" >> $GITHUB_OUTPUT
+          [ ! -z "${{ secrets.SWAGGERHUB_API_KEY }}" ]  && 
+          [ ! -z "${{ secrets.SWAGGERHUB_USER }}" ]  &&
```
paullatzelsperger, line 63: Pls add `echo "HAS_SWAGGER=true >> $GITHUB_OUTPUT` here.
Pls also add the `HAS_SWAGGER: ${{ steps.secret-presence.outputs.HAS_SWAGGER }}` to the job outputs (after L52), otherwise it won't be available

.github/workflows/publish-swaggerhub.yaml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/setup-node@v3
+
+      - name: Install tools
+        run: |
+          npm i -g swaggerhub-cli
+          sudo apt install -y crudini
+
+      - name: Extract versions
+        run: |
+          export DOWNSTREAM_VERSION=$(crudini --get ./gradle.properties '' version)
```
paullatzelsperger, line 56: no need for `crudini`, just use 
```bash 
grep "version" gradle.properties  | awk -F= '{print $2}')
```

.github/workflows/publish.yaml
```diff
@@ -126,4 +127,11 @@ jobs:
...
           echo "Publishing Version $VERSION to Sonatype"
-          ./gradlew publishToSonatype ${cmd} --no-parallel -Pversion=$VERSION -Psigning.gnupg.executable=gpg -Psigning.gnupg.passphrase="${{ secrets.ORG_GPG_PASSPHRASE }}"
\ No newline at end of file
+          ./gradlew publishToSonatype ${cmd} --no-parallel -Pversion=$VERSION -Psigning.gnupg.executable=gpg -Psigning.gnupg.passphrase="${{ secrets.ORG_GPG_PASSPHRASE }}"
+
+  publish-to-swaggerhub:
+    name: "Publish OpenAPI spec to Swaggerhub"
+    permissions:
+      contents: read
+    needs: [ secret-presence ]
```
paullatzelsperger, line 136: please add the 
```yaml
if: needs.secret-presence.output.HAS_SWAGGER
```
otherwise this job will always run. `needs` only expresses dependency and ordering.

.github/workflows/publish-swaggerhub.yaml
paullatzelsperger, line 56: I'd prefer having the `DOWNSREAM_VERSION` passed as optional workflow input, and only read it from `gradle.properties` if the input is not present.

paullatzelsperger (changes requested):
The changes in the `Dockerfiles` should be reverted - the change is already in `main`. You'll have it after rebasing.

Other comments inline.

.github/workflows/publish-swaggerhub.yaml
bcronin90: I considered that. But that approach is not aware of sections (and quotes, but that's less important), which our .ini files use. I know it's one more apt call, but it's worth it.

bcronin90:
Once again, successful run here: https://github.com/bcronin90/tractusx-edc/actions/runs/5678365723

I just needed to remove the push trigger afterwards.

.github/workflows/publish-swaggerhub.yaml
paullatzelsperger: Sections are not relevant for the task at hand, neither is writing, updating or deleting lines from the file.
We are not using 99% of what that tool offers, and installing random tools for a task that can be solved with standard tools is *not* the way to go.
Thats not even mentioning all the issues third-party tools have in general (api changes, lifecycle, potential security holes,...).

Please get rid of `crudini` and use standard CLI tools.

bcronin90: True, that is EDC policy. I'll do it with preinstalled tools instead.

bcronin90:
Newest version, before the rebase, but still: https://github.com/bcronin90/tractusx-edc/actions/runs/5679954414

paullatzelsperger:
@bcronin90 the tests fail, because the cUrl version has apparently changed on Alpine. Could you please replace `8.2.0-r1`->`8.2.1-r0` in all Dockerfiles?

This has nothing to do with your PR, the problem is that Alpine doesn't keep old cUrl versions around.

#686 should fix this problem in the future.

PR #620, opened 2023-07-17
feat(EDR): adds EDR state machine for handling EDR renewal

### LINKED ISSUE #600: Auto renewal of EDR tokens
## Describe the solution you'd like

In the current state with the custom API around EDR, we are able with one request to negotiation and cache the EDR on our side in an EDR/cache store. If the EDR expire people should re-fire the EDR negotiation from scratch to obtain a new token.
It would be cool if we implement an auto renewal mechanism.
  
The implementation should:

- Check the EDR store/cache for expired or about to expire EDR.
- Fire a new transfer process for getting a new EDR
- Store the new EDR
- Clean the expired one

wolf4ood:
## WHAT

Implements a state machine for EDR renewal. Once an EDR is first negotiated the state machine will boostrap the
cached entry with a `Negotiated` state.

The state machine is composed by the following states:

- `Negotiated` the primordial state. The EDR has been negotiated with the provided and stored in the EDR store
- `Refreshing` the EDR is about to expire and another transfer process has been started for negotiating another `EDR` token with the same parameters
- `Expired`: Once a new EDR is negotiated the old one is marked as `Expired` waiting for deletion
- `Deleting`: Once the the retention period is expired EDR are marked as deleting

The EDR cached entry gets deleted  by picking from the state machine the one that are in `Deleting` state.

## WHY

User experience. Once the first Negotiate EDR request has been fulfill the EDR extension will ensure to renew automatically the EDR token before it expires. 

## FURTHER NOTES

Also in this PR the E2E test based on PostgreSQL has been moved to testcontainers. When wi will upgrade to EDC > 0.1.3 
we could also leverage to the refactored `PostgresqlStoreSetupExtension` that uses testcontainers as well for testing the EDR storage. 

Also in this PR for supporting testcontainers I've created a `TxPostgresqlLocalInstance` copy of the upstream one. It's mark for removal once we upgrade to the next EDC version.

Closes #600

### DIFF
--- a/core/edr-cache-core/src/main/java/org/eclipse/tractusx/edc/edr/core/defaults/InMemoryEndpointDataReferenceCache.java
+++ b/core/edr-cache-core/src/main/java/org/eclipse/tractusx/edc/edr/core/defaults/InMemoryEndpointDataReferenceCache.java
@@ -14,6 +14,8 @@
 
 package org.eclipse.tractusx.edc.edr.core.defaults;
 
+import org.eclipse.edc.spi.entity.StatefulEntity;
+import org.eclipse.edc.spi.persistence.Lease;
 import org.eclipse.edc.spi.query.Criterion;
 import org.eclipse.edc.spi.query.QuerySpec;
 import org.eclipse.edc.spi.result.StoreResult;
@@ -24,18 +26,22 @@ import org.eclipse.tractusx.edc.edr.spi.types.EndpointDataReferenceEntry;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.time.Clock;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
 
 import static java.util.Collections.emptyList;
+import static java.util.Comparator.comparingLong;
 import static java.util.stream.Collectors.toList;
 import static org.eclipse.edc.spi.result.StoreResult.notFound;
 import static org.eclipse.edc.spi.result.StoreResult.success;
@@ -44,20 +50,33 @@ import static org.eclipse.edc.spi.result.StoreResult.success;
  * An in-memory, threadsafe implementation of the cache.
  */
 public class InMemoryEndpointDataReferenceCache implements EndpointDataReferenceCache {
+    private static final long DEFAULT_LEASE_TIME_MILLIS = 60_000;
     private final LockManager lockManager;
-
     private final EdrCacheEntryPredicateConverter predicateConverter = new EdrCacheEntryPredicateConverter();
 
     private final Map<String, List<EndpointDataReferenceEntry>> entriesByAssetId;
 
     private final Map<String, EndpointDataReferenceEntry> entriesByEdrId;
+
     private final Map<String, EndpointDataReference> edrsByTransferProcessId;
+    private final String lockId;
+
+    private final Map<String, Lease> leases;
+
+    private final Clock clock;
 
     public InMemoryEndpointDataReferenceCache() {
+        this(UUID.randomUUID().toString(), Clock.systemUTC(), new ConcurrentHashMap<>());
+    }
+
+    public InMemoryEndpointDataReferenceCache(String lockId, Clock clock, Map<String, Lease> leases) {
+        this.lockId = lockId;
         lockManager = new LockManager(new ReentrantReadWriteLock());
         entriesByAssetId = new HashMap<>();
         entriesByEdrId = new ConcurrentHashMap<>();
         edrsByTransferProcessId = new HashMap<>();
+        this.leases = leases;
+        this.clock = clock;
     }
 
     @Override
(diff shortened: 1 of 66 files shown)


core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/service/EdrServiceImpl.java
```diff
@@ -15,48 +15,40 @@
...
+    public EdrServiceImpl(EdrManager edrManager, EndpointDataReferenceCache endpointDataReferenceCache) {
+        this.edrManager = edrManager;
         this.endpointDataReferenceCache = endpointDataReferenceCache;
     }
 
     @Override
     public ServiceResult<ContractNegotiation> initiateEdrNegotiation(NegotiateEdrRequest request) {
-        var contractNegotiation = contractNegotiationService.initiateNegotiation(createContractRequest(request));
-        return ServiceResult.success(contractNegotiation);
+        var contractNegotiation = edrManager.initiateEdrNegotiation(request);
```
ndr-brt, line 46: Lately, I'thinking that the "sync" operations should be made directly in the service without passing through the manager that should only be the running state machine. In fact also in the `ContractNegotiation` and `TransferProcess` services, the managers are only used to initiate the processes, but I think I will inline them the next time I will pass there, or switch them to command/command handler. At that point the "*Manager" could be called "*StateMachine". (just raving around a little :grin: )

core/edr-core/src/test/java/org/eclipse/tractusx/edc/edr/core/EdrCoreExtensionTest.java
```diff
@@ -42,9 +43,9 @@ void setUp(ObjectFactory factory, ServiceExtensionContext context) {
     @Test
     void shouldInitializeTheExtension(ServiceExtensionContext context) {
         extension.initialize(context);
-        
-        var service = extension.adapterTransferProcessService();
-        assertThat(service).isInstanceOf(EdrServiceImpl.class);
+
+        var service = context.getService(EdrManager.class);
```
ndr-brt, line 47: it can be passed through method parameters (it's injected by the DependencyInjectionExtension

core/edr-core/src/test/java/org/eclipse/tractusx/edc/edr/core/EdrCoreServiceExtensionTest.java
```diff
@@ -0,0 +1,50 @@
...
+@ExtendWith(DependencyInjectionExtension.class)
+public class EdrCoreServiceExtensionTest {
+
+    EdrCoreServiceExtension extension;
+
+    @BeforeEach
+    void setUp(ObjectFactory factory, ServiceExtensionContext context) {
+        context.registerService(EdrManager.class, mock(EdrManager.class));
+        context.registerService(EndpointDataReferenceCache.class, mock(EndpointDataReferenceCache.class));
+        extension = factory.constructInstance(EdrCoreServiceExtension.class);
```
ndr-brt, line 39: not needed anymore, you can pass the `EdcCoreServiceExtension` as a method parameter in the test directly

edc-extensions/edr/edr-cache-sql/src/test/java/org/eclipse/tractusx/edc/edr/store/sql/PostgresqlTransactionalStoreSetupExtension.java
```diff
@@ -46,14 +49,17 @@
 public class PostgresqlTransactionalStoreSetupExtension implements BeforeEachCallback, AfterEachCallback, BeforeAllCallback, ParameterResolver {
 
     private final String datasourceName;
+    private final QueryExecutor queryExecutor;
+    private final Monitor monitor = mock(Monitor.class);
```
ndr-brt, line 53: just a nit: it's not necessary to pass the class to the `mock()` method if the type is explicitly declared

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/manager/EdrManagerImpl.java
```diff
@@ -0,0 +1,368 @@
...
+                    .onRetryExhausted((n, failure) -> transitionToError(n, format("Failed delete EDR token: %s", failure.getFailureDetail())))
+                    .execute("Start an EDR token deletion");
+        } else {
+            breakLease(edrEntry);
+            return false;
+        }
+    }
+
+    private StatusResult<Void> deleteEntry(EndpointDataReferenceEntry entry) {
+        this.transitionToDeleted(entry);
```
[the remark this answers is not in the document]
wolf4ood: Actually this is tricky :D, i cannot the delete directly the thing because it's under lease, so i needed i way to transition
 to a state that's not picked by the state machine, which is then safe to delete. Rethinking this probably i will do a custom delete method in the store interface for being able to "unsafe" delete  stuff

core/edr-cache-core/src/main/java/org/eclipse/tractusx/edc/edr/core/defaults/InMemoryEndpointDataReferenceCache.java
```diff
@@ -44,45 +50,70 @@
...
+
+    private final Map<String, Lease> leases;
+
+    private final Clock clock;
 
     public InMemoryEndpointDataReferenceCache() {
+        this(UUID.randomUUID().toString(), Clock.systemUTC(), new ConcurrentHashMap<>());
+    }
+
+    public InMemoryEndpointDataReferenceCache(String lockId, Clock clock, Map<String, Lease> leases) {
```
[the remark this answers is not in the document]
wolf4ood: I though about that, but i needed to rewrite the cache basically :), but i totally make sense.

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/service/EdrServiceImpl.java
wolf4ood: Will open an issue about this. In fact it was inlined previously :)

paullatzelsperger (approved):

paullatzelsperger:
deployment tests will work again after https://github.com/eclipse-tractusx/tractusx-edc/pull/633 is merged, and this PR is rebased onto `main`

core/edr-cache-core/src/main/java/org/eclipse/tractusx/edc/edr/core/defaults/InMemoryEndpointDataReferenceCache.java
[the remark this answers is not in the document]
ndr-brt: @wolf4ood maybe we should think about moving it up into some library module (util?)

PR #630, opened 2023-07-19
chore(MIWApiClient): resolve holderIdentifier from participantId

### LINKED ISSUE #629: MIW Client Extension: Use participantId as holderIdentifier if present.
**Is your feature request related to a problem? Please describe.**
Actually the relation between "holderIdentifier" and "ParticipantId" (connector BPN) is made within the MIW, but can also be provided by the MIW Api Client directly.

We can use the "participantId" if it is present in the actual context, otherwise fallback to the MIW side resolution again.
From my understanding "participantId" is anyway a mandatory setting for the connector used by multiple extensions.

**Describe the solution you'd like**
Within the MIW Api Client the "holderIdentifier" can be resolved from the participantId provided by the actual context.

domreuter:
## WHAT

refer to #629 

## WHY

refer to #629

Closes #629


paullatzelsperger:
before I even look at this PR or anything, please be sure to read, understand and comply with the pr etiquette. simply referring to the issue in the PR desc is not acceptable.

PR #678, opened 2023-07-28
feat: simplify data encryptor

### LINKED ISSUE #672: Simplify the `DataEncrypter`
## WHAT

The implementation of the `DataEncryter` - actually the entire `edc-extensions/data-encryption` module should be greatly simplified. There is no need for a custom implementation of AES encryption, all that's needed is already shipped with Java

## WHY
Simplification. 

## HOW
basically this: https://www.baeldung.com/java-aes-encryption-decryption

paullatzelsperger:
## WHAT

Replaces the old, quite convoluted AES data encryptor extension with a sleeker one.
In addition, the bouncy castle provider was replaced with a plain JCE implementation.

## WHY

Simplification of code. We don't need layers on layers of factories and providers. KISS.

## FURTHER NOTES

- Currently only AES/GCM with a 128 bit IV is supported. A `NoopEncryptor` is instantiated as fallback.
- Postgres test setup was improved slightly

Closes #672


jimmarino (approved):
🪓

ndr-brt (approved):
:axe: :axe: :axe:

PR #679, opened 2023-07-29
chore: run omejdn as nonRoot in chart tractusx-connector-legacy

fty4:
## WHAT

The omejdn / daps server should not be executed with the user root.
Therefore the user nobody (65534) will be set.

Also the securityContext will only be set if configured in the values file.

## WHY

Otherwise deploying the chart will fail if root is permited.

## FURTHER NOTES

This is a requirement for kyverno worflow introduced in eclipse-tractusx/e2e-testing#28.

<hr />


paullatzelsperger:
@fty4 generally the same applies here as for #683: technically we can merge this, but the legacy chart is very likely to get removed in one of the next releases.

PR #683, opened 2023-07-31
fix(helm): remove legacy suffix from legacy Helm chart in cp

fty4:
## WHAT

Currently a wrong / non existent image will be used in the legacy Helm chart.

## WHY

This PR changes the correct image to be choosen:

`tractusx/edc-controlplane-postgresql-hashicorp-vault-legacy`-> `tractusx/edc-controlplane-postgresql-hashicorp-vault-legacy`

Only the `-legacy` suffix will be removed.
Otherwise a ImagePullBackOff error will occur.

<hr />

### DIFF
--- a/charts/tractusx-connector-legacy/templates/deployment-controlplane.yaml
+++ b/charts/tractusx-connector-legacy/templates/deployment-controlplane.yaml
@@ -64,7 +64,7 @@ spec:
           {{- if .Values.controlplane.image.repository }}
           image: "{{ .Values.controlplane.image.repository }}:{{ .Values.controlplane.image.tag | default .Chart.AppVersion }}"
           {{- else }}
-          image: "tractusx/edc-controlplane-postgresql-hashicorp-vault-legacy:{{ .Values.controlplane.image.tag | default .Chart.AppVersion }}"
+          image: "tractusx/edc-controlplane-postgresql-hashicorp-vault:{{ .Values.controlplane.image.tag | default .Chart.AppVersion }}"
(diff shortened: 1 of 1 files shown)


paullatzelsperger (changes requested):
this will **not** work, because the `edc-controlplane-postgres-hashicorp-vault` image will not have DAPS dependencies, only SSI ones.

The correct way to fix this _would_ be to update the publish workflow and include a docker image for the legacy image. 
However, the legacy chart is scheduled for deletion, so I wouldn't bother updating it any further.

@fty4 if you want to contribute, one way to do it would be to remove the legacy from the code base. Otherwise pls close this PR.

fty4:
This means the in `0.5.0` introduced legacy-chart was never working?
Why was it then created?

Would it be an option to pin the image to a older version ([`0.4.1`](https://github.com/eclipse-tractusx/tractusx-edc/releases/tag/0.4.1))?

paullatzelsperger:
No point in pinning this to anything, because the issue is not the image path in the chart template, the issue that the docker image was never created and published, which admittedly was an oversight on our part.

The intention was to give people somewhat of a transition period, while DAPS was still running, which it isn't anymore. 
They would have had to use a locally-built image.

In addition, the legacy chart was never actually tested, which is explicitly stated in the migration guide.
Short of deleting it, I don't see any reason to fiddle with it.

fty4:
Alright - then it seems to be too much effort to get this chart running.
I will raise another PR to delete this non functional chart.

PR #687, opened 2023-08-02
feat: improve bpn validation

### LINKED ISSUE #673: Improve the `business-partner-validation` extensions
## WHAT
The current implementation of the BPN validation simply validates a BPN against a hard-coded list of BPNs stored on a policy. While this works well for testing, it is not scalable in a production environment, because
- every contract definition is explicitly bound to specific BPNs
- changing intended recipients of a data offer requires updating the policy
- the same function is bound for `Duty`, `Permission`, `Prohibition`, where it only should be `Permission`

## WHY
Inflexible and actually wrong implementation of BPN validation

## HOW
Add a new extension, that performs the validation not against a hard-coded list, but rather reads that list from a database table. This table can be manipulated via a new API endpoint. Thus the following implementation items are necessary:
- In-Mem implementation
- Postgres implementation
- REST controller
- a new Policy evaluation function "BPN in getListFromDatabase()", 
- deprecate but don't delete the old extension

paullatzelsperger:
## WHAT

Improves the BPN validation in the following aspects:
- BPNs are not hardcoded on the policy anymore
- BPNs are stored in a small persistence backend (in-mem + postgres are available)
- the evaluation function understands a few more operators

## WHY

The old BPN validation extension was inflexible and not scalable and there were technical errors in it.

## FURTHER NOTES

- For compatibility reasons the old policy evaluation will remain in place but issue a deprecation warning.
- I unimaginatively called the new module `business-partner-validation2` - suggestions welcome
- the new module isn't yet used anywhere - e2e tests, runtimes, docker images, helm charts... that'll come in another PR
- [edit] a REST API to manipulate BPN+groups will follow in another PR
 
Closes #673


ndr-brt (approved):
maybe `bpn-validation` could be good for naming?

wolf4ood (approved):

wolf4ood:
:+1:  for `bpn-validation`

edc-extensions/business-partner-store-sql/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/store/sql/BusinessPartnerGroupStatements.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Contributors:
+ *       Bayerische Motoren Werke Aktiengesellschaft (BMW AG) - initial API and implementation
+ *
+ */
+
+package org.eclipse.tractusx.edc.validation.businesspartner.store.sql;
+
```
jimmarino, line 16: Documentation

edc-extensions/business-partner-validation2/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/BusinessPartnerEvaluationExtension.java
```diff
@@ -0,0 +1,79 @@
...
+ *     <li>{@link org.eclipse.edc.connector.contract.spi.offer.ContractDefinitionResolver#CATALOGING_SCOPE}</li>
+ *     <li>{@link org.eclipse.edc.connector.contract.spi.validation.ContractValidationService#NEGOTIATION_SCOPE}</li>
+ *     <li>{@link org.eclipse.edc.connector.contract.spi.validation.ContractValidationService#TRANSFER_SCOPE}</li>
+ * </ul>
+ * The rule to which the function is bound is {@link BusinessPartnerEvaluationExtension#BUSINESS_PARTNER_CONSTRAINT_KEY}. That means, that policies that are bound to these scopes look
+ * like this:
+ * <pre>
+ * {
+ *     "constraint": {
+ *         "leftOperand": "BusinessPartnerGroup",
```
jimmarino, line 43: We should be careful here and namespace the operand value because ODRL defines leftOperand as an IRI. This will result in expansion either placing it in the vocabulary specified by a prefix or, if one is not specified, in the default EDC namespace.

jimmarino (changes requested):
Looks great so far. Two things:

1. We should have a DR for this
2. The left operand should have a prefix/namespace as explained in the one comment

edc-extensions/business-partner-validation2/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/BusinessPartnerEvaluationExtension.java
paullatzelsperger: wouldn't the default namespace be the TX one (`tx:https://w3id.org/tractusx/v0.0.1/ns/`)?

paullatzelsperger:
> Looks great so far. Two things:
> 
> 1. We should have a DR for this
> 2. The left operand should have a prefix/namespace as explained in the one comment

PR incoming.

jimmarino (approved):

edc-extensions/business-partner-validation2/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/BusinessPartnerEvaluationExtension.java
jimmarino: not sure if it is C-X or T-X, I always get confused. @wolf4ood

paullatzelsperger: currently, we have this: 
```java
    public static final String TX_PREFIX = "tx";
    public static final String TX_NAMESPACE = "https://w3id.org/tractusx/v0.0.1/ns/";
```

PR #688, opened 2023-08-03
feat: add API for the BPN validation extension

### LINKED ISSUE #673: Improve the `business-partner-validation` extensions

paullatzelsperger:
## WHAT

Adds a (JSON-LD) REST API for the BusinessPartnerNumber extension that was implemented in #673 
- the module `edc-extensions/bpn-validation` was restructured:
  - `bpn-validation-spi`: contains interfaces
  - `bpn-validation-core`: contains implementations and core clases
  - `bpn-validation-api`: contains API + controller
  - `business-partner-store-sql`: contains PostgreSQL implementation for the store.

Using the policy in e2e tests, docker images etc. will come in another PR.
## WHY

Clients need to be able to manipulate/seed their BPN groups

## FURTHER NOTES

- API uses JSON-LD to provide consistency
- integrating these modules in deliverables and more e2e tests will come in a subsequent PR

Closes #673


wolf4ood (approved):
LGTM!

PR #703, opened 2023-08-11
feat(edrs): add init edr request api validator

### LINKED ISSUE #702: feat: implement EDR API validator
Add `JsonObject` validation in the EDR APIs with upstream helpers

REF: https://github.com/eclipse-edc/Connector/issues/3011

wolf4ood:
## WHAT

Adds validator when initiating a new EDR negotiation request

## WHY

consistency

## FURTHER NOTES

In all EDRs api the call to expand/compact where removed, since there is the interceptor that does that already

Closes #702


paullatzelsperger (approved):
🔥

PR #707, opened 2023-08-12
feat(helm): add customCaCerts value

fty4:
## WHAT

Add Helm chart value (`customCaCerts`) to support adding additional custom ca-certs to the control- and dataplane pods.

## WHY

If your company is using a certificate authority which is not inside the common `/opt/java/openjdk/lib/security/cacerts` file connections will not be trusted.
A connection in between will then fail in lack of trusted certificates.
When only using public signed certificate you will not need this value.

## HOW

With this change you will be able to specify additional custom certificates as follows in your values file:

```yaml
customCaCerts:
  YourCompany.pem: |
    -----BEGIN CERTIFICATE-----
    MIICuzCCAaMCAQAwMzELMAkGA1UEBhMCREUxEDAOBgNVBAgMB0JhdmFyaWExEjAQ
    ...
    -----END CERTIFICATE-----
```

When one or more certificates are added to the dict they will be each added via a initContainer and emptyDir to both cp and dp.

To view the added certificates you can run the keytool with following command:

`keytool -list -keystore /opt/java/openjdk/lib/security/cacerts -storepass changeit | grep -i your-cert-name`

<hr />

### DIFF
--- a/charts/tractusx-connector-azure-vault/README.md
+++ b/charts/tractusx-connector-azure-vault/README.md
@@ -164,6 +164,7 @@ helm install my-release tractusx-edc/tractusx-connector-azure-vault --version 0.
 | controlplane.url.protocol | string | `""` | Explicitly declared url for reaching the dsp api (e.g. if ingresses not used) |
 | controlplane.volumeMounts | list | `[]` | declare where to mount [volumes](https://kubernetes.io/docs/concepts/storage/volumes/) into the container |
 | controlplane.volumes | list | `[]` | [volume](https://kubernetes.io/docs/concepts/storage/volumes/) directories |
+| customCaCerts | object | `{}` | Add custom ca certificates to the truststore |
 | customLabels | object | `{}` | To add some custom labels |
 | dataplane.affinity | object | `{}` |  |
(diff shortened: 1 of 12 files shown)


mbctec (approved):

paullatzelsperger (commented):
Seems to work fine, but is there any way to test this? For example could there be a helm test, that executes the `keytool` command via a remote shell, and verifies that the cert is in the resulting list?
Don't know if it's worth it though, just a thought.

fty4:
Testing is always a good thing - but I see some challenges here.
If it will be implemented as a Helm test a customCaCerts must be added on the installation then the test could verify if this specific certificate was added.
The problem here would be that a Helm test will fail on every other executions (when someone else is testing the edc installation without the test-certificate). So a generic test will always fail.

Another way would be to implement a separate test via e.g. CI which might be a little bit overwhelming here.

paullatzelsperger:
> Testing is always a good thing - but I see some challenges here.
If it will be implemented as a Helm test a customCaCerts must be added on the installation then the test could verify if this specific certificate was added.
The problem here would be that a Helm test will fail on every other executions (when someone else is testing the edc installation without the test-certificate). So a generic test will always fail.

I see your point. 

> Another way would be to implement a separate test via e.g. CI which might be a little bit overwhelming here.

Maybe it could be as easy as opening a shell to the running container, and execute the `keytool` command? We'd have to effectively duplicate out test runs, and generate Certs beforehand, but it seems doable. Definitely should be done in another issue though.

PR #713, opened 2023-08-17
fix: add binding to tx namespace for credential policy

### LINKED ISSUE #616: Policy Constraint `leftOperand` does not match registered functions
## Description

The Policy Constraint `leftOperand` does not match the registered functions.

Consider the following `Policy` example:

```json
{
  "@context": {
    "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
  },
  "@type": "PolicyDefinition",
  "@id": "{{policy-id'}}",
  "policy": {
    "@context": "http://www.w3.org/ns/odrl.jsonld",
    "permission": {
      "action": "use",
      "constraint": {
        "leftOperand": "Dismantler",
        "operator": "eq",
        "rightOperand": "active"
      }
    }
  }
}
```

The value of `leftOperand` will be expanded to:

```json
{
  "http://www.w3.org/ns/odrl/2/leftOperand": [
    {
      "@id": "https://w3id.org/edc/v0.0.1/ns/Dismantler"
    }
  ]
}
```

Eventually, it will be deserialized to `Constraint.leftOperand` as `"https://w3id.org/edc/v0.0.1/ns/Dismantler"`. However, this value does not match the registered value of `"Dismantler"` in the `RuleBindingRegistry`. As a result, the `Constraint` is being filtered out from the Policy.

To resolve this issue, you can take one of the following approaches:

- Change the function keys in their registration by prefixing them with EDC's namespace, for example, `"https://w3id.org/edc/v0.0.1/ns/Dismantler"`.
- Expand `leftOperand` as a primitive `@value` instead of a unique instance identifier. For example:

```json
{
  "leftOperand": {
    "@value": "Dismantler"
  }
}
```

It will be deserialized to `Constraint.leftOperand` "Dismantler", matching the function key.

The same registering pattern is followed by all the other functions, except `inForceDate`, which is registered with the EDC namespace prefix (https://w3id.org/edc/v0.0.1/ns/inForceDate).

## Context Informations

- Tractus-x version: 0.5.0

wolf4ood:
## WHAT

Add fix for `leftOperand` when using `@context` of odrl which get expanded with wrong namespace.

to fix this the functions are registered also in the `tx` namespace and the final usage will be 

```json
{
  "@context": {
    "@vocab": "https://w3id.org/edc/v0.0.1/ns/",
    "tx" : "https://w3id.org/tractusx/v0.0.1/ns/"
  },
  "@type": "PolicyDefinition",
  "@id": "{{policy-id'}}",
  "policy": {
    "@context": "http://www.w3.org/ns/odrl.jsonld",
    "permission": {
      "action": "use",
      "constraint": {
        "leftOperand": "tx:Dismantler",
        "operator": "eq",
        "rightOperand": "active"
      }
    }
  }
}
```

## WHY

bugfix

## FURTHER NOTES

In addition in order to simplify policy expressions 2 context where drafted and added in the cache:

- tractusx 
- edc 

This two for the moment are just an exploration, once we decided that they works well we can proceed to upstream the
edc one and fetch the two persistent id in `https://w3id.org`

Here an example on how it will look like with `@context`

```json
{
  "@context": [
    "https://w3id.org/edc/v0.0.1",
    "https://w3id.org/tractusx/edc/v0.0.1",
    "http://www.w3.org/ns/odrl.jsonld"
  ],
  "@type": "PolicyDefinitionRequest",
  "@id": "${POLICY_ID}",
  "policy": {
    "@type": "Set",
    "permission": [
      {
        "action": "use",
        "constraint": [
          {
            "leftOperand": "Dismantler",
            "operator": "eq",
            "rightOperand": "active"
          }
        ]
      }
    ]
  }
}

```

Closes #616

### DIFF
--- a/core/json-ld-core/build.gradle.kts
+++ b/core/json-ld-core/build.gradle.kts
@@ -17,6 +17,7 @@ plugins {
 }
 
 dependencies {
+    implementation(project(":spi:core-spi"))
     implementation(libs.edc.spi.core)
     implementation(libs.edc.spi.jsonld)
     testImplementation(testFixtures(libs.edc.junit))
--- a/core/json-ld-core/src/main/java/org/eclipse/tractusx/edc/jsonld/JsonLdExtension.java
+++ b/core/json-ld-core/src/main/java/org/eclipse/tractusx/edc/jsonld/JsonLdExtension.java
@@ -29,10 +29,15 @@ import java.util.Map;
 
 import static java.lang.String.format;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.EDC_CONTEXT;
+import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.TX_CONTEXT;
+import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.TX_NAMESPACE;
+import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.TX_PREFIX;
 
 public class JsonLdExtension implements ServiceExtension {
 
     public static final String CREDENTIALS_V_1 = "https://www.w3.org/2018/credentials/v1";
+
     public static final String CREDENTIALS_SUMMARY_V_1 = "https://w3id.org/2023/catenax/credentials/summary/v1";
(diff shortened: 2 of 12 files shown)


wolf4ood:
\cc @yurimssilva

PR #715, opened 2023-08-18
Docs/docusaurus publishing

bcronin90:
## WHAT

Add a workflow to compile and publish the docusaurus docs. This will go hand in hand with a PR on their side to simplify our part of the sidebar.

## WHY

Our current process was a bit messy and a lot error prone. This should be less so.
Changes in .md files are either to fix linting issues or just for general cleanup.

## NOTE
Currently, the publishing part doesn't work yet. I'll get to that once I'm back.


bcronin90:
Successful run: https://github.com/bcronin90/tractusx-edc/actions/runs/6119485972/job/16609501339

bcronin90:
Need to figure out why the tests are failing. But this should be ready for review otherwise.

.github/workflows/publish-swaggerhub.yaml
```diff
@@ -61,39 +50,38 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
 
-      - name: Install Swagger CLI
+      - name: Install tools
```
paullatzelsperger, line 53: It seems you have not rebased onto `main` before working on this. Please revert all your changes of `publish-swaggerhub.yaml` and `publish.yaml` to the current `main` branch.

edc-controlplane/edc-controlplane-postgresql-hashicorp-vault/notice.md
```diff
@@ -2,24 +2,24 @@
 
 An EDC Control Plane using PostgreSQL as persistence backend, and HashiCorp Vault as secret store.
 
-DockerHub: https://hub.docker.com/r/tractusx/edc-controlplane-postgresql-hashicorp-vault
+DockerHub: <https://hub.docker.com/r/tractusx/edc-controlplane-postgresql-hashicorp-vault>
```
paullatzelsperger, line 5: I'm assuming these changes are required by docasaurus? If so, please note this in the PR description.

edc-extensions/dataplane-proxy/edc-dataplane-proxy-provider-api/src/main/java/org/eclipse/tractusx/edc/dataplane/proxy/provider/api/DataPlaneProxyProviderApiExtension.java
```diff
@@ -74,6 +94,19 @@ public String name() {
 
     @Override
     public void initialize(ServiceExtensionContext context) {
+        var port = context.getSetting(PROVIDER_PORT, DEFAULT_PROVIDER_PORT);
```
paullatzelsperger, line 97: why are these changes necessary? Your PR is dealing with packaging for - and publishing to - docasaurus. Why are code-level changes needed here? 
And if they really are necessary they should be clearly explained in the PR description. 

In any event, the `CONSUMER_CONFIG_KEY` and `CONSUMER_API_ALIAS` constants are undefined.

.github/workflows/publish-docusaurus.yaml
```diff
@@ -0,0 +1,115 @@
...
+          cp ./docs/development/coding-principles.md ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/02_coding_principles.md
+          cp ./pr_etiquette.md ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/03_pr_etiquette.md
+          cp ./styleguide.md ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/04_styleguide.md
+          mkdir -p ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/resources
+          cp ./resources/tx-checkstyle-config.xml ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/resources
+          cp ./SECURITY.md ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/05_security.md
+          mkdir -p ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/resources
+          cp ./resources/save_actions_scr.png ${{ runner.temp }}/tractusx-edc/Connector\ Kit/Documentation/resources/
+          
+          swaggerhub api:get ${{ env.SWAGGERHUB_USER }}/tractusx-edc/${{ env.DOWNSTREAM_VERSION }} > ${{ runner.temp }}/tractusx-edc/tractusx-edc-${{ env.DOWNSTREAM_VERSION }}.yaml
```
paullatzelsperger, line 106: generally, I recommend using `curl` to download the yaml, because using the Swaggerhub CLI fails when downloading from other APIs (e.g. `eclipse-edc`). The logged-in user / token is not valid then. Granted, that's not the case here, but using plain `curl` also avoids having to install the Swaggerhub CLI, which saves some runner execution time.

.github/workflows/publish-new-release.yml
```diff
@@ -249,3 +249,14 @@ jobs:
     with:
       downstream-version: ${{ needs.release-version.outputs.RELEASE_VERSION }}
     secrets: inherit
+
+  publish-docusaurus:
+    name: Publish docusaurus docs
+    permissions:
+      contents: write
+      packages: write
+    needs: [ secret-presence, publish-to-swaggerhub ]
```
paullatzelsperger, line 258: haven't tried this, but you probably want to also add the `github-release` job to the `needs` list, otherwise you could potentially reference a `downstream-version`, for which no GH Release exists yet.

.github/workflows/publish-docusaurus.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+      - name: "Install tools"
+        run: |
+          npm i -g swaggerhub-cli markdownlint-cli2 @synx-ai/oas3-mdx
+
+      - name: "Extract version"
+        id: extract_version
+        run: |
+          if [ -z ${{ inputs.downstream-version }} ]; then
+            echo "DOWNSTREAM_VERSION=$(sed -nr "{ :l /^version[ ]*=/ { s/[^=]*=[ ]*//; p; q;}; n; b l;}" ./gradle.properties | tr -d ' ' | tr -d '"')" >> "$GITHUB_OUTPUT"
```
paullatzelsperger, line 60: please use the commands from the `publish-swaggerhub.yaml` file.

paullatzelsperger (changes requested):

edc-controlplane/edc-controlplane-postgresql-hashicorp-vault/notice.md
bcronin90: Indirectly. Markdownlint (and apparently some parsers) doesn't like bare links.

edc-extensions/dataplane-proxy/edc-dataplane-proxy-provider-api/src/main/java/org/eclipse/tractusx/edc/dataplane/proxy/provider/api/DataPlaneProxyProviderApiExtension.java
bcronin90: Yeah, this does seem lie a rebase issue. I'll look into that tomorrow.

.github/workflows/publish-swaggerhub.yaml
bcronin90: I thought I had, but apparently there was some problem there. I'll look into it when I'm back.

.github/workflows/publish-docusaurus.yaml
bcronin90: I think there's a problem with those. I've run into a case where the timing of the substitution of variables by the workflow led to an empty variable being used. Hence I switched to this approach and would recommend adapting it for `publish-swaggerhub.yaml` too.

edc-controlplane/edc-controlplane-postgresql-hashicorp-vault/notice.md
bcronin90: Mentioned it in the description, as requested.

.github/workflows/publish.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
           [ ! -z "${{ secrets.ORG_OSSRH_USERNAME }}" ] && 
           [ ! -z "${{ secrets.ORG_OSSRH_PASSWORD }}" ]  && echo "HAS_OSSRH=true" >> $GITHUB_OUTPUT
           [ ! -z "${{ secrets.SWAGGERHUB_API_KEY }}" ]  && 
-          [ ! -z "${{ secrets.SWAGGERHUB_USER }}" ]  && echo "HAS_SWAGGER=true" >> $GITHUB_OUTPUT
+          [ ! -z "${{ secrets.SWAGGERHUB_USER }}" ]  && echo "HAS_SWAGGER=true >> $GITHUB_OUTPUT
```
paullatzelsperger, line 64: pls revert

.github/workflows/publish.yaml
```diff
@@ -135,6 +135,13 @@ jobs:
     permissions:
       contents: read
     needs: [ secret-presence ]
-    if: needs.secret-presence.outputs.HAS_SWAGGER
+    if: needs.secret-presence.output.HAS_SWAGGER
```
paullatzelsperger, line 138: pls revert

.github/workflows/publish.yaml
```diff
@@ -135,6 +135,13 @@ jobs:
...
     uses: ./.github/workflows/publish-swaggerhub.yaml
-    secrets: inherit
+
+  publish-docusaurus:
+    name: Publish docusaurus docs
+    permissions:
+      contents: read
+    needs: [ secret-presence, publish-to-swaggerhub ]
+    if: needs.secret-presence.output.HAS_SWAGGER
```
paullatzelsperger, line 146: it must be `outputs`, plural

.github/workflows/publish-swaggerhub.yaml
```diff
@@ -31,17 +31,6 @@ on:
         required: false
         type: string
 
-  workflow_dispatch:
```
paullatzelsperger, line 34: please revert all changes in this file.

.github/workflows/publish-docusaurus.yaml
```diff
@@ -0,0 +1,111 @@
...
+    name: "Publish docusaurus documentation as zip file"
+    runs-on: ubuntu-latest
+    env:
+      SWAGGERHUB_API_KEY: ${{ secrets.SWAGGERHUB_API_KEY }}
+      SWAGGERHUB_USER: ${{ secrets.SWAGGERHUB_USER }}
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: "Setup node"
+        uses: actions/setup-node@v3
```
paullatzelsperger, line 50: why is this needed?

.github/workflows/publish-docusaurus.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+      - name: "Extract version"
+        id: extract_version
+        run: |
+          if [ -z ${{ inputs.downstream-version }} ]; then
+            echo "DOWNSTREAM_VERSION=$(sed -nr "{ :l /^version[ ]*=/ { s/[^=]*=[ ]*//; p; q;}; n; b l;}" ./gradle.properties | tr -d ' ' | tr -d '"')" >> "$GITHUB_OUTPUT"
```
paullatzelsperger, line 56: pls use the `grep` command from `publish-swaggerhub.yaml`: 
```shell
export DOWNSTREAM_VERSION=$(grep "version" gradle.properties  | awk -F= '{print $2}')
```
it is shorter and more readable. also, `sed` is usually used to replace text, which is not what you're doing here.

.github/workflows/publish-docusaurus.yaml
```diff
@@ -0,0 +1,111 @@
...
+          else
+            echo "DOWNSTREAM_VERSION=${{ inputs.downstream-version }}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: "Gather documentation files"
+        run: |
+          mkdir -p ${{ runner.temp }}/tractusx-edc/Connector\ Kit
```
paullatzelsperger, line 63: could all those `cp`/`rsync` commands etc. be stored in a `create_kit.sh` file? that way, it would be easy to assemble a KIT locally and verify its contents.

paullatzelsperger (changes requested):

.github/workflows/publish-swaggerhub.yaml
bcronin90: Sorry about that mess. I can only guess this got messed between the many rebases. Fixed.

.github/workflows/publish-docusaurus.yaml
bcronin90: Yes. Good idea.

bcronin90:
Should be ready aside from the dependencies issue.

paullatzelsperger (approved):
LGTM

wolf4ood (approved):

PR #722, opened 2023-08-21
chore: add LICENSE file to all charts

### LINKED ISSUE #718: [TRG 5.02] Add `LICENSE` to Charts
## Description

While the source code is automatically put under the repos root `LICENSE`, we should also include that info on released artifacts like the release Charts. 

The easiest way to achieve this, is to put a copy of the `LICENSE` to each Charts root folder (i.e `/charts/tractusx-connector-azure-vault/LICENSE`)

related to: #714

paullatzelsperger:
## WHAT

Copies the LICENSE file to all charts. That way they get packaged during release.

## WHY

all published charts should contain a LICENSE file

## FURTHER NOTES

Closes #718

### DIFF
--- a/charts/tractusx-connector-azure-vault/LICENSE
+++ b/charts/tractusx-connector-azure-vault/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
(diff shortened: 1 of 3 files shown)


SebastianBezold (approved):
👍

PR #764, opened 2023-09-13
chore: add default values to helm charts

### LINKED ISSUE #724: Clarification: Deployability of Helm Chart with default config
## Description

The description of [TRG 5.01](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-01/) contains a section, that Charts should be installable on any environment with just the config provided by the default `values.yaml`.

I tried that with all of the Charts contained in this repo and while I can see, that `tractusx-connector-azure-vault` would most likely need config in any circumstance, I was expecting the `tractusx-connector-memory` to work without further config.

For `tractusx-connector-memory`, the `.Values.participant.id` property is not present in the default values. A placeholder would be finde. We do know, that some functionality wont work as intended without proper config, but at least a successful install (and warnings in the logs about missing config for example) is the goal of that description in the TRG

blocks #714

## Additional info

All of the tests where done with the released Charts in version `0.5.0` 

I also tried `tractusx-connector` and `tractusx-connector-legacy`, just to give every of the Charts a shot, but no success

tuncaytunc-zf:
## WHAT

Add default values to helm charts

## WHY

Because of QG checks [TRG 5.01](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-01/)

## FURTHER NOTES

- Cleanup daps values from memory chart
- Recreate README.md files

Closes #724

### DIFF
--- a/charts/tractusx-connector-azure-vault/README.md
+++ b/charts/tractusx-connector-azure-vault/README.md
@@ -247,7 +247,7 @@ helm install my-release tractusx-edc/tractusx-connector-azure-vault --version 0.
 | imagePullSecrets | list | `[]` | Existing image pull secret to use to [obtain the container image from private registries](https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry) |
 | install.postgresql | bool | `true` |  |
 | nameOverride | string | `""` |  |
-| participant.id | string | `"BPNL00DATAP00001"` | BPN Number |
+| participant.id | string | `"BPNLCHANGEME"` | BPN Number |
 | postgresql.auth.database | string | `"edc"` |  |
 | postgresql.auth.password | string | `"password"` |  |
 | postgresql.auth.username | string | `"user"` |  |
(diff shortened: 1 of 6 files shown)


charts/tractusx-connector-azure-vault/README.md
```diff
@@ -262,10 +262,10 @@ helm install my-release tractusx-edc/tractusx-connector-azure-vault --version 0.
 | tests | object | `{"hookDeletePolicy":"before-hook-creation,hook-succeeded"}` | Configurations for Helm tests |
 | tests.hookDeletePolicy | string | `"before-hook-creation,hook-succeeded"` | Configure the hook-delete-policy for Helm tests |
 | vault.azure.certificate | string | `nil` |  |
-| vault.azure.client | string | `""` |  |
+| vault.azure.client | string | `"53ba6f2b-6d52-4f5c-8ae0-7adc20808854"` |  |
```
paullatzelsperger, line 265: since this default value is arbitrary, we should use something like `<AZURE_CLIENT_ID>` instead of a UUID, so that it is very obvious. 

Same is true for `vault.azure.tenant`

charts/tractusx-connector-azure-vault/values.yaml
```diff
@@ -38,7 +38,7 @@ customLabels: {}
 
 participant:
   # -- BPN Number
-  id: ""
+  id: "BPNL00DATAP00001"
```
paullatzelsperger, line 41: this should be more obvious, something like `BPNLCHANGEME` or similar

charts/tractusx-connector-memory/values.yaml
```diff
@@ -32,7 +32,7 @@ customLabels: {}
 
 participant:
   # -- BPN Number
-  id: ""
+  id: "BPNL00DATAP00001"
```
paullatzelsperger, line 35: see comments above

paullatzelsperger (approved):
LGTM

charts/tractusx-connector-azure-vault/values.yaml
```diff
@@ -520,7 +520,8 @@ vault:
     transferProxyTokenEncryptionAesKey: transfer-proxy-token-encryption-aes-key
 
 backendService:
-  httpProxyTokenReceiverUrl: ""
+  # Specifies a backend service which will receive the EDR
+  httpProxyTokenReceiverUrl: "httpProxyTokenReceiverUrl"
```
florianrusch-zf, line 524: Add `--` so that this will be appear in the README.md in the description column and maybe also add a real url which also indicates that it needs to be changed:

```suggestion
  # -- Specifies a backend service which will receive the EDR
  httpProxyTokenReceiverUrl: "https://example.com"
```

PR #794, opened 2023-09-26
chore(helm): re-enable resource requests and limits

### LINKED ISSUE #720: [TRG 5.04] Define resource limit/request for Charts
**Is your feature request related to a problem? Please describe.**
In our release guidelines, that are defined for the whole Eclipse Tractus-X project, we specify, that each Chart should provide
default resource -request and -limit values. 
In best case the values are gathered by monitoring the EDC under different load circumstances. This would give users a good initial understanding or resource consumption

**Describe the solution you'd like**
There are already request and limit values specified as comments in the values.yaml. To follow the release guideline, they could simply be used as actual defaults.

**Describe alternatives you've considered**
If there are valid reasons, why resource requests and limits cannot be set, we should change the [existing TRG 5.04](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-04/), so the guidelines are followed and we do not need to track exceptions.

**Additional context**
related to, but not blocking #714

saschaisele-zf:
## WHAT

Re-enables the ressource requests and ressource limits in the helm charts.
Additionally gives example default values.

## WHY

See #720

## FURTHER NOTES

Closes #720


charts/tractusx-connector/values.yaml
```diff
@@ -253,12 +253,12 @@ controlplane:
...
   # resources, such as Minikube. If you do want to specify resources, uncomment the following
   # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-  # limits:
-  #   cpu: 100m
-  #   memory: 128Mi
-  # requests:
-  #   cpu: 100m
-  #   memory: 128Mi
+  limits:
+    cpu: 1.2
```
florianrusch-zf, line 257: `1.2` looks strange to me :D I've never seen a helm chart where someone set something like that. I would guess that most are using 0.5 or even 1 increments. Also 1 vs 1.2 won't make that much difference.

```suggestion
    cpu: 1
```

charts/tractusx-connector/values.yaml
```diff
@@ -451,12 +451,12 @@ dataplane:
...
```
florianrusch-zf, line 455:
```suggestion
    cpu: 1
```

florianrusch-zf (approved):

charts/tractusx-connector/values.yaml
saschaisele-zf: Changed upwards to 1.5 instead to avoid small spikes in processing power casing issues.

charts/tractusx-connector-azure-vault/values.yaml
```diff
@@ -453,12 +453,12 @@ dataplane:
...
     # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
-    # limits:
-    #   cpu: 100m
-    #   memory: 128Mi
-    # requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+  limits:
+    cpu: 1.5
+    memory: 512Mi
```
florianrusch-zf, line 458: This is different in regard to the other helm charts

saschaisele-zf: Has been adjusted.

florianrusch-zf (approved):

wolf4ood (approved):

PR #796, opened 2023-09-27
Feature: Add config parameter for ProviderGatewayController context registration

### LINKED ISSUE #365: Improve context registration of the `ProviderGatewayController`
Suggestion: if there is a dedicated `web.http.gateway.*` config, we register the controller into that, and otherwise we'll register it into the `default` context

_Originally posted by @jimmarino in https://github.com/eclipse-tractusx/tractusx-edc/pull/357#discussion_r1196030839_

bcronin90:
## WHAT

Adds an optional setting to specify a jetty context into which to register the ProviderGatewayController.

## WHY

[365](https://github.com/eclipse-tractusx/tractusx-edc/issues/365)

## FURTHER NOTES

I don't know whether this is actually the correct approach.

Closes #365

### DIFF
--- a/edc-extensions/dataplane-proxy/edc-dataplane-proxy-provider-api/README.md
+++ b/edc-extensions/dataplane-proxy/edc-dataplane-proxy-provider-api/README.md
@@ -19,5 +19,6 @@ When the proxy receive a request, it must contain the EDR, which will be decoded
 ## Configuration
 
 | Key                                        | Required | Default       | Description                                                 |
-|---------------------------------|----------|----------------------------------------------------------------------------------------|
+|--------------------------------------------|----------------------------------------------------------------------------------------|
 | tx.dpf.provider.proxy.thread.pool          |          | 10            | Thread pool size for the provider data plane proxy gateway  |
+| web.http.gateway.context                   |          | default       | Context to register the ProviderGatewayController into      |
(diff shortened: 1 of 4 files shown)


wolf4ood:
I would do mostly like the upstream management configuration does

https://github.com/eclipse-edc/Connector/blob/main/extensions/common/api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java

The important bit is `.useDefaultContext(true)` on `WebServiceSettings.Builder` which make it register to the default context if the context config key is not found, and it will make it backward compatible if no configuration is provided.

Let me know if this helps

bcronin90:
> I would do mostly like the upstream management configuration does
> 
> https://github.com/eclipse-edc/Connector/blob/main/extensions/common/api/management-api-configuration/src/main/java/org/eclipse/edc/connector/api/management/configuration/ManagementApiConfigurationExtension.java
> 
> The important bit is `.useDefaultContext(true)` on `WebServiceSettings.Builder` which make it register to the default context if the context config key is not found, and it will make it backward compatible if no configuration is provided.
> 
> Let me know if this helps

The problem there is that iirc we don't really have defaults for port and path. We just check whether they're provided and if they are not, we just like let the WebService handle things.

wolf4ood (approved):

wolf4ood:
Thanks @bcronin90

PR #801, opened 2023-09-29
feat(SqlEdrStore): add configurable vault path

### LINKED ISSUE #800: EDR-Cache Extension: Add ability to store tokens in specific Vault folder
## WHAT

The EDR-Cache Extension should be able to store EDR tokens in a Vault folder.

## WHY

If no folder can be specified and the tokens are written to the "main" vault folder of the service, it can be difficult to keep track of them.
Vault EDR Tokens

## HOW

A new string parameter should be passed to the constructor. By default, this parameter should be empty. If it is not empty, a folder can be specified.

```java
    public static final String SEPARATOR = "--";
    private final String vaultPath;
    public final String vaultPrefix;
    ...
    
    this.vaultPrefix = vaultPath + "edr" + SEPARATOR;
    ...
    
    cache = new SqlEndpointDataReferenceCache(extension.getDataSourceRegistry(), extension.getDatasourceName(), "edr-cache/", extension.getTransactionContext(), statements, typeManager.getMapper(), vault, clock, queryExecutor, CONNECTOR_NAME);
```

wolf4ood:
## WHAT

Add the ability to configure a path in the vault for storing the EDRs

## WHY

Flexibility

## FURTHER NOTES

Closes #800

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -406,6 +406,7 @@ maven/mavencentral/org.testcontainers/jdbc/1.19.0, Apache-2.0, approved, #10348
 maven/mavencentral/org.testcontainers/junit-jupiter/1.19.0, MIT, approved, #10344
 maven/mavencentral/org.testcontainers/postgresql/1.19.0, MIT, approved, #10350
 maven/mavencentral/org.testcontainers/testcontainers/1.19.0, Apache-2.0 AND MIT, approved, #10347
+maven/mavencentral/org.testcontainers/vault/1.19.0, MIT, approved, clearlydefined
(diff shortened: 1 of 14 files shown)


awellnitz-materna (approved):
Executed and tested looks good to me.

What happens to the old tokens that are still valid?
Can we move them to the folder?

wolf4ood:
@awellnitz-materna 

the basic functionality is there, I'm adding some tests on our end for the CI and it should be ready for review. 

If you configure the folder when it's released and move/copy them on the new folder they should be valid We do not store the vault path on our end. So the resolved secret will be `vaultUrl` + ?`vaultDir` + `edr-<uuid>`, so i think you can migrate them.

paullatzelsperger (approved):

Issue #853, opened 2023-10-30
Build and dockerize the project under windows 


mustafasalfiti:
using ./gradlew dockerize under windows will lead to a confusion inside the Dockerfile
because how paths are handled in windows 

for example running the current command in windows 

`./gradlew.bat :edc-controlplane:edc-controlplane-memory-hashicorp-vault:dockerize
`

will lead into this error

```

> Task :edc-controlplane:edc-controlplane-memory-hashicorp-vault:dockerize FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':edc-controlplane:edc-controlplane-memory-hashicorp-vault:dockerize'.
> Could not build image: COPY failed: file not found in build context or excluded by .dockerignore: stat build\opentelemetry-javaagent.jar: file does not exist


```

### Expected behavior

to successfuly build the code

### Screenshots/Error Messages

![image](https://github.com/eclipse-tractusx/tractusx-edc/assets/49871916/700157fe-4355-4d0c-bcbd-1c18067cab59)

_Add any other context or information about the problem here._

- Used version: release/0.5.1
- OS: Windows
- Docker Version: 24.0.5

## Possible Implementation

replace the (windows slashes) in the build.gradle.kts for OTEL_JAR to make it work inside the Dockerfile

paullatzelsperger:
The dockerfile is intended for our build process, which runs on Linux.  The official deliverable of tx-edc is a helm chart plus the prebuilt images. Why do you need to build your own connector image, and why do you need the dockerfile for that to be hosted in the tx-edc repository (as opposed to: your own infrastructure)?

PR #854, opened 2023-10-30
Fix/dockerfile windows

mustafasalfiti:
this made it work under windows
but I need a confirmation that this still work under Linux and MacOS as i think it should continue working normally

#853


paullatzelsperger:
Please see my comment in the associated issue. The docker file is only intended to be used by our internal build process, and I don't see a reason why we should care about windows paths.

florianrusch-zf:
I agree with @paullatzelsperger. Also the change you suggest will break the functionality in all Unix environments.

If you want to make it run on your machine, you can try to use the WSL, which I always recommend when developing on Windows.

paullatzelsperger:
Unrelated: please check out our pr guidelines, as your pr violates many of them

mustafasalfiti:
thank you @florianrusch-zf for your feedback, i can explain exactly what i did in the build.gradle.kts 
I just changed this line buildArgs.put("OTEL_JAR", agentFile.relativeTo(dockerContextDir).path) 
into buildArgs.put("OTEL_JAR", agentFile.relativeTo(dockerContextDir).path.replace("\\", "/"))

so that will change any backslash into forward slash which will cause no problems in the dockerfile.
From my own experience using linux there wont be any backslash in path strings
when the code inside the build.gradle.kts runs under MacOS or Linux we will get a similiar path to
build/ELEMENT.jar

but under windows

build\ELEMENT.jar

so the replace will only change the \ into /

I would be really grateful if you can tell me what do you mean by that and how?
,,Also the change you suggest will break the functionality in all Unix environments"

explaining more will help me to understand

mustafasalfiti:
@paullatzelsperger I think the reason why you should care about windows path and how it should be handled correctly its because windows is the most popular used OS currently.

It can take sometimes hours and days trying to figure out whats wrong and why its not building the projects, so i would prefer that to be written in documentation that its not compatible to be tested and developed under windows because we don't care about it.

it will save us time

thank you for pointing the PR violations i will take it that into consideration for next time.

paullatzelsperger:
> @paullatzelsperger I think the reason why you should care about windows path and how it should be handled correctly its because windows is the most popular used OS currently.
> 
> 
> 
> It can take sometimes hours and days trying to figure out whats wrong and why its not building the projects, so i would prefer that to be written in documentation that its not compatible to be tested and developed under windows because we don't care about it.
> 
> 
> 
> it will save us time
> 
> 
> 
> thank you for pointing the PR violations i will take that into consideration for next time.

Again, the docker file is not intended for public consumption, as the official deliverables are docker images and helm charts. No point in documenting all unsupported scenarios. Users either stick to the supported ones, or they are on their own, as that would be an advanced use case, such as contributing to the project directly.

Advanced users are expected to figure out such basic build principles.

If for some reason you want to build your own docker image of tx-edc, you should create and maintain your own dockerfile instead of relying an internal tx-edc one.

PR #874, opened 2023-11-13
feat: adds support for multiple trusted issuers 

### LINKED ISSUE #814: IATP: extend configuration for multiple trusted issuers
## WHAT

Currently, a connector can only have one "allowed issuer", i.e. it can only accept credentials created by one entity. In preparation for adopting the Identity And Trust Protocols (IATP), we need to extend this to allow multiple issuers.

## WHY
Ultimately, we will have multiple issuers (~operating companies), so we need to prepare for that.

## HOW
There are two general ways to approach this: 
1. add the allowed issuers via configuration: this is how we do it currently (with one issuer), and it's by far the easiest way.
2. download a list of allowed issuers: this raises the difficult question of how the authenticity and integrity of such a list can be verified, and as of now, there is no agreed-upon way of doing that. It may come at a later point in time.

_We will go with approach 1 for now_.

Specifically, this means that the `SsiCredentialIssuerValidationRule` now has to be able to deal with a `List<String>` instead of a `String`. Configuration-wise, it should just be a comma-separated list.
 

## FURTHER NOTES

- issuer-IDs are in practice DID's

wolf4ood:
## WHAT

adds supports for multiple trusted issuers  in `SsiCredentialIssuerValidationRule`

## WHY

allow multiple trusted issuers

Closes #814


wolf4ood:
This will be moved in upstream 

https://github.com/eclipse-edc/Connector/issues/3602

PR #903, opened 2023-12-05
chore: remove edc-controlplane-memory-hashicorp-vault

### LINKED ISSUE #902: Remove `edc-controlplane-memory-hashicorp-vault` from various publications
## Describe the bug

Currently, many Github Workflows (Veracode, Trivy, etc.) reference the `edc-controlplane-memory-hashicorp-vault` - a variant which is not used anywere. We publish a docker image, but we're not using that in any of the official helm charts or anything.
Deleting was even mandated in [this decision-record](https://github.com/eclipse-tractusx/tractusx-edc/blob/568c5b7448f47935551524b129f5c83d8b170529/docs/development/decision-records/2023-04-11_refactor_helmcharts/README.md) but never materialized.

While it isn't a bug in the strict sense of the word, it still is an oddity and not needed.

## Possible Implementation

Get rid of the `edc-controlplane-memory-hashicorp-vault` module, and all occurrences in all workflows.

paullatzelsperger:
## WHAT

This PR removes the `edc-controlplane-memory-hashicorp-vault` as it is not used anywhere.

## WHY

we should not keep unused code around.

## FURTHER NOTES

Closes #902


jimmarino (approved):
🪓

ndr-brt (approved):
:axe:

PR #905, opened 2023-12-05
chore: remove markdown linter

paullatzelsperger:
## WHAT

Removes the `markdown-linter` step from the verify workflow.

## WHY

Linting markdown is not required by any TRG or other guide line and we have an editorconfig for code

## FURTHER NOTES


florianrusch-zf:
I would strongly recommend not removing it! I've simply seen far too much miserable Markdown code in the past.

It's not just about the aesthetics of the code itself, but also about ensuring basic things like the correct order of headings. I see it more as a check to make the documentation look more standardised in the end.

Have there been problems with this workflow in the past, or what is the reason behind your intention to remove it now?

paullatzelsperger:
> I would strongly recommend not removing it! I've simply seen far too much miserable Markdown code in the past.
> 
> 
> 
> It's not just about the aesthetics of the code itself, but also about ensuring basic things like the correct order of headings. I see it more as a check to make the documentation look more standardised in the end.
> 
> 
> 
> Have there been problems with this workflow in the past, or what is the reason behind your intention to remove it now?

I had previously thought that it was mandatory to have the linter when in fact it is not. One problem is that there are standardized documents such as the SECURITY.md which violates the linter config (no raw URLs allowed). Not sure we can or should require the file to change for all of Catena X.
In my personal experience this linter has only ever caused headaches and annoyances and never improved anything.

SebastianBezold (approved):
👍

ndr-brt (approved):
:axe:

PR #908, opened 2023-12-06
fix(helm): Add missing public service

KilianHaag:
## WHAT

Add the public service and includes it in the "Public / Internet facing" Ingress

## WHY

The public service is referenced in the values under runtime.endpoints.public, but is not created as a service. Therefore it is also missing in the ingress.


florianrusch-zf:
We need this for the community days, therefor I will hit the merge button without waiting for further reviews ;-)

PR #910, opened 2023-12-07
chore: add legal info to Docker images

### LINKED ISSUE #889: Clarification: Legal information for distributions (TRG 7.05)
## Disclaimer 

I am currently unsure, if we include all the necessary legal information for distribution. If my did miss something, please point that out.

## Situation

I can see, that the Dockerfiles do include an open telemetry agent: https://github.com/eclipse-tractusx/tractusx-edc/blob/main/edc-controlplane/edc-controlplane-memory-hashicorp-vault/src/main/docker/Dockerfile#L44

The download URL and therefore version seems to be defined in the `build.gradle` (https://github.com/eclipse-tractusx/tractusx-edc/blob/main/build.gradle.kts#L155 and read during build (https://github.com/eclipse-tractusx/tractusx-edc/blob/main/.github/actions/publish-docker-image/action.yml#L53-L58)

## Potential missing information

I cannot find any mention of the open telemetry agent in the `notice.md` file, that is used for publishing on Dockerhub. In case we alter the base image and add additional apps, the notice should point out any available information, that we have.
The open telemetry mentions in the `DEPENDENCIES` file do not seem to cover the agent.

## Bonus

We should also include the legal documentation present in our repo as part of our distributed Docker images.
A simple `COPY LICENSE NOTICE.md DEPENDENCIES SECURITY.md app/` should do the trick

paullatzelsperger:
## WHAT

This PR adds legal information about the OpenTelemetry Agent to the `notice.md` that we include with each Docker image.

In addition, the following legal documents are copied to the application root of the docker image: NOTICE.md, SECURITY.md, LICENSE, DEPENDENCIES

## WHY

TRG 7.07

## FURTHER NOTES

Closes #889

### DIFF
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -35,7 +35,9 @@ in the GitHub organization <https://github.com/eclipse-tractusx>:
 
 This project leverages the following third party content.
 
-See DEPENDENCIES file.
+- OpenTelemetry Agent v1.32.0: <https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.32.0>
+
+For additional dependencies, see [DEPENDENCIES](./DEPENDENCIES) file.
 
 ## Cryptography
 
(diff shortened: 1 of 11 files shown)


SebastianBezold (approved):
👍

PR #928, opened 2023-12-11
feat: increase EDR token validity

### LINKED ISSUE #927: Increase token validity for Consumer Pull scenario
## WHAT

Increase `edc.transfer.proxy.token.validity.seconds` to `2592000` (30 days) as default value.

## WHY

Currently the `edc.transfer.proxy.token.validity.seconds` configuration is not set in our charts the upstream EDC default is `600` (10 min).

This has a couple of downside:

- When using the EDR api, it will cause to much refreshs and therefore too much transfer processes
- When managing EDR manually, it can also be affected since users might need to have access to data for a longer period than 10 minutes and it will require anyway to open new transfer processes every time the token  expires

wolf4ood:
## WHAT

Increase EDR token validity to 30 days

## WHY

Lowering down the number of transfer process for data transfers

Closes #927


paullatzelsperger (approved):
LGTM, but didn't we talk about increasing it to a month?

wolf4ood:
Ah right, let me change it

ndr-brt (approved):

PR #934, opened 2023-12-14
fix: backport of transfer process completion on EDR expired

wolf4ood:
## WHAT

Backport the completion of the transfer process an an EDR expired

## WHY

To not leave the transfer process `STARTED` if the token is expired


ndr-brt (approved):

edc-extensions/edr/edr-callback/src/main/java/org/eclipse/tractusx/edc/callback/TransferProcessLocalCallback.java
```diff
@@ -105,6 +105,15 @@ private void cleanOldEdr(String assetId, String agreementId) {
...
             entry.transitionToExpired();
             edrCache.update(entry);
+
+            var transferProcess = transferProcessStore.findById(entry.getTransferProcessId());
+
+            if (transferProcess != null && transferProcess.canBeCompleted()) {
+                transferProcess.transitionCompleting();
+                transferProcessStore.save(transferProcess);
+            } else {
+                monitor.info(format("Cannot terminate transfer process with id: %s", entry.getTransferProcessId()));
```
paullatzelsperger, line 115: maybe move to `TERMINATED`?

ndr-brt: yes, maybe transitioning to `TERMINATING` is more correct, as it is the same behavior the policy-monitor is following as well.

wolf4ood: Afaik currently the terminating on consumer side does not trigger a `TransferTerminationMessage` consumer -> provider, and the provider will still have the his TP in started state.

paullatzelsperger: it should, at least if the state is "higher" than `REQUESTED`. From `TransferProcessManagerImpl`:
```java
    private boolean processTerminating(TransferProcess process) {
        if (process.getType() == CONSUMER && process.getState() < REQUESTED.code()) {
            transitionToTerminated(process);
            return true;
        }

        return entityRetryProcessFactory.doSyncProcess(process, () -> terminateDataFlow(process))
                .onSuccess((p, dataFlowResponse) -> sendTransferTerminationMessage(p))
                .onFailure((t, failure) -> transitionToTerminating(t, failure.getFailureDetail()))
                .onFatalError((p, failure) -> transitionToTerminated(p, failure.getFailureDetail()))
                .onRetryExhausted((p, failure) -> transitionToTerminated(p, failure.getFailureDetail()))
                .execute("Terminate data flow");
    }
```

paullatzelsperger (approved):

PR #937, opened 2023-12-15
fix(helm): Remove deprecated/separate validation endpoint

### LINKED ISSUE #909: Deprecated validation endpoint
## Describe the bug

The memory chart references and creates a validation endpoint and service.
Afaik this endpoint is not available anymore and results in a 404 error during token validation.

### To Reproduce

1. Deploy the default edc memory chart two times
2. Try a data transfer between the two EDCs
3. Fails with 404 by the token transfer endpoint

### Expected behavior

The template of the "normal" chart uses the CONTROL_PATH and CONTROL_PORT instead.
After manually changing the EDC_DATAPLANE_TOKEN_VALIDATION_ENDPOINT to the control endpoint, the transfer succeeds.

## Context Information

- Used version: chart 0.5.3

## Possible Implementation

I guess the fix would be to use the same template as in the "normal" chart.
But I wanted to confirm, before I create a PR.

KilianHaag:
## WHAT

The validation endpoint is not separate anymore and instead part of the /control endpoint.
This PR removes all references to a separate /validation endpoint and instead uses the /control endpoint.
Also updates any references to the validation port in the docs.

## WHY

The memory chart currently doesn't work, as the WEB_HTTP_VALIDATION_PORT is not functional and returns a 404.

## FURTHER NOTES

There is a clunky workaround, by setting additional environment variables and changing the validation endpoint path and port:
```
  runtime:
    endpoints:
      validation:
        path: /foobar
        port: 8092
    env:
      WEB_HTTP_VALIDATION_PATH: '/foobar'
      WEB_HTTP_VALIDATION_PORT: '8091'
      EDC_DATAPLANE_TOKEN_VALIDATION_ENDPOINT: http://***-runtime:8083/control/token
```
But this causes issues, when deployed via argoCD, as the duplicate Env vars cause proplems during the diff phase.

Closes #909


paullatzelsperger (approved):
LGTM

wolf4ood (approved):

ndr-brt (approved):
:rocket:

PR #971, opened 2024-01-02
fix: rename sftp service extensions path

### LINKED ISSUE #970: SFTP extensions are failing to initiate during runtime
## Describe the bug
The SFTP extensions (transferprocess-sftp-provisioner, transferprocess-sftp-client) are failing to initiate during runtime.

## Possible Implementation
The issue lies in the incorrect path specified in the SFTP ServiceExtension files. The SFTP ServiceExtension files, currently residing in the resources directory, need to be renamed to the correct path. Currently, they are using the path **org.eclipse.dataspaceconnector.spi.system.ServiceExtension**, which is incorrect. They should be updated to use the correct **path org.eclipse.edc.spi.system.ServiceExtension**.

**Following file path**:
edc-extensions/transferprocess-sftp-provisioner/src/main/resources/META-INF/service/org.eclipse.dataspaceconnector.spi.system.ServiceExtension
edc-extensions/transferprocess-sftp-client/src/main/resources/META-INF/services/org.eclipse.dataspaceconnector.spi.system.ServiceExtension

git-masoud:
## WHAT

Fix the stfp service extension files' names to the expected path which edc requires. 

## WHY

The SFTP extensions (transferprocess-sftp-provisioner, transferprocess-sftp-client) are failing to initiate during runtime, becuase the service extension files name has the old path of edc. 

Closes #970


wolf4ood (approved):
Thanks @git-masoud

PR #1029, opened 2024-02-07
build: add header check to comply with TRG 7.02

### LINKED ISSUE #88: Fix license header to match Eclipse recommendations
Please fix license header in src code files to match Eclipse recommendations.

ndr-brt:
## WHAT

Add a check on checkstyle to ensure the headers are in the correct format according to [TRG 7.02](https://eclipse-tractusx.github.io/docs/release/trg-7/trg-7-02/)

## WHY

TRG compliance

## FURTHER NOTES

Closes #88


wolf4ood (approved):
:rocket:

build.gradle.kts
```diff
@@ -109,11 +111,18 @@ allprojects {
         configFile = rootProject.file("resources/tx-checkstyle-config.xml")
         configDirectory.set(rootProject.file("resources"))
 
+        tasks.checkstyleMain {
```
paullatzelsperger, line 114: why do we need this? shouldn't `checkstyleMain` run on the `main` source set anyway?

ndr-brt: I was looking into checking also `resources` file (as `.ServiceExtension` files), but the gradle checkstyle plugin by default is excluding everything not java:
https://github.com/gradle/gradle/blob/f8ee6cae5bd056483bafec72043744540f532e09/platforms/jvm/code-quality/src/main/groovy/org/gradle/api/plugins/quality/CheckstylePlugin.java#L139

in any case I forgot to do it :sweat_smile:  will do it right now

paullatzelsperger: interesting 🤔 maybe a comment would be useful then?

paullatzelsperger (approved):

PR #1040, opened 2024-02-09
improvement(memory-edc): remove EDC_VAULT_SECRETS rewrite

### LINKED ISSUE #1039: improvement: Remove rewriting of ENV variable in memory Dockerfile and chart
## WHAT

The chart should set the ENV variable EDC_VAULT_SECRETS instead of SECRETS and the Dockerfile should not set '-Dedc.vault.secrets=\"${SECRETS}\"'

## WHY

With the current setup, this is the only ENV variable, that is rewritten. While checking the documentation of the edc components, this rewrite is not apparent. In short: I don't see any reason for it, except potential confusion.

## HOW

If accepted I would be happy to create a PR

KilianHaag:
## WHAT

Removes the rewrite of ENV variable "SECRETS" to edc.vault.secrets and instead uses the EDC_VAULT_SECRETS directly in the memory chart and Dockerfile.

## WHY

With the current setup, this is the only ENV variable, that is rewritten. While checking the documentation of the edc components, this rewrite is not apparent. In short: I don't see any reason for it, except potential confusion.

## FURTHER NOTES

Also fixes the documentation regarding the seeding of secrets with semicolon separated k:v list

Closes #1039

### DIFF
--- a/charts/tractusx-connector-memory/templates/deployment-runtime.yaml
+++ b/charts/tractusx-connector-memory/templates/deployment-runtime.yaml
@@ -252,7 +252,7 @@ spec:
             ###########
 
             # see extension https://github.com/eclipse-tractusx/tractusx-edc/tree/develop/edc-extensions/hashicorp-vault
-            - name: "SECRETS"
+            - name: "EDC_VAULT_SECRETS"
               value: {{ .Values.vault.secrets | quote}}
 
             #####################
--- a/charts/tractusx-connector-memory/values.yaml
+++ b/charts/tractusx-connector-memory/values.yaml
@@ -18,7 +18,6 @@
 #  SPDX-License-Identifier: Apache-2.0
 #################################################################################
 
-
 ---
 # Default values for eclipse-dataspace-connector.
 # This is a YAML-formatted file.
@@ -289,7 +288,7 @@ runtime:
     public: ""
     readiness: ""
 vault:
-  # secrets can be seeded by supplying them in a comma separated list key1:secret2,key2:secret2
+  # secrets can be seeded by supplying them in a semicolon separated list key1:secret2;key2:secret2
   secrets: ""
   secretNames:
     transferProxyTokenEncryptionAesKey: transfer-proxy-token-encryption-aes-key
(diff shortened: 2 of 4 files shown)


paullatzelsperger:
Please provide proper PR description, i.e. _what_ you changed, and _why_ you changed it and why in this way.

KilianHaag:
I created https://github.com/eclipse-tractusx/tractusx-edc/issues/1058 for the potentially flaky test and changed the description.

In this case I can not think of a test, as the changes should not alter the existing behavior.
But the [deployment-test / test-in-memory](https://github.com/eclipse-tractusx/tractusx-edc/actions/runs/7844049578/job/21405612001?pr=1040#logs) should be covering this, as I expect the deployment to fail, if secrets are missing in the (memory-)vault.

wolf4ood:
@KilianHaag  I've restarted the test meanwhile

KilianHaag:
@wolf4ood Thanks for triggering them again.
Seems like all the tests for this PR are successful now, but (sorry for the stupid question) what are these 2 runs under my name:
image

I don't understand how they got triggered.

paullatzelsperger:
> I expect the deployment to fail, if secrets are missing in the (memory-)vault.
it will not, as the deployment test only tests that the runtime can be deployed and booted up. Missing secrets could potentially go unnoticed.

paullatzelsperger (approved):

KilianHaag:
@paullatzelsperger I might be mistaken, but when I explicitly omit the '--set vault.secrets="client-secret:$(cat client.secret)"' from the test deployment, the result will be a CrashLoopBackOff:
`Caused by: java.lang.NullPointerException: Client secret could not be retrieved
    at java.base/java.util.Objects.requireNonNull(Unknown Source)
    at org.eclipse.tractusx.edc.iam.ssi.miw.SsiMiwOauth2ClientExtension.createConfiguration(SsiMiwOauth2ClientExtension.java:75)
    at org.eclipse.tractusx.edc.iam.ssi.miw.SsiMiwOauth2ClientExtension.oauth2Client(SsiMiwOauth2ClientExtension.java:67)
    ... 23 more`

So what I meant is: If my change would have made all secret(s) unavailable, the deployment should fail, as the client-secret is retrieved during startup.

PR #1055, opened 2024-02-14
chore: backport of issue upstream#3607

wolf4ood:
## WHAT

Backport of https://github.com/eclipse-edc/Connector/pull/3607 in 0.5.x

### DIFF
--- a/edc-extensions/auth-tokenbased/src/main/java/org/eclipse/tractusx/edc/api/auth/token/TokenBasedAuthenticationService.java
+++ b/edc-extensions/auth-tokenbased/src/main/java/org/eclipse/tractusx/edc/api/auth/token/TokenBasedAuthenticationService.java
@@ -1,0 +1,55 @@
+/*
+ *  Copyright (c) 2020 - 2022 Microsoft Corporation
+ *
+ *  This program and the accompanying materials are made available under the
+ *  terms of the Apache License, Version 2.0 which is available at
+ *  https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


edc-extensions/auth-tokenbased/src/main/java/org/eclipse/tractusx/edc/api/auth/token/TokenBasedAuthenticationService.java
```diff
@@ -0,0 +1,55 @@
...
+package org.eclipse.tractusx.edc.api.auth.token;
+
+import org.eclipse.edc.api.auth.spi.AuthenticationService;
+import org.eclipse.edc.web.spi.exception.AuthenticationFailedException;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public class TokenBasedAuthenticationService implements AuthenticationService {
```
paullatzelsperger, line 24: nit: I'd mark this as `@Deprecated`, and maybe even call the module `auth-tokenbased-backport`, so we don't forget to remove it in future Tx-EDC releases.

paullatzelsperger (approved):

edc-extensions/auth-tokenbased/src/main/java/org/eclipse/tractusx/edc/api/auth/token/TokenBasedAuthenticationService.java
wolf4ood: This only applies to the `0.5.x` branch and not in `main`  one. We will probably keep it until the next official release is done

PR #1075, opened 2024-02-20
feat: register CX Policy Functions

paullatzelsperger:
## WHAT

This PR registers the new C-X Policy constraint functions with the `PolicyEngine` and dynamically binds them.

In addition, some E2E-Tests were added leveraging the IATP test configuration.

## WHY

The policy functions need to be registered so that VerifiableCredentials are properly evaluated.

## FURTHER NOTES

- had to go back to using `0.5.2-SNAPSHOT` to get the very latest upstream fixes from Feb 20, 2024. We can pin the version to a nightly version the day after.
- MIW integration tests are disabled, as the SummaryCredential eval functions would interfere with the new policy eval functions. Removing them altogether and cleaning up all the SummaryCredential stuff will be done in another PR


wolf4ood (approved):
LGTM :rocket:

PR #1120, opened 2024-03-11
feat: add Token RefreshAPI to data plane

paullatzelsperger:
## WHAT

This PR adds the Refresh API to the dataplane, effectively exposing the possibility to refresh dataplane access tokens.

The Refresh API is registered into th `"public"` context, thus it is available under `POST <HOST>:<PUBLIC_API_PORT>/<PUBLIC_API_PATH>/v1/token`. 
Using all defaults, that is:

```
POST http://localhost:8185/api/v1/public/token
```

or - using the v2 endpoint:

```
POST http://localhost:8185/api/v2/public/token
```

using the V2 endpoint is **highly encouraged** as the V1 endpoint will not use data plane signaling, access tokens or token refresh!

## WHY

Short-lived tokens are a security best-practice
## FURTHER NOTES

### DIFF
--- a/DEPENDENCIES
+++ b/DEPENDENCIES
@@ -4,13 +4,18 @@ maven/mavencentral/com.apicatalog/iron-verifiable-credentials/0.8.1, Apache-2.0,
 maven/mavencentral/com.apicatalog/titanium-json-ld/1.0.0, Apache-2.0, approved, clearlydefined
 maven/mavencentral/com.apicatalog/titanium-json-ld/1.3.1, Apache-2.0, approved, #8912
 maven/mavencentral/com.apicatalog/titanium-json-ld/1.3.3, Apache-2.0, approved, #8912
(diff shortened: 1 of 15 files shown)


edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/build.gradle.kts
```diff
@@ -19,6 +19,7 @@
 
 plugins {
     `java-library`
+    `maven-publish`
```
ndr-brt, line 22: is this necessary? I think edc-build plugin is already loading the publish plugin

ndr-brt (approved):

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/build.gradle.kts
paullatzelsperger: that is correct - removed it.

wolf4ood (approved):

PR #1140, opened 2024-03-18
feat: add new V2 EDR API

### LINKED ISSUE #1110: Management endpoint to trigger refresh (consumer)

paullatzelsperger:
## WHAT

This PR replicates the `/edrs` API from EDC upstream, but adds the possibility to refresh EDRs automatically when obtaining them.
In addition, the convenience method to negotiated EDRs, that is known from previous versions, was ported over.

## WHY

client applications, that simply use the `/edrs` api need a way to have their tokens automatically refreshed while providing a manual way to refresh tokens.

## FURTHER NOTES
- the old `/edrs` api is still available 
- the new `/edrs` api is available under `/v2/edrs`
- a new test module containing tests for the new EDR API was added
- introduced a `matrix` test strategy for e2e tests

Closes #1110


edc-extensions/edr/edr-api-v2/src/main/java/org/eclipse/tractusx/edc/api/edr/v2/EdrCacheApiController.java
```diff
@@ -0,0 +1,194 @@
...
+            monitor.debug("Token expired, need to refresh.");
+            return tokenRefreshHandler.refreshToken(id, edr);
+        }
+        return ServiceResult.success(edr);
+    }
+
+    // todo: move this method into a service once the "old" EDR api,service,etc. is removed
+    private boolean isExpired(DataAddress edr, org.eclipse.edc.edr.spi.types.EndpointDataReferenceEntry metadata) {
+        var expiresInString = edr.getStringProperty(TX_AUTH_NS + "expiresIn");
+        if (expiresInString == null) {
```
wolf4ood, line 182: probably if the expiration is not present, we cannot trigger a refresh request? 
it could be a non-refreshable token right?

paullatzelsperger: uh... thats right.

wolf4ood (approved):

PR #1208, opened 2024-04-10
docs: add Decision-Record about removing SFTP extensions

paullatzelsperger:
## WHAT

Adds a decision record about removing SFTP.

## WHY

documenting the decision

## FURTHER NOTES

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -125,14 +125,14 @@ jobs:
 
       - name: Run E2E tests (${{ matrix.variant.dir }})
         run: |
-          ./gradlew compileJava compileTestJava
-          ./gradlew -p ${{ matrix.variant.dir }} test -DincludeTags="EndToEndTest"
+          ./gradlew compileJava compileTestJava --refresh-dependencies
+          ./gradlew -p ${{ matrix.variant.dir }} test -DincludeTags="EndToEndTest" 
 
   postgres-tests:
     runs-on: ubuntu-latest
     services:
       postgres:
-        image: postgres:14.2
+        image: postgres:16.1
         ports:
           - 5432:5432
         env:
@@ -143,7 +143,7 @@ jobs:
       - uses: ./.github/actions/setup-java
 
       - name: Run Postgresql E2E tests
-        run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest"
+        run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest" --refresh-dependencies
 
   dataplane-tests:
     runs-on: ubuntu-latest
--- a/docs/development/decision-records/2024-04-11_remove_sftp/README.md
+++ b/docs/development/decision-records/2024-04-11_remove_sftp/README.md
@@ -1,0 +1,14 @@
+# Remove SFTP modules
+
+## Decision
+
+Starting with the next release of Tractus-X EDC, the SFTP modules will be removed from the code base. Specifically `transferprocess-sftp-client`, `transferprocess-sftp-common` and `transferprocess-sftp-provisioner`.
+
+## Rationale
+
+These modules don't implement any official Catena-X Standard, and as such, have not been part of any official Tractus-X EDC distribution for a significant period of time. 
+Further, those modules are not up to the coding standards established by Tractus-X EDC, and would have to be refactored significantly. Even then, having dead code in a repository is bad hygiene, as it generates maintenance churn. Tractus-X EDC is not responsible for maintaining niche extensions, that are not officially sanctioned.
+
+## Approach
+
+Starting with version `0.8.x` of Tractus-X EDC, these modules will be deleted from the code base. At that time, stakeholders interested in continuing to maintain them are welcome to adopt them into their own repositories.


jimmarino (approved):

ndr-brt (approved):
:axe:

florianrusch-zf:
At least ZF would like to have an FTP Dataplane and we are also working on it together with Cofinity-X. We would like to propose a working dataplane as a contribution, but as usual other things have more priority on our side at the moment. 

The reason why we would like the EDC to support another protocol is that S3 (and also Azure Blob) is a proprietary protocol and therefore does not fulfil the needs of companies that want to exchange big data without being forced to use one of the big cloud providers as the storage provider.

\cc @dr-markus

wolf4ood (approved):

florianrusch-zf:
Of course we could "copy/paste" the extensions into our EDC build (which we have already done during testing/development). But we're not just thinking of ourselves. Many small and medium sized businesses won't have a cloud subscription or want to host a full-blown MinIO instance. Most of them have almost no IT department. However, for us as Tractus-X, it makes perfect sense to offer SFTP, which is also an already established standard for data storage/exchange worldwide. Refusing to have this in our "standard" Tractus-X EDC is a big mistake for me and @dr-markus.

Of course I totally agree with you that the current implementation is not sufficiently maintained! But instead of saying "let's get rid of it" or "you can provide it yourselves" we should think about how we can get rid of the maintenance issue and how we can make sure it is maintained as a core Tractus-X EDC protocol in the future.

paullatzelsperger:
Hosting MinIO is not harder than hosting (and properly configuring!) an SFTP server, so I fail to see the validity in that argument. 
My point is that for almost a year, no-one that uses our standard distribution in Tractus-X could have been using SFTP, because we didn't package it. And we have received exactly 0 requests to that effect. Your mileage may vary, which is why I think that functionality is better suited for a ZF/Cofinity variant.

We absolutely cannot keep those modules in the Tx-EDC code base in their current form, because they are poorly implemented and violate several coding and architectural principles, and they pose a potential security hole. 

We discussed removing them already, about a year ago, and the decision back then was to keep them, but to rewrite and improve them, which obviously has not happened. Since it is exclusively you who wants to keep them around, and no-one (including you) has picked up that baton, I can only assume that the priority for you is not high enough. 

I think one year is enough of a grace period. 

The "you can provide it yourself" is key aspect and architectural principle of EDC.

jimmarino:
> Of course we could "copy/paste" the extensions into our EDC build (which we have already done during testing/development). But we're not just thinking of ourselves. Many small and medium sized businesses won't have a cloud subscription or want to host a full-blown MinIO instance. Most of them have almost no IT department. However, for us as Tractus-X, it makes perfect sense to offer SFTP, which is also an already established standard for data storage/exchange worldwide. Refusing to have this in our "standard" Tractus-X EDC is a big mistake for me and @dr-markus.
> 
> Of course, I totally agree with you that the current implementation is not sufficiently maintained! But instead of saying "let's get rid of it" or "you can provide it yourselves," we should think about how we can eliminate the maintenance issue and ensure that it is maintained as a core Tractus-X EDC protocol in the future.

I don't think that contradicts what this DR proposes. A separate repository can be established in TX to maintain and release the extension properly, which would appear to fulfill your requirements. As it stands now, the current code is not properly maintained or up to the standards required for general use. 

I also don't think the use case for small and medium-sized businesses is common. If a business has limited IT staff, it is much cheaper (on the pennies!), less technically challenging, and more secure to use a cloud provider's object storage than to host and properly maintain (S)FTP servers. Properly operating scalable, reliant (S)FTP servers is challenging and typically not something an SMB would invest in.    

The core TX-EDC team has limited resources, and we are trying to deploy them judiciously to cover high-priority use cases. Unfortunately, limited resources force us to make choices. The prioritization seems clear based on the fact that the SFTP extensions have not been maintained or included in a TX distribution for a long time, nor have they been proposed as a C-X standard.

lgblaumeiser:
Hi, @paullatzelsperger @jimmarino,

Actually, we just prepared a contribution to complete the SFTP extension, so that it can be used in a dataplane. It is a planned feature of Dataspace OS and we (Cofinity-X in association with ZF) are willing to maintain the feature in the future. Of course it could be maintained in a different repository, but I would strongly suggest to keep it where it is to simplify the consistency of the extension.

Btw., Minio is in productive use a limited option, due to legal issues and efforts to fulfill legal constraints.

lgblaumeiser:
> Tx-EDC is not a toolbox of random extensions, but a reference implementation. I thought that was understood.

If the Tractus-X repository is dedicated for the task of providing the "official" distribution of the Tractus-X EDC, I understand and support the deletion of the piece. On the other hand, I would see the necessity to have a place to share quality extensions within the community, so that end users can make use of the feature easily. What do you propose as place for something like that? Does this already exist?

Actually, with simpler I meant, that an own repository comes with additional efforts in setting the repository up to provide the necessary quality infrastructure, visibility is decreased, so people who might need the feature won't find it, and finally it is an additional repository to be maintained by the overall project infrastructure.

paullatzelsperger:
This could very well be a separate tractusx repo, that is maintained by ZF/Cof. It is very easy to set that up. The EF has a tool called Otterdog where additional repos can be requested, and we even provide a template repo for that exact purpose, that comes complete with all required legal documents, quality processes etc. It does not get any easier than that.

jimmarino:
> > Tx-EDC is not a toolbox of random extensions, but a reference implementation. I thought that was understood.
> 
> If the Tractus-X repository is dedicated for the task of providing the "official" distribution of the Tractus-X EDC, I understand and support the deletion of the piece. On the other hand, I would see the necessity to have a place to share quality extensions within the community, so that end users can make use of the feature easily. What do you propose as place for something like that? Does this already exist?
> 
> Actually, with simpler I meant, that an own repository comes with additional efforts in setting the repository up to provide the necessary quality infrastructure, visibility is decreased, so people who might need the feature won't find it, and finally it is an additional repository to be maintained by the overall project infrastructure.

A place to share quality extensions or a directory would be a very useful thing to have. We could do something similar to upstream EDC's [Known Friends Directory](https://eclipse-edc.github.io/docs/#/documentation/KNOWN_FRIENDS).

We already have a Github template project that does most of the heavy lifting to set up an extension repository. We could continue to develop the template project to automate this process even further.

lgblaumeiser:
That sounds good, we will discuss the approach and move forward. Thanks, Paul, Jim.

florianrusch-zf:
For me and @dr-markus it totally makes sense to provide sftp as a standard protocol in the tractus-x edc.

> the official Tx-EDC distribution and repository can only contain officially sanctioned and standardized C-X extensions, which SFTP is not.

@paullatzelsperger Where is it documented which EDC extensions are officially standardised by C-X and where is the process documented for proposing an extension to be officially standardised by C-X? And who decides on this?

In the past it was (more or less) decided by us developers or not?

Please don't misunderstand my comments - I agree with you that we should get rid of unused code! I just want to clarify that there is a need for SFTP and that we should clarify how we can proceed with including SFTP as a official supported protocol in the Tractus-X EDC. So for me it would be perfectly fine if we get rid of the existing code if we have a plan for the future.

paullatzelsperger:
I think a good way forward would be to spin up a community or "contrib" repository as outlined in earlier comments, where people can develop and maintain their extensions, and if a strong need arises to adopt one into the standard distribution of EDC, we decide that on a case-by-case basis. We can even do so without moving any code, provided that a Maven publication exists.
I do think that certain acceptance criteria and rules have to be established for that.

jimmarino:
> For me and @dr-markus it totally makes sense to provide sftp as a standard protocol in the tractus-x edc.
> 
> > the official Tx-EDC distribution and repository can only contain officially sanctioned and standardized C-X extensions, which SFTP is not.
> 
> @paullatzelsperger Where is it documented which EDC extensions are officially standardised by C-X and where is the process documented for proposing an extension to be officially standardised by C-X? And who decides on this?
> 
> In the past it was (more or less) decided by us developers or not?
> 
If you want to standardize SFTP, you should join the CX Standardization Working Group and discuss it there. Keep in mind that standardization would require all implementations of the Catena-X wire protocol (including EDC) and all participants to support the feature, so the bar for standardization is high.

PR #1220, opened 2024-04-15
fix(docs): update Helm documentation

paullatzelsperger:
## WHAT

Updates the helm chart documentation to reflect the new config values

## WHY

## FURTHER NOTES


jerzy-wachowiak:
Hi, 

looking into changes in charts/tractusx-connector/README.md.gotmpl line 24+
image

Does it mean, that we can not use Azure Vault any more and have to switch to HashiCorp Vault on short notice?

PR #1264, opened 2024-04-30
feat: add Mock-EDC and a sample how to use it

paullatzelsperger:
## WHAT

This PR adds a stub for EDCs Management API in the form of a standalone Docker image. It enables downstream projects to use this stub as a drop-in replacement for the connector to decrease friction and facilitate easier CI/CD integration.

## WHY

Contract-based testing should be more widely used in Tractus-X.

## FURTHER NOTES

- An additional docker image will published after this PR is merged


docs/development/mock-edc.md
```diff
@@ -0,0 +1,220 @@
+# Using the Mock-EDC for contract-based testing
```
ndr-brt, line 1: Maybe `Mock-Connector` would be a more meaningful name, as EDC looks too generic in this context

ndr-brt (approved):
this is really amazing! Only a comment about the name, all good for the rest.

Issue #670, opened 2023-07-25
Concept for Upgradeability for EDC (ext: CXAR-901)


Phil91:
*Background:*

Current Product EDC is still in a phase of continuous development which leads to frequent enhancements and changes between subsequent versions. One of the reasons for this status is that IDSA protocol (which is the base for the EDC) is still in progress/ definition phase (current version 0.8) and has no official released version. Consequently, this status forces regular, major changes to EDC which are not compatible to previous EDC versions. The effort for data providers/ data consumers updating there connector is currently significant and causes delays in EDC deployment. In addition, different EDC versions lead to incompatibility issues when participants are transferring data using different EDC versions. This situation slows down the scalability of Catena-X network.

*Benefit Hypothesis:*
A concept is required which supports companies which are using EDC to enable a smoother and faster update process and allows better scalability of Catena-X network.

*Short description:*

To bridge time until IDSA protocol version 1.0 is released a concept is required which:
 * offers a step-by-step guide to upgrade EDC
 * provides guardrails for data providers/ data consumers
 * describe general framework to configure and set-up EDC in an optimized way
 * Leverage automated documentation creation from EDC
*  EDC compatibility check before creating a contract offer

_Components_: EDC
_Sprints_: N/A
_Fix Versions_: N/A
_StoryPoints_: N/A
_Attachments_: N/A

stefan-ettl:
Different dimensions
- [ ] SemVer - document the current approach.
- [ ] Answer/Define the term compatibility of EDCs.
- [ ] Concept for compatibility and its dependencies. (Goes beyond EDC, i.e. MIW - consistence voc. across the products)
- [ ] Compatibility check between EDCs (define what counts into it like Policies, Extensions, ...)
- [ ] Answer/Define the term Backwards Compatibility.
- [ ] Create rules compatibility of versions.

gerbigf:
After a chat with an enterprise architect I learned that this requirement comes from Stefan, thus maybe being a requirement from consortia lead. I wasn't able to identify a stakeholder that can give us requirements.

stefan-ettl:
Continue with Features in PI11

PR #1277, opened 2024-05-03
fix: enable resource config & remove unnecessary configs from helm-charts

### LINKED ISSUE #1274: Helm chart contains some bugs
## Describe the bug

While testing the TX-EDC 0.7.0 we identified some problems within the helm chart(s). We just looked at the `tractusx-connector`, but maybe the following points are also valid for the other charts:

- Resources are not indented correctly and are not enabled on default as required by [TRG 5.04](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-04)
- `vault.secretNames` are not used anywhere.

### To Reproduce

Take a deeper look at the helm charts ;-)

### Expected behavior

- Default resources values are provided
- No unnecessary values in the values.yaml

## Context Information

_Add any other context or information about the problem here._

- Used version: EDC 0.7.0

florianrusch-zf:
## WHAT

- Enabling default resource configuration in helm charts
- Removing unnecessary configurations from helm charts
- Improve documentations

## WHY

To fulfil TRG [5.04](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-04) and to avoid confusions

## FURTHER NOTES

Closes #1274


edc-tests/deployment/src/main/resources/helm/tractusx-connector-test.yaml
```diff
@@ -110,6 +105,6 @@ vault:
         /bin/vault kv put secret/aes-keys content=YWVzX2VuY2tleV90ZXN0Cg==
         }
 backendService:
-  httpProxyTokenReceiverUrl: "http://backend:8080"
+  httpProxyTokenReceiverUrl: "http://backend:8080"x
```
paullatzelsperger, line 108:
```suggestion
  httpProxyTokenReceiverUrl: "http://backend:8080"
```

paullatzelsperger (approved):
all good, apart from one rogue "x"

paullatzelsperger (commented):
We cannot change the test value files yet, otherwise the upgradeability test will fail: we install the old (0.5.4) version first, and then do a `helm upgrade`, and for that version we still need the values. 

Once we have released 0.7.1 we can remove those values, because then, the upgradeability check will be 0.7.0 -> 0.7.1.

saschaisele-zf (approved):
Looks good and can be merged after the release of 0.7.1

florianrusch-zf:
Fine for me! And thanks for the review :-) 

@paullatzelsperger Regarding the upgradability check: This check only checks whether the EDC starts successfully after changing the helm chart version, right? I guess it doesn't check if all necessary database migrations are in place, or? Of course, if the migration fails, the pod will not start successfully - but in the past there have been missing migrations that basically resulted in a corrupted database after the upgrade.

related feature: https://github.com/eclipse-tractusx/tractusx-edc/issues/670

wolf4ood:
Hi Folks can we merge this?

paullatzelsperger:
> @paullatzelsperger Regarding the upgradability check: This check only checks whether the EDC starts successfully after changing the helm chart version, right? I guess it doesn't check if all necessary database migrations are in place, or? Of course, if the migration fails, the pod will not start successfully - but in the past there have been missing migrations that basically resulted in a corrupted database after the upgrade.

those upgradeability checks only exist to satisfy [TRG 5.11](https://eclipse-tractusx.github.io/docs/release/trg-5/trg-5-11). The only way to be 100% certain that nothing is broken is to hit every table of the database. We have no plans currently to do that. Apart from that, I would expect any serious production deployment of EDC to provide their own migration feature.

charts/tractusx-connector-azure-vault/values.yaml
```diff
@@ -464,23 +507,27 @@ dataplane:
...
-    memory: 1024Mi
-  requests:
-    cpu: 500m
-    memory: 128Mi
+  resources:
+    limits:
+      # -- Maximum CPU limit
+      cpu: 1.5
+      # -- Maximum memory limit
+      memory: 1024Mi
```
paullatzelsperger, line 522: What's the reason for dataplanes getting 1024Mib by default but control planes only get 512Mib?
Is this based only on "gut feeling" or on a use case/experience/measurement/empirical evidence of some sort?

florianrusch-zf: this needs to be answered by @saschaisele-zf. He configured that initially within https://github.com/eclipse-tractusx/tractusx-edc/pull/794

saschaisele-zf: The values came from experience with the EDC deployment we (ZF) had at that date.

saschaisele-zf (approved):

charts/tractusx-connector-azure-vault/values.yaml
florianrusch-zf: @paullatzelsperger can we close this conversation and merge the PR?

charts/tractusx-connector-azure-vault/templates/deployment-controlplane.yaml
```diff
@@ -289,10 +289,8 @@ spec:
             ## BDRS CLIENT ##
             #################
 
-            {{- if .Values.controlplane.bdrs.server.url }}
             - name: "TX_IAM_IATP_BDRS_SERVER_URL"
               value: {{ .Values.controlplane.bdrs.server.url | required ".Values.controlplane.bdrs.server.url is required" | quote }}
-            {{- end }}
```
florianrusch-zf, line 295: @paullatzelsperger I'm not sure if this was implemented on purpose back then. Do you possibly know?

florianrusch-zf: I already [asked in matrix](https://matrix.to/#/!mYxOilDPMLCQhMoIVc:matrix.eclipse.org/$mYmAhWWFSODdIgAzVb5ezMMUJFLrWRFc8iYffr4Y2ms?via=matrix.eclipse.org&via=matrix.org&via=dev-null.rocks), but haven't received any answer

charts/tractusx-connector-azure-vault/values.yaml
paullatzelsperger: I am generally ok with it, but we either use identical resources, or we put that information somewhere: documentation, a source comment,.... because the next developer will not have that "hidden" empirical knowledge

charts/tractusx-connector-azure-vault/templates/deployment-controlplane.yaml
paullatzelsperger: I don't recall :(
Since the BDRS Url is mandatory anyway, I think we could remove the `if`

paullatzelsperger (approved):

paullatzelsperger:
@florianrusch-zf I think you gotta rebase

charts/tractusx-connector-azure-vault/values.yaml
florianrusch-zf: Then let's just use the same value ;-)

florianrusch-zf: done. Please feel free to resolve the conversation and reapprove ;-)

paullatzelsperger (approved):

PR #1300, opened 2024-05-13
chore: disable dim int test

wolf4ood:
## WHAT

disable DIM int test

## WHY

currenty try to run it in the `verify-dim` workflow but it fails for missing configuration. For now we just comment it as `@Test` till we decide to configure it or delete it 

## FURTHER NOTES

### DIFF
--- a/edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/DimCatalogIntegrationTest.java
+++ b/edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/DimCatalogIntegrationTest.java
@@ -22,12 +22,9 @@ package org.eclipse.tractusx.edc.tests.catalog;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.eclipse.tractusx.edc.lifecycle.DimParticipant;
 import org.eclipse.tractusx.edc.lifecycle.ParticipantRuntime;
-import org.eclipse.tractusx.edc.tag.DimIntegrationTest;
 import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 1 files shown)


edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/DimCatalogIntegrationTest.java
```diff
@@ -98,7 +95,7 @@ static void afterAll() {
         bdrsServer.stop();
     }
 
-    @Test
+    //@Test
```
ndr-brt, line 98: Could it be just `@Disabled`?

ndr-brt (approved):

edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/DimCatalogIntegrationTest.java
wolf4ood: Unfortunatelly no, since this depends on environment variable being there and it still fails in initialization even if it's disabled.

ndr-brt: that's a bummer! LGTM the way it is then!

wolf4ood: Yeah it's ugly, it's marked as removal anyway i doubt that we will have a config for running this nightly

PR #1329, opened 2024-05-23
refactor: deprecate dataplane-selector-configuration in favor of self-registration

### LINKED ISSUE #1325: replace `dataplane-selector-configuration` with data-plane self registration
## WHAT

After EDC 0.6.4 there's a new `data-plane-self-registration` extension available that permits the data-plane to register itself to the control plane, this will make `dataplane-selector-configuration` obsolete.

## WHY
use upstream solutions

## FURTHER NOTES

ndr-brt:
## WHAT

Deprecate `dataplane-selector-configuration` and remove it from the control-plane runtimes, and add `data-plane-self-registration` on the data-plane ones.

## WHY

data plane self-registration

## FURTHER NOTES

- this change should not not break and it should permit flawless update

Closes #1325


edc-dataplane/edc-dataplane-base/build.gradle.kts
```diff
@@ -34,7 +34,7 @@ dependencies {
     runtimeOnly(libs.edc.core.did) // for the DID Public Key Resolver
     runtimeOnly(libs.edc.identity.did.web)
     runtimeOnly(libs.edc.config.filesystem)
-    runtimeOnly(libs.edc.auth.tokenbased)
+//    runtimeOnly(libs.edc.auth.tokenbased)
```
paullatzelsperger, line 37:
```suggestion
```

paullatzelsperger (approved):

edc-dataplane/edc-dataplane-base/build.gradle.kts
ndr-brt: yes, I was just discovering that for some reasons I got 401 on the readiness check, commenting that solves the problem, in any case data-plane api are only supposed to be called by control plane so they they don't have to be secured (for the moment), I just removed that dependency because it does not make sense as it is supposed to secure `management-api` endpoints.

paullatzelsperger: the Observability API (which contains the readiness check) should not be secured, but it also should never be exposed outside of the pod.

yes, commenting (or better: removing) this line may solve the problem, but if somehow the `TokenBasedAuthenticationService` fires for the Observability API, that would be a bug

PR #1355, opened 2024-06-10
fix: adopt EDC test runtime model

paullatzelsperger:
## WHAT

This PR adopts the new EDC test runtime model

## WHY

keep up-to-date with recent developments, decrease test runtime

## FURTHER NOTES


wolf4ood (approved):

edc-tests/e2e-tests/src/test/java/org/eclipse/tractusx/edc/lifecycle/ParticipantRuntime.java
```diff
@@ -62,30 +59,25 @@ public void afterTestExecution(ExtensionContext context) {
     }
 
     @Override
-    public void beforeAll(ExtensionContext context) throws Exception {
+    public void beforeAll(ExtensionContext context) {
         //only run this once
         super.beforeTestExecution(context);
```
ndr-brt, line 64: I think this class should either override `beforeTestExecution` or extend the `RuntimePerClassMethodExtension`

paullatzelsperger:
merging, as @ndr-brt's only comment was addressed, and there is one other approval.

PR #1365, opened 2024-06-18
feat(build): bump EDC version to latest nightly

paullatzelsperger:
## WHAT

bumps the EDC version to the latest nightly, and fixes several compile errors.
Also fixes a small bug where the request header `Content-Type` was not set in the `TokenRefreshHandler`

## WHY

avoid technical debt

## FURTHER NOTES


wolf4ood (approved):

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
```diff
@@ -178,6 +178,7 @@ private Result<Request> createTokenRefreshRequest(String refreshEndpoint, String
 
         return success(new Request.Builder()
                 .addHeader("Authorization", bearerToken)
+                .addHeader("Content-Type", "application/x-www-form-urlencoded")
```
florianrusch-zf, line 181: Why does it need to be a `application/x-www-form-urlencoded`? And why does the body need to be a `new byte[0]`? Why can't it be empty?

florianrusch-zf: Okay, found the answer for the first question: OAuth2 standard

ndr-brt (approved):

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
paullatzelsperger: the header is needed because of the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1)
the body is there because post request must have a body, at least if we're using OkHttp

Issue #1393, opened 2024-07-03
Published swagger documentation is not complete


florianrusch-zf:
## Describe the bug

Since TX-EDC 0.7.3 the [published swagger documentation](https://app.swaggerhub.com/apis/eclipse-tractusx-bot/tractusx-edc/0.7.3) doesn't contain the upstream APIs.

### Expected behavior

All available APIs in TX-EDC are listed in the published swagger documentation.

## Hint

[@ndr-brt mentioned on matrix](https://matrix.to/#/!mYxOilDPMLCQhMoIVc:matrix.eclipse.org/$UV6Fe9WqoHFynbSOyxWwPi0wosvAABjVVNrA46Aa7Z4?via=matrix.eclipse.org&via=matrix.org&via=dev-null.rocks): 

> That happened because upstream changed the way api are versioned and uploaded (not on swaggerhub anymore).

They are publishing it now on their GitHub Page: https://eclipse-edc.github.io/docs/#/submodule/Connector/docs/developer/

ndr-brt:
The short term fix would be to point to the latest management api openapi spec.
Long term fix is to generate the openapi spec from the used modules in the runtime (upstream issue: https://github.com/eclipse-edc/GradlePlugins/issues/244 )

PR #1402, opened 2024-07-05
docs(api): get openapi files from dependencies repository

ndr-brt:
## WHAT

Add a gradle task that downloads all the openapi specs of the dependencies and use them to build the merged spec to be uploaded to swaggerhub

## WHY

documentation

## FURTHER NOTES
- this will need to be refined further because currently downloads ALL the openapi spec, also the ones that likely we don't want to show in the docs (e.g. dsp, control...). at the moment there's no way to discriminate.

Part of #1393


edc-controlplane/edc-controlplane-base/build.gradle.kts
```diff
@@ -77,3 +77,34 @@ dependencies {
     runtimeOnly(libs.edc.controlplane.callback.dispatcher.http)
 
 }
+
+tasks.register("downloadOpenapi") {
```
paullatzelsperger, line 81: this should probably be in the root build file, because the dataplane also has an API and thus openAPI files

ndr-brt: I was thinking, do we need to list the dataplane api as well? Because those should be considered internal ones, same as "control" and "dsp", that's why I'm working to filter them out (update to PR will come up soon).
likely to external viewers only "management" and "observability" would be interesting, wdyt?

otherwise a possibility could be to publish them separately as we're doing on upstream

ndr-brt: not "external", but "internal". 
token refresh is already part of the merged spec because it's not downloaded from maven repo but generated by the `resolve` task.

anyway I will add the task for both control-plane and data-plane, for `management-api`, `observability-api` and `public-api`, then let's wait for some feedback

paullatzelsperger: sorry, i meant to say that I would not consider them _internal_ ofc. 

> token refresh is already part of the merged spec because it's not downloaded from maven repo but generated by the resolve task.

i know, my point was that the dataplane very much has APIs that are public, such as DPS and token refresh. especially with DPS i see no reason _not_ to include it.

paullatzelsperger (approved):

PR #1431, opened 2024-07-18
docs: decisio-record abour versioning scheme

paullatzelsperger:
## WHAT

Adds a decision record outlining changes to the versioning scheme and release cadence.

## WHY

documentation, transparency

## FURTHER NOTES

### DIFF
--- a/docs/development/decision-records/2024-07-18-Improvements-versioning-scheme/README.md
+++ b/docs/development/decision-records/2024-07-18-Improvements-versioning-scheme/README.md
@@ -1,0 +1,66 @@
+# Improvements to the Tractus-X EDC versioning scheme
+
+## Decision
+
+Starting with `0.8.x`, Tractus-X EDC will publish an official release roughly every 12 weeks (once per quarter).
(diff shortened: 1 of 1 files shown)


ndr-brt (approved):
Only note: upstream currently doesn't have such a clear versioning scheme, so referring to the the point 4 of the `Backporting fixed`:
> Tractus-X EDC must request that an upstream bugfix version be released. This upstream bugfix can then be incorporated into the Tractus-X EDC bugfix version.

currently this won't be achievable (in an easy way), so maybe it worth discussing about this on upstream as well

paullatzelsperger:
> Only note: upstream currently doesn't have such a clear versioning scheme, so referring to the the point 4 of the `Backporting fixed`:
> 
> > Tractus-X EDC must request that an upstream bugfix version be released. This upstream bugfix can then be incorporated into the Tractus-X EDC bugfix version.
> 
> currently this won't be achievable (in an easy way), so maybe it worth discussing about this on upstream as well

there isn't a process for it, but it has been done in the past

Issue #1411, opened 2024-07-09
BPNL Validation function evaluates the same way for EQ and isAllOf operators


rafaelmag110:
## Describe the bug

Using the `isAllOf` operator in a policy, the validation fails if a BPNL belongs to more groups than the ones specified in the policy. 

In the BPNL validation extension, the validation function evaluates the `EQ` and the `isAllOf` operators in the same way, however even the documentation states they should evaluate the groups differently, as shown in the expected behaviour.

Also, in the `BusinessPartnerGroupFunctionTest` class, the following circumstance is tested:
`
Arguments.of("Overlapping groups", IS_ALL_OF, List.of(TEST_GROUP_1, TEST_GROUP_2, "different-group", "another-different-group"), false)`

And being `TEST_GROUP_1` and `TEST_GROUP_2` part of the allowed groups, this case should return `true` and not `false`

### Expected behavior

Operator `EQ`: must be exactly in - and only in - that particular group or set of groups
Operator `IS_ALL_OF`: must be in all of the specified groups

danielkenji01:
Can I assign this issue to me? @ndr-brt @rafaelmag110

lgblaumeiser:
Please have a look at: https://github.com/eclipse-tractusx/sig-release/issues/775#issuecomment-2242235805, there is an additional issue with operators

danielkenji01:
Issue was solved on the following Pull Request: https://github.com/eclipse-tractusx/tractusx-edc/pull/1456

PR #775, opened 2023-09-19
feat: update to EDC v0.3.0

paullatzelsperger:
## WHAT

Upgrades to EDC v0.3.0 (milestone 11)

## WHY

keep up with recent developments

## FURTHER NOTES

_List other areas of code that have changed but are not necessarily linked to the main feature. This could be method signature changes, package declarations, bugs that were encountered and were fixed inline, etc._

Closes # <-- _insert Issue number if one exists_

PR #1456, opened 2024-07-23
Fix: Adjusted BPN validations on different operators

danielkenji01:
## WHAT
- Fixed **IS_ALL_OF** operator validation. It had the same behavior of **EQ** operator
- Fixed **IS_NONE_OF** operator validation. It had the same behavior of the **NEQ** operator
- Fixed **NEQ** operator validation. It was considering the opposite result of the IN operator, and it should be the opposite of the EQ operator
- Added a new rule when the BPN groups or assigned groups are empty, and if the operator is **NEQ** or **IS_NONE_OF**, it returns true instead of false

## WHY
In order to keep the correct behavior on BPNL validations

## FURTHER NOTES
Some minor adjusts reported by Sonar, including:
- Adjusted String format and casting

Closes #https://github.com/eclipse-tractusx/tractusx-edc/issues/1411
Closes #https://github.com/eclipse-tractusx/tractusx-edc/issues/775

### DIFF
--- a/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
+++ b/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
@@ -137,7 +137,7 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
 
         // right-operand is anything other than String or Collection
         var rightOperand = parseRightOperand(rightValue, policyContext);
-        if (rightOperand.isEmpty()) {
+        if (rightOperand == null) {
             return false;
         }
 
@@ -145,8 +145,7 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         return OPERATOR_EVALUATOR_MAP.get(operator).apply(new BpnGroupHolder(assignedGroups, rightOperand));
     }
 
-    private boolean evaluateEmptyGroupsResponse(final Operator operator, final PolicyContext policyContext,
-                                                final String problemMessage) {
+    private boolean evaluateEmptyGroupsResponse(Operator operator, PolicyContext policyContext, String problemMessage) {
         if (NEQ.equals(operator) || IS_NONE_OF.equals(operator)) {
             return true;
         }
@@ -165,7 +164,7 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         }
 
         context.reportProblem(format("Right operand expected to be either String or a Collection, but was %s", rightValue.getClass()));
-        return Collections.emptyList();
+        return null;
     }
     
     private Boolean evaluateIn(BpnGroupHolder bpnGroupHolder) {
--- a/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
+++ b/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
@@ -24,6 +24,8 @@ import org.eclipse.edc.policy.engine.spi.PolicyContext;
 import org.eclipse.edc.policy.model.Operator;
 import org.eclipse.edc.policy.model.Permission;
 import org.eclipse.edc.spi.agent.ParticipantAgent;
+import org.eclipse.edc.spi.result.StoreFailure;
+import org.eclipse.edc.spi.result.StoreFailure.Reason;
 import org.eclipse.tractusx.edc.validation.businesspartner.spi.BusinessPartnerStore;
 
 import java.util.Arrays;
@@ -78,7 +80,6 @@ import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.TX_NAMESPACE;
 public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Permission> {
     public static final String BUSINESS_PARTNER_CONSTRAINT_KEY = TX_NAMESPACE + "BusinessPartnerGroup";
     private static final List<Operator> ALLOWED_OPERATORS = List.of(EQ, NEQ, IN, IS_ALL_OF, IS_ANY_OF, IS_NONE_OF);
-    private static final List<Operator> POSITIVE_OPERATORS = List.of(EQ, IN, IS_ALL_OF, IS_ANY_OF);
     private static final Map<Operator, Function<BpnGroupHolder, Boolean>> OPERATOR_EVALUATOR_MAP = new HashMap<>();
     private final BusinessPartnerStore store;
 
@@ -87,8 +88,8 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         OPERATOR_EVALUATOR_MAP.put(EQ, this::evaluateEquals);
         OPERATOR_EVALUATOR_MAP.put(NEQ, this::evaluateNotEquals);
         OPERATOR_EVALUATOR_MAP.put(IN, this::evaluateIn);
-        OPERATOR_EVALUATOR_MAP.put(IS_ALL_OF, this::evaluateIsAllOf);
-        OPERATOR_EVALUATOR_MAP.put(IS_ANY_OF, this::evaluateIn);
+        OPERATOR_EVALUATOR_MAP.put(IS_ALL_OF, this::evaluateIn);
+        OPERATOR_EVALUATOR_MAP.put(IS_ANY_OF, this::evaluateIsAnyOf);
         OPERATOR_EVALUATOR_MAP.put(IS_NONE_OF, this::evaluateIsNoneOf);
     }
 
@@ -98,8 +99,7 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
      * <ul>
      *     <li>No {@link ParticipantAgent} was found on the {@link PolicyContext}</li>
      *     <li>The operator is invalid. Check {@link BusinessPartnerGroupFunction#ALLOWED_OPERATORS} for valid operators.</li>
-     *     <li>No database entry was found for the BPN (taken from the {@link ParticipantAgent}) and the operator is 'positive' {@link BusinessPartnerGroupFunction#POSITIVE_OPERATORS}</li>
-     *     <li>A database entry was found, but no group-IDs were assigned and the operator is 'positive' {@link BusinessPartnerGroupFunction#POSITIVE_OPERATORS}</li>
+     *     <li>No database entry was found for the BPN (taken from the {@link ParticipantAgent}) and the {@link StoreFailure#getReason()} is different than {@link Reason#NOT_FOUND}</li>
      *     <li>The right value is anything other than {@link String} or {@link Collection}</li>
      * </ul>
      */
@@ -123,21 +123,17 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         var bpn = participantAgent.getIdentity();
         var groups = store.resolveForBpn(bpn);
 
-        var isPositiveOperator = POSITIVE_OPERATORS.contains(operator);
-        var errorMessage = "";
+        var assignedGroups = groups.getContent();
 
         // BPN not found in database
         if (groups.failed()) {
-            errorMessage = groups.getFailureDetail();
+            boolean isNotFound = Reason.NOT_FOUND.equals(groups.getFailure().getReason());
+            if (!isNotFound) {
+                policyContext.reportProblem(groups.getFailureDetail());
+                return false;
+            }
 
-        // BPN was found, but it does not have groups assigned.
-        } else if (groups.getContent().isEmpty()) {
-            errorMessage = "No groups were assigned to BPN " + bpn;
-        }
-
-        if (!errorMessage.isBlank() && isPositiveOperator) {
-            policyContext.reportProblem(errorMessage);
-            return false;
+            assignedGroups = List.of();
         }
 
         // right-operand is anything other than String or Collection
@@ -145,8 +141,6 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         if (rightOperand == null) {
             return false;
         }
-
-        var assignedGroups = groups.getContent();
 
         //call evaluator function
         return OPERATOR_EVALUATOR_MAP.get(operator).apply(new BpnGroupHolder(assignedGroups, rightOperand));
@@ -168,17 +162,10 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
     private Boolean evaluateIn(BpnGroupHolder bpnGroupHolder) {
         var assigned = bpnGroupHolder.assignedGroups;
         // checks whether both lists overlap
-        return bpnGroupHolder.allowedGroups
-                .stream()
-                .distinct()
-                .anyMatch(assigned::contains);
+        return bpnGroupHolder.allowedGroups.containsAll(assigned);
     }
 
     private Boolean evaluateNotEquals(BpnGroupHolder bpnGroupHolder) {
-        if (bpnGroupHolder.allowedGroups.isEmpty()) {
-            return true;
-        }
-
         return !evaluateEquals(bpnGroupHolder);
     }
 
@@ -186,24 +173,20 @@ public class BusinessPartnerGroupFunction implements AtomicConstraintFunction<Pe
         return bpnGroupHolder.allowedGroups.equals(bpnGroupHolder.assignedGroups);
     }
 
-    private boolean evaluateIsAllOf(BpnGroupHolder bpnGroupHolder) {
-        if (bpnGroupHolder.allowedGroups.isEmpty()) {
-            return false;
-        }
-
-        var assigned = bpnGroupHolder.assignedGroups;
-        return bpnGroupHolder.allowedGroups
-                .stream()
-                .distinct()
-                .allMatch(assigned::contains);
-    }
-
-    private boolean evaluateIsNoneOf(BpnGroupHolder bpnGroupHolder) {
-        if (bpnGroupHolder.assignedGroups == null) {
+    private boolean evaluateIsAnyOf(BpnGroupHolder bpnGroupHolder) {
+        if (bpnGroupHolder.allowedGroups.isEmpty() && bpnGroupHolder.assignedGroups.isEmpty()) {
             return true;
         }
 
-        return !evaluateIn(bpnGroupHolder);
+        var allowedGroups = bpnGroupHolder.allowedGroups;
+        return bpnGroupHolder.assignedGroups
+                .stream()
+                .distinct()
+                .anyMatch(allowedGroups::contains);
+    }
+
+    private boolean evaluateIsNoneOf(BpnGroupHolder bpnGroupHolder) {
+        return !evaluateIsAnyOf(bpnGroupHolder);
     }
 
     /**
--- a/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
+++ b/edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
@@ -24,7 +24,6 @@ import org.eclipse.edc.policy.engine.spi.PolicyContext;
 import org.eclipse.edc.policy.model.Operator;
 import org.eclipse.edc.policy.model.Permission;
 import org.eclipse.edc.spi.agent.ParticipantAgent;
-import org.eclipse.edc.spi.result.StoreResult;
 import org.eclipse.tractusx.edc.validation.businesspartner.spi.BusinessPartnerStore;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -120,44 +120,52 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
-        if (rightOperand == null) {
+        if (rightOperand.isEmpty()) {
             return false;
         }
 
         //call evaluator function
         return OPERATOR_EVALUATOR_MAP.get(operator).apply(new BpnGroupHolder(assignedGroups, rightOperand));
     }
 
+    private boolean evaluateEmptyGroupsResponse(final Operator operator, final PolicyContext policyContext,
```
paullatzelsperger, line 148: this method is quite obscure, from reading the signature it is not clear which combinations of parameters produce which result. Its fine to have a method to encapsulate error handling, but this is not whats happening here.
The method name is misleading as well. Please rewrite this, and a word of documentation can't hurt either.

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -120,44 +120,52 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
             return Arrays.asList(tokens);
         }
         if (rightValue instanceof Collection<?>) {
             return ((Collection<?>) rightValue).stream().map(Object::toString).toList();
         }
 
-        context.reportProblem(format("Right operand expected to be either String or a Collection, but was " + rightValue.getClass()));
-        return null;
+        context.reportProblem(format("Right operand expected to be either String or a Collection, but was %s", rightValue.getClass()));
+        return Collections.emptyList();
```
paullatzelsperger, line 168: is there a reason for this? how can you then distinguish between an empty array `"[]"` and an invalid right operand?

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -82,14 +85,14 @@ void setUp() {
     void evaluate_noParticipantAgentOnContext() {
         reset(context);
         assertThat(function.evaluate(EQ, "test-group", createPermission(EQ, List.of()), context)).isFalse();
-        verify(context).reportProblem(eq("ParticipantAgent not found on PolicyContext"));
+        verify(context).reportProblem("ParticipantAgent not found on PolicyContext");
```
paullatzelsperger, line 88: please don't just change unrelated code. Supplying a matcher is good practice IMO, because it increases expressivity.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -82,14 +85,14 @@ void setUp() {
...
     }
 
     @ParameterizedTest(name = "Invalid operator {0}")
     @ArgumentsSource(InvalidOperatorProvider.class)
     @DisplayName("Invalid operators, expect report in policy context")
     void evaluate_invalidOperator(Operator invalidOperator) {
-        when(context.getContextData(eq(ParticipantAgent.class))).thenReturn(new ParticipantAgent(Map.of(), Map.of()));
+        when(context.getContextData(ParticipantAgent.class)).thenReturn(new ParticipantAgent(Map.of(), Map.of()));
```
paullatzelsperger, line 95: see above.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -98,7 +101,7 @@ void evaluate_invalidOperator(Operator invalidOperator) {
     @DisplayName("Right-hand operand is not String or Collection<?>")
     void evaluate_rightOperandNotStringOrCollection() {
         when(store.resolveForBpn(TEST_BPN)).thenReturn(StoreResult.success(List.of("test-group")));
-        when(context.getContextData(eq(ParticipantAgent.class))).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
+        when(context.getContextData(ParticipantAgent.class)).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
```
paullatzelsperger, line 104: ditto

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -117,29 +120,51 @@ void evaluate_rightOperandNotStringOrCollection() {
     void evaluate_validOperator(String ignored, Operator operator, List<String> assignedBpn, boolean expectedOutcome) {
 
         var allowedGroups = List.of(TEST_GROUP_1, TEST_GROUP_2);
-        when(context.getContextData(eq(ParticipantAgent.class))).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
+        when(context.getContextData(ParticipantAgent.class)).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
```
paullatzelsperger, line 123: ditto

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -117,29 +120,51 @@ void evaluate_rightOperandNotStringOrCollection() {
...
         when(store.resolveForBpn(TEST_BPN)).thenReturn(StoreResult.success(assignedBpn));
         assertThat(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context)).isEqualTo(expectedOutcome);
     }
 
-    @Test
-    void evaluate_noEntryForBpn() {
-        var operator = NEQ;
+    @EnumSource(value = Operator.class, names = {"NEQ", "IS_NONE_OF"})
+    @ParameterizedTest
+    void evaluate_noEntryForBpnWithNegativeOperator_shouldBeTrue(final Operator operator) {
```
paullatzelsperger, line 130: please no `final`s in method signatures, local vars, etc. `final` is only allowed in field declarations. check out our coding guidelines.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -117,29 +120,51 @@ void evaluate_rightOperandNotStringOrCollection() {
...
         var allowedGroups = List.of(TEST_GROUP_1, TEST_GROUP_2);
-        when(context.getContextData(eq(ParticipantAgent.class))).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
+        when(context.getContextData(ParticipantAgent.class)).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
         when(store.resolveForBpn(TEST_BPN)).thenReturn(StoreResult.notFound("foobar"));
 
-        assertThat(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context)).isFalse();
+        assertTrue(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context));
```
paullatzelsperger, line 135: please don't change existing code without good reason. we use assertJ assertions to keep consistency. goes for the whole file.

paullatzelsperger (changes requested):
TL;DR: don't just change unrelated code, don't use `final` except for field declarations

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
danielkenji01: It was more of a suggestion from Sonar to return an empty list instead of a null value, then I changed the validation after the method to verify if the list is empty.

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
danielkenji01: I refactored the method and added a documentation to be more clear

danielkenji01:
> TL;DR: don't just change unrelated code, don't use `final` except for field declarations

Thank you for the comments, I followed them and adjusted the code.

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -120,20 +121,13 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
         var bpn = participantAgent.getIdentity();
         var groups = store.resolveForBpn(bpn);
 
-        // BPN not found in database
-        if (groups.failed()) {
-            policyContext.reportProblem(groups.getFailureDetail());
-            return false;
-        }
-        var assignedGroups = groups.getContent();
+        var errorMessage = evaluateEmptyGroups(groups, operator, bpn);
```
paullatzelsperger, line 127: results should not be passed into methods. ever. they have a fluent API, so please use it. at this point though, the `evaluateEmptyGroups` method is not needed and its content can be inlined here.

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -143,20 +137,57 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
+     */
+    private String evaluateEmptyGroups(StoreResult<List<String>> groups, Operator operator, String bpn) {
+        var isNegativeOperator = NEGATIVE_OPERATORS.contains(operator);
+
+        if (isNegativeOperator) {
+            return null;
+        }
+
+        // BPN not found in database
+        if (groups.failed()) {
```
paullatzelsperger, line 167: like I remarked above, `Result`s should be passed into and evaluate in methods. The code flow of this method and its caller can be simplified/unified and made more readable.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -122,24 +124,46 @@ void evaluate_validOperator(String ignored, Operator operator, List<String> assi
...
+    @ParameterizedTest
+    void evaluate_noEntryForBpnWithNegativeOperator_shouldBeTrue(Operator operator) {
+        var allowedGroups = List.of(TEST_GROUP_1, TEST_GROUP_2);
+        when(context.getContextData(eq(ParticipantAgent.class))).thenReturn(new ParticipantAgent(Map.of(), Map.of(PARTICIPANT_IDENTITY, TEST_BPN)));
+        when(store.resolveForBpn(TEST_BPN)).thenReturn(StoreResult.notFound("foobar"));
+
+        assertThat(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context)).isTrue();
+    }
+
+    @EnumSource(value = Operator.class, names = {"NEQ", "IS_NONE_OF"})
```
paullatzelsperger, line 137: Test coverage is insufficient. you're only testing one scenario, i.e. left-hand operand is empty, right-hand operand is non-empty.
What about `[a,b,c] NEQ []`, `[a,b,c] IS_NONE_OF []`? same goes for the "positive" operator test.

paullatzelsperger (changes requested):

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
danielkenji01: I simplified the validation and removed the method.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
danielkenji01: I added the test scenarios.

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -120,20 +120,23 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
-            policyContext.reportProblem(groups.getFailureDetail());
-            return false;
-        }
-        var assignedGroups = groups.getContent();
+            errorMessage = groups.getFailureDetail();
 
         // BPN was found, but it does not have groups assigned.
-        if (assignedGroups.isEmpty()) {
-            policyContext.reportProblem("No groups were assigned to BPN " + bpn);
+        } else if (groups.getContent().isEmpty()) {
```
paullatzelsperger, line 134: this code can be simplified, since an empty result is only an error if the operator is "positive", so your if-else cascades should reflect that in the form of a nested if-else: 
```java
if(result.isFailed()){
  policyContext.reportProblem(result.getFailureDetail());
  return false;
}
var groups = result.getGroups();
if (isPositiveOperator && groups.isEmpty()) {
    policyContext.reportProblem("No groupse assigned...");
    return false;
}

```

paullatzelsperger (changes requested):

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
danielkenji01: Regarding the first question: even though I assign an error message to the variable, I just report the error if it is a positive operator, as it is validated on the line 138. 
About the other test scenarios, I will add them here.

paullatzelsperger: no, i'm saying that an empty assigned group is _not_ an error, because if the policy states ` eq []`, then a BPN that has no groups assigned to it would satisfy that condition. 

Generally though, even if the code were semantically correct, the control flow is too complicated, there are duplicated fragments.

danielkenji01: So, in this case, would you agree that both _groups.failed()_ and _groups.getContent().isEmpty()_ validations wouldn't make sense anymore? Since in both cases, we could consider the result as an empty list, then validate directly on the operator evaluators. 

For example, we would have:
- `[] eq []` = true
- `[] isAllOf []` = true
- `[] isAnyOf []` = false
- `[] in []` = false

paullatzelsperger: nope, that's incorrect. there is a fundamental difference between a `Result.failure` and an empty list. A `failure` could be caused e.g. by a database error or something. That is an error situation. An empty list is not an error situation, because a particular BPN might simply not have any groups assigned. There could still be constraints that evaluate to `true`. 

I also disagree with your examples, because `[] isAnyOf []` should be true. Any element on the left side (Which is none) is equal to any element of the right side (which is also none). And `[] in []` is also true, because the entire list on the left side (which is empty) is contained fully in the right side (also being empty).

If you have no apples, and I ask you to give me all your apples, and I end up with no apples - did you give me everything you had?

danielkenji01: Okay, so just this validation `groups.getContent().isEmpty()` would not make sense anymore? 
In regard to the validations from each operator, I will check and fix any inconsistent behavior.

danielkenji01: @paullatzelsperger I don't know why, but I'm not being able to re-request a review, as you asked me to do after every conversation solved on this PR.

edc-extensions/bpn-validation/bpn-validation-core/src/test/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunctionTest.java
```diff
@@ -115,31 +117,45 @@ void evaluate_rightOperandNotStringOrCollection() {
...
-        assertThat(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context)).isFalse();
+        assertThat(function.evaluate(operator, allowedGroups, createPermission(operator, allowedGroups), context)).isTrue();
+        verify(context, never()).reportProblem("foobar");
+    }
+
+    @ArgumentsSource(OperatorForEmptyGroupsProvider.class)
+    @ParameterizedTest
+    void evaluate_groupsAssignedButNoGroupsSentToEvaluate(Operator operator, List<String> assignedBpnGroups,
+                                                          boolean expectedOutcome) {
+        var allowedGroups = List.<String>of();
```
paullatzelsperger, line 153: nit:
```suggestion
        List<String> allowedGroups = List.of();
```

paullatzelsperger (changes requested):

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
```diff
@@ -148,35 +142,48 @@ public boolean evaluate(Operator operator, Object rightValue, Permission rule, P
...
-        return !evaluateIn(bpnGroupHolder);
+        return !evaluateEquals(bpnGroupHolder);
     }
 
     private Boolean evaluateEquals(BpnGroupHolder bpnGroupHolder) {
         return bpnGroupHolder.allowedGroups.equals(bpnGroupHolder.assignedGroups);
     }
 
+    private boolean evaluateIsAnyOf(BpnGroupHolder bpnGroupHolder) {
+        if (bpnGroupHolder.allowedGroups.isEmpty() && bpnGroupHolder.assignedGroups.isEmpty()) {
```
paullatzelsperger, line 172: not necessary, this is covered by the stream operation.

paullatzelsperger (changes requested):

edc-extensions/bpn-validation/bpn-validation-core/src/main/java/org/eclipse/tractusx/edc/validation/businesspartner/functions/BusinessPartnerGroupFunction.java
danielkenji01: It doesn't cover, that why I had to add this separate empty lists validation. I can try to find another why to incorporate this validation with the stream() methods, but I still don't know how. I accept any suggestions that you might have for this case.

paullatzelsperger (approved):

PR #1460, opened 2024-07-25
feat: remove project root from the docker image

### LINKED ISSUE #1436: published docker images contain all the project tree
## WHAT

We're publishing docker images that are containing all the project tree, instead of just what's needed (jars, legal docs).
This could be caused by the way we're calling the `build-push-action`:
https://github.com/eclipse-tractusx/tractusx-edc/blob/61c4af152f29aec0723fc43f46a5c467787bf7cf/.github/actions/publish-docker-image/action.yml#L102-L112

To replicate, just run a remote tx-edc image locally and print `ls -1` in it.

## HOW

we should strip down the image to contain only what's needed.

## FURTHER NOTES

ndr-brt:
## WHAT

Avoid to put the whole content of the project into the docker image, aligning the actions image build process with the `dockerize` task.

## WHY

optimization

## FURTHER NOTES

further cleanups:
- removed not-needed `do_push` parameter from the `publish-docker-image` action
- cleaned up the `downloadOtel` task using a more "gradle-y" approach
- add opentelemetry version on the version catalog, so it will be kept updated by dependabot
- removed not necessary exclusion from the shadowJar config

Closes #1436


build.gradle.kts
```diff
@@ -140,77 +140,62 @@ allprojects {
...
+                .dependsOn(copyDockerfile)
+                .dependsOn(copyLegalDocs)
+                .dependsOn(downloadOpentelemetryAgent)
 
             //actually apply the plugin to the (sub-)project
             apply(plugin = "com.bmuschko.docker-remote-api")
 
             val dockerTask: DockerBuildImage = tasks.create("dockerize", DockerBuildImage::class) {
-                dockerFile.set(project.layout.buildDirectory.asFile.get().toPath().resolve("resources").resolve("docker").resolve("Dockerfile").toFile())
+                dockerFile.set(File("build/resources/docker/Dockerfile"))
```
paullatzelsperger, line 182: this will likely cause problems on windows. I know we don't officially support it, but must we intentionally break it?

paullatzelsperger (commented):
to be honest, I don't quite get the issue. The `Dockerfile` only copies the runtime JAR, the opentelemetry JAR and the additional files, i.e. legal docs?

ndr-brt:
> to be honest, I don't quite get the issue. The `Dockerfile` only copies the runtime JAR, the opentelemetry JAR and the additional files, i.e. legal docs?

the root of the problem is that, if you don't specify the `ADDITIONAL_FILES` build-arg, it will copy everything in the `context` folder. So, only specifying that build-arg would fix the issue, the rest is cleanup to remove duplication and make everything more understandable.

you can try but checking the difference of the filesystem of the images built by these two commands:
```shell
# this will contain everything
docker build . -f edc-controlplane/edc-controlplane-postgresql-azure-vault/build/resources/docker/Dockerfile

# this will contain only what's needed
docker build edc-controlplane/edc-controlplane-postgresql-azure-vault -f edc-controlplane/edc-controlplane-postgresql-azure-vault/build/resources/docker/Dockerfile --build-arg "ADDITIONAL_FILES=build/legal/*"
```

build.gradle.kts
ndr-brt: in any case the buiild-args passed at line 192-194 do specify linux paths. I can switch all of them, not really sure about the value given

paullatzelsperger:
> the root of the problem is that, if you don't specify the `ADDITIONAL_FILES` build-arg, it will copy everything in the `context` folder. 

yep, that is a problem indeed :)

build.gradle.kts
paullatzelsperger: yeah, true.

ndr-brt:
> > the root of the problem is that, if you don't specify the `ADDITIONAL_FILES` build-arg, it will copy everything in the `context` folder.
> 
> yep, that is a problem indeed, but we should never have a docker build _without_ `ADDITIONAL_FILES`, as the legal docs are mandated by EF? Wouldn't this make the problem theoretical in nature?

if you look at the changelog, in the `publish-docker-image` it wasn't, so that line I added in the `build-push-action` job fixed the issue, then I did some refactor to make the `dockerize` task and the `build-push-action` call more similar to each other.

paullatzelsperger (approved):

PR #1475, opened 2024-07-31
feat: add persistence for data-plane-selector

### LINKED ISSUE #1474: data-plane-selector sql persistence
## WHAT

We should enable SQL persistence for data-plane-selector

## WHY
To keep data-plane instances registered after control-plane restart

## HOW

// if possible, outlines a solution proposal

## FURTHER NOTES

ndr-brt:
## WHAT

Add sql persistence for data-plane-selector

## WHY

it was missing

## FURTHER NOTES

- removed duplicated dependency from the version catalog

Closes #1474


edc-extensions/migrations/control-plane-migration/src/main/resources/org/eclipse/tractusx/edc/postgresql/migration/dataplaneinstance/V0_0_1__Init_Dataplaneinstance.sql
```diff
@@ -0,0 +1,34 @@
...
+
+CREATE TABLE IF NOT EXISTS edc_data_plane_instance
+(
+    id                   VARCHAR NOT NULL PRIMARY KEY,
+    data                 JSON,
+    lease_id             VARCHAR
+        CONSTRAINT data_plane_instance_lease_id_fk
+                    REFERENCES edc_lease
+                    ON DELETE SET NULL
+);
```
florianrusch-zf, line 34: Do the schemas of these tables match the old tables, so that in case of an upgrade of the edc (even if it is not supported) there should be no problems?

rafaelmag110:
This is a relevant fix for 0.7.3, as that version already ships the dataplane self-registration extension.

gradle/libs.versions.toml
```diff
@@ -2,8 +2,9 @@
 format.version = "1.1"
 
 [versions]
-edc = "0.8.1"
-fc = "0.8.1"
+edc = "0.8.2-20240801-SNAPSHOT"
+fc = "0.8.2-20240801-SNAPSHOT"
```
paullatzelsperger, line 6: i think we could converge these two again, no reason to have them separate.

paullatzelsperger (approved):

wolf4ood (approved):

PR #1487, opened 2024-08-09
feat: separate store from datasource configurations

ndr-brt:
## WHAT

As [done in upstream](https://github.com/eclipse-edc/Connector/issues/3811), this PRs separate store configuration from the datasource one, this way it will be possibile to define and use different datasources for different stores, and by default they will all be brought back to `default` datasource.

## WHY

## FURTHER NOTES

- moved actual db migration into the `prepare` phase, where it should belong.


wolf4ood (approved):
:axe:

PR #1510, opened 2024-08-22
docs(api): fix published api specs (#1399)

rafaelmag110:
## WHAT

Cherry-pick of that commit to include the correct upstream specs during publishing in SwaggerHub.


rafaelmag110:
@wolf4ood @paullatzelsperger 
We need to cherry pick this into 0.7.4 in order for the specs are published correctly in swaggerhub.

PR #1518, opened 2024-08-27
chore(workflows): Add publish-bugfix-release workflow to main

rafaelmag110:
## WHAT

Brings in the temporary publish-bugfix workflow to main from the bugfix/0.7.4 branch.
This is needed for the workflow to be executed manually from the Actions tab.

### DIFF
--- a/.github/workflows/publish-bugfix-release.yml
+++ b/.github/workflows/publish-bugfix-release.yml
@@ -71,6 +71,8 @@ jobs:
         run: |-
           # replace the project's (default) version, could be overwritten later with the -Pversion=... flag
           sed -i  's/version=.*/version=${{ inputs.version }}/g' gradle.properties
+          # replace the openapi version with the upcoming one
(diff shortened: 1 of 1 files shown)


rafaelmag110:
@paullatzelsperger @wolf4ood 
Can you review this PR please?

paullatzelsperger:
why do we need this bandaid on `main`? There are plans to "properly" refactor the CI workflows anyway, and there is no urgency at the moment.
Version `0.8.0` is not due for some time.

rafaelmag110:
Github will not identify this workflow exists unless it is on the main branch. For us to run what the bugfix branch has, it has to be here.

We can change the contents and have it do the bare mininum aswell. Like printing "this workflow is not ready on main"...

paullatzelsperger:
> Github will not identify this workflow exists unless it is on the main branch. For us to run what the bugfix branch has, it has to be here.
> 
> We can change the contents and have it do the bare mininum aswell. Like printing "this workflow is not ready on main"...

ah, that's right, forgot about that.

paullatzelsperger (approved):

Issue #1519, opened 2024-08-29
Memory Leak During Transfer-PULL Operation


zub4t:
## Describe the bug

In a `transfer-PULL` scenario, a `requestID` is associated with a `DataSource` using a HashMap, but the associated objects are never released.

The `PipelineServiceImpl` class stores an entry in a HashMap named `sources`, linking each `requestID` to its corresponding `DataSource`. However, the function responsible for removing entries from this map is never invoked. As a result, the `sources` variable grows linearly with the number of requests, leading to a memory leak.

### To Reproduce
You can confirm this error by performing a transfer-PULL operation and then inspecting the `sources` member variable in `PipelineServiceImpl`. You'll notice that previous transfer information has not been properly cleaned up.

### Screenshots/Error Messages

Below is the memory usage profile during asset transfers via `PULL` at a rate of 10 Hz.
<img width="944" alt="image" src="https://github.com/user-attachments/assets/50d40570-e5bd-4fd2-997b-67247c3d2bda">

## Context Information
- Used version: 0.7.3

rafaelmag110:
Fixed upstream in PR: https://github.com/eclipse-edc/Connector/pull/4441
Fixed for 0.7.3 in PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/1520

ndr-brt:
fixed in 0.7.4 and also on 0.8.0-rc3

PR #1520, opened 2024-08-29
Fix: PipelineService datasources pull transfers

bmg13:
## WHAT

Similar [to an issue raised in the Upstream](https://github.com/eclipse-edc/Connector/issues/4438), it adds an extension where basically overrides PipelineService implementation to include the new fix.

## WHY

Regarding [an issue found and fixed](https://github.com/eclipse-edc/Connector/issues/4438) in the (most recent version of the) EDC where the DataSources were not closed in PULL transfers, we faced the need of having this fix on a 0.7.X version as well.

## FURTHER NOTES

The transfer() method has Span annotation related with the OpenTelemetry in the Upstream, however was removed here since the addition of dependency just for this case seemed unnecessary (and for any other usage, there is in the upstream).
If need for signalling the OT instrumentation is needed, please raise a comment.

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/1519

### DIFF
--- a/edc-extensions/dataplane/dataplane-self-registration/build.gradle.kts
+++ b/edc-extensions/dataplane/dataplane-self-registration/build.gradle.kts
@@ -21,7 +21,7 @@ dependencies {
     implementation(libs.edc.spi.web)
     implementation(libs.edc.spi.dataplane.selector)
 
-    runtimeOnly(project(":edc-extensions:pipeline-service"))
+    runtimeOnly(project(":edc-extensions:dataplane:pipeline-service"))
 
     testImplementation(libs.edc.junit)
 }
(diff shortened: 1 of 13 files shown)


edc-extensions/dataplane/dataplane-self-registration/build.gradle.kts
```diff
@@ -21,6 +21,8 @@ dependencies {
     implementation(libs.edc.spi.web)
     implementation(libs.edc.spi.dataplane.selector)
 
+    runtimeOnly(project(":edc-extensions:dataplane:pipeline-service"))
+
```
wolf4ood, line 25: I would not put the dependency here, but in the `edc-dataplane-base`

wolf4ood (changes requested):

edc-extensions/dataplane/dataplane-self-registration/build.gradle.kts
bmg13: you're right, thank you, updated in https://github.com/eclipse-tractusx/tractusx-edc/pull/1520/commits/d72b9a2b574f7d97cf2556e4de99c6be8e62b48d!

wolf4ood (approved):

PR #1522, opened 2024-08-30
fix(workflows): Adds prepare release commit push to bugfix branch

rafaelmag110:
## WHAT

Adds prepare release commit push to bugfix branch

## WHY
It isn't pushing to branch and the commit is lost

### DIFF
--- a/.github/workflows/publish-bugfix-release.yml
+++ b/.github/workflows/publish-bugfix-release.yml
@@ -62,8 +62,6 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
       - name: Initialize mandatory git config
         run: |
           git config user.name "eclipse-tractusx-bot"
@@ -94,7 +92,7 @@ jobs:
           git add gradle.properties $(find charts -name Chart.yaml) $(find charts -name README.md)
           git commit --message "Prepare release ${{ inputs.version }}"
           
-          git push origin ${{ github.ref_name }}
+          git push
 
 
   # Release: Maven Artifacts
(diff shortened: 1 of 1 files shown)


rafaelmag110:
Requesting review
@wolf4ood

wolf4ood (approved):

.github/workflows/publish-bugfix-release.yml
```diff
@@ -62,6 +62,8 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
ndr-brt, line 66: this shouldn't be necessary unless there's the need to checkout to an existent commit (like it's done in the `helm-release` job)

.github/workflows/publish-bugfix-release.yml
```diff
@@ -91,6 +93,8 @@ jobs:
         run: |
           git add gradle.properties $(find charts -name Chart.yaml) $(find charts -name README.md)
           git commit --message "Prepare release ${{ inputs.version }}"
+          
+          git push origin ${{ github.ref_name }}
```
ndr-brt, line 97:
```suggestion
          git push
```

rafaelmag110: Wasn't sure just the push would work fine so decided to pass the `ref_name`. Then I wasn't sure the branch would be discoverable and then added the `fetch-depth: 0`.

Nevertheless I tested this in a private repo and the simpler version works, so i'll change.

rafaelmag110:
@ndr-brt Please merge if the new commit its ok.

PR #1498, opened 2024-08-13
Fix: Backport Add Persistence for DataplaneSelector.

bmg13:
## WHAT

This PR backports the added DataPlane instance selector logic to the v0.7.4.

## WHY

To avoid the dataplane registry only in one ControlPlane in memory, this change enables the storage of DataPlane's availability in the bd. So, there should be enabled SQL persistence for data-plane-selector in backported 0.7.4 version.

## FURTHER NOTES

Logic here is partially extracted from https://github.com/eclipse-tractusx/tractusx-edc/pull/1475 PR

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/1497

PR #1529, opened 2024-09-03
fix: upgradeability tests

### LINKED ISSUE #1528: Upgradeability tests broken on main
## Describe the bug

Upgradeability tests are currently failing on `main`

https://github.com/eclipse-tractusx/tractusx-edc/actions/runs/10659260122/job/29596305810?pr=1526

This is due the latest stable version 0.7.4 which contains a data plane migration `V0_0_1__Init_Dataplaneinstance.sql` which is differs from the one in `main` and the flyway checksum fails

wolf4ood:
## WHAT

upgradeability tests

## WHY

upgradeability tests are failing on `main

## FURTHER NOTES

To fix that we have to change the `V0_0_1__Init_Dataplaneinstance.sql` to be equal to the one in 0.7.4. By doing so the migration from `0.8.0-rc2` to the next RC version `0.8.0-rc3` will fail instead.

Closes #1528


ndr-brt (commented):
I'm not sure what happened, so it won't be possible to update 0.7.4 to 0.8.0?

wolf4ood:
Not this is a fix for migration from 0.7.4 to future 0.8.0.

We backported the migration [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1498) for the data plane instance store but the init migration file differs from the one in main, hence the test upgradeability 0.7.4 -> main fails

ndr-brt (approved):

PR #1536, opened 2024-09-06
fix: add Azure Blob Provisioner extension

paullatzelsperger:
## WHAT

This PR backports the upstream `provision-blob` extension that allows for creating blob containers and temporary SAS tokens.
Also adds a End-To-End test for cloud transfers, because up until now, we only had tests that solely used the dataplane.

Note that the S3 Provisioner was **not** backported, because the are some inconsistencies with the endpoint override. For some SDK interactions, the endpoint override can be specified dynamically, for others it is based on a configuration value. That may not be usable in a practical environment. 

Technically, this would have meant:
- significant changes to the AWS SDK client instantiation are necessary (cf [this PR](https://github.com/eclipse-edc/Technology-Aws/pull/436))
- the S3 provisioner itself needs some work to tunnel through the override

## WHY

Allowing the creation of temporary SAS tokens when using Azure Blob Storage as data destination.

## FURTHER NOTES

- Once this PR is merged, it should be cherry-picked into the `bugfix/0.7.5` branch
- The Azure Blob provisioner was _copied_ into the `edc-extensions/backport` directory


paullatzelsperger:
@rafaelmag110 FYI

wolf4ood (approved):

PR #1543, opened 2024-09-11
feat(release workflows): Improve Release workflows to enable bugfix versions

### LINKED ISSUE #1448: Release workflow: make base branch configurable
improvement to the release flow: 
  - configurable base: currently releases are always created off of `main`, but for isolated bugfixes we will need an option to do it off of some other branch
  - duplicated branches: we currently have `releases` and `release/x.y.z`, which seems spurious. We can probably skip the `releases` branch altogether.

rafaelmag110:
## WHAT

Implements the decisions made in the following DR: https://github.com/eclipse-tractusx/tractusx-edc/tree/main/docs/development/decision-records/2024-08-28-Release-Workflow-Refactor

- The PR that triggers the release now has the `main` branch as base instead of `releases`. No need to be moving commits around from releases and then to main.
- A new action `publish-tag-and-release` was created to enable a parametrized execution of the previous GH tag and release set of steps, depending if its a latest or bugfix release.
- Release version is now resolved as the version in `gradle.properties` and not from branch names.
- A manual workflow to trigger manual bugfix releases was introduced. It basically re-uses the release workflow.
- Adds a validation step to skip the run if the tag to be created already exists.
- Adds a gate that prevents unintentional workflow runs in forks.

## WHY

- Enable bugfix versions.
- Refactor existing release workflows to remove the need for a `releases/ ` branch and even `release/*` branches.
- General CI improvements.

Closes #1448


rafaelmag110:
@paullatzelsperger 
@wolf4ood 
Please review.

.github/workflows/draft-release.yaml
```diff
@@ -19,28 +20,64 @@
...
   workflow_dispatch:
     inputs:
       version:
-        description: 'The version you want to release.'
+        description: 'The version you want to release. Ref should be either main for latest releases or a tag for bugfixes. '
         required: true
 
 jobs:
+  validate-release:
+    name: "Validate if version can be released"
```
paullatzelsperger, line 35: nit: I would put a more explicit description like "Validates that tag <TAG> does not already exist"

.github/workflows/draft-release.yaml
```diff
@@ -19,28 +20,64 @@
...
+          fi
+
+  resolve-branch-prefix:
+     name: "Resolve branch prefix (release or bugfix)"
+     needs: validate-release
+     runs-on: ubuntu-latest
+     outputs:
+       branch_name: ${{ steps.release.outputs.branch_name || steps.bugfix.outputs.branch_name }}
+     steps:
+       - id: release
```
paullatzelsperger, line 61: Personally, I prefer a simple shell script over conditional steps or dedicated actions, because shell scripts can be easily replicated in a local environment and are generally more easily understandable.

paullatzelsperger (commented):
minor remarks, looks good to me otherwise

.github/actions/publish-tag-and-release/action.yml
```diff
@@ -0,0 +1,92 @@
...
+name: "Publish Github release and tag"
+description: "Publish Github release and tag"
+inputs:
+  version:
+    required: true
+    description: "The version to be used in the tag and release publication"
+  token:
+    required: true
+    description: "Github token"
+  isLatest:
```
ndr-brt, line 32: nit: in other composite actions we used different case. the most used looks to be the snake_case. 
(for the record, in yaml files for some reasons I like kebab-case more, but snake will fit better in this project)

.github/actions/publish-tag-and-release/action.yml
```diff
@@ -0,0 +1,92 @@
...
+    - name: Create Release Tag
+      id: create_release_tag
+      shell: bash
+      run: |
+        # informative
+        git branch -a
+        git tag
+
+        # Create & push tag
+        git tag --force ${{ inputs.version }}
```
ndr-brt, line 55: using `--force` does not look really safe

.github/actions/publish-tag-and-release/action.yml
```diff
@@ -0,0 +1,92 @@
...
+        # Compute new snapshot version, do not increment snapshot on non-final releases, e.g. -rc1
+        if [ -z $SNAPSHOT ]; then
+          # snapshot
+          echo "${{ inputs.version }} is a final release version, increase patch for next snapshot"
+          INC=1
+        else
+          echo "${{ inputs.version }} is not a final release version (contains \"$SNAPSHOT\"), will not increase patch"
+        fi
+
+        VERSION="$RELEASE_VERSION_MAJOR.$RELEASE_VERSION_MINOR.$((RELEASE_VERSION_PATCH+$INC))-SNAPSHOT"
```
ndr-brt, line 82: I'd say that new snapshot on main branch should be:
`"$RELEASE_VERSION_MAJOR.$((RELEASE_VERSION_MINOR+$INC)).0"-SNAPSHOT`

.github/workflows/release.yml
```diff
@@ -0,0 +1,154 @@
...
+
+  workflow_call:
+
+
+jobs:
+  # Gate
+  check-head:
+    name: "Check if repository is not fork AND head is release OR base is bugfix"
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'eclipse-tractusx/tractusx-edc' && (startsWith(github.ref_name, 'bugfix/') || startsWith(github.event.pull_request.head.ref, 'release/')) }}
```
ndr-brt, line 40: this looks complicated plus it will be evaluated for every PR closed. given that we're following a sort-of gitflow approach, why don't we release on the `release` branch and at the end we merge it back to `main` later? this will simplify things a little.

.github/actions/publish-tag-and-release/action.yml
rafaelmag110: Agree. I overlooked this and figured it was needed at some point. Maybe to override existing tags? I don't think it's a good practice anyway.

.github/workflows/release.yml
rafaelmag110: Yes, it will be evaluated for every merged PR on `main` but it won't run because of the condition. And since this is a job-level conditional, github will just skip its execution and not throw any error (btw, that is the reason why I prefer job/action level conditionals vs everything script).

Like such.

image


If we release from the `release` branch we can't create the tag from it, otherwise the release branch cannot be deleted.

I wouldn't say it looks complicated. Its actually very readable. It's just a veeeery long line 😅

ndr-brt: well, the tag can be created after the release branch has been merged into main. I think that's the typical gitflow approach

.github/actions/publish-tag-and-release/action.yml
ndr-brt: I would just remove that, the only advantage is that it speeds-up the process in case of re-launch (because of a failure) but it exposes the release to affect older releases.

.github/workflows/release.yml
rafaelmag110: I don't think I got your point then, sorry.
Your suggestion is to change the trigger of the release workflow to be the push on the `release` branch?

ndr-brt: Either having this workflow to run when a PR is closed on a `release` branch or making it manual as for the bugfix version. 
At the end of the release we could merge back to main if ref_name starts with `release` and snapshot version can be bumped.
The more the `release` and `bugfix` process are similar, the easier will be to maintain them

rafaelmag110: The workflow is being triggered when a PR is closed on a `release` branch. If we really want to avoid the workflow to be evalutated on every closed PR to main, we can make it manual only.

.github/actions/publish-tag-and-release/action.yml
rafaelmag110: [3ea619b](https://github.com/eclipse-tractusx/tractusx-edc/pull/1543/commits/3ea619b7c4e94c36c9efbb977336e6dbd521b639)

.github/actions/publish-tag-and-release/action.yml
rafaelmag110: [3ea619b](https://github.com/eclipse-tractusx/tractusx-edc/pull/1543/commits/3ea619b7c4e94c36c9efbb977336e6dbd521b639)

.github/actions/publish-tag-and-release/action.yml
rafaelmag110: [3ea619b](https://github.com/eclipse-tractusx/tractusx-edc/pull/1543/commits/3ea619b7c4e94c36c9efbb977336e6dbd521b639)

.github/workflows/draft-release.yaml
rafaelmag110: I refactored the validation jobs in [788baa1](https://github.com/eclipse-tractusx/tractusx-edc/pull/1543/commits/788baa13d5600130017551642181721f394b7682) since I also agree those could be simpler.

On a personal level, I prefer the conditionals on the jobs or actions. The workflow execution graph is calculated before it is sent to the runner and false conditionals are skipped. When having everything in a shell script you have to be more careful what behaviors should lead to job failure and explicitly have it fail.

I'm fine either way. Let me know if the refactor is ok.

paullatzelsperger: We've seen this numerous times, where we want to replicate the runner's behaviour and debug or test a workflow locally, e.g. after a refactoring or a maintenance change. If everything is a job with conditional execution, there is zero chance of doing that, and ACT only gets you so far.

Let's get this merged as-is so as not to block the PR. 

Generally speaking, I would be very cautious when adding GH actions, because we are effectively introducing undebuggable foreign code, that you have no control over.

paullatzelsperger (approved):

PR #1555, opened 2024-09-16
docs: Proposal for FederatedCatalog Distribution and TargetNodeDirectory

bmg13:
## WHAT

Decision Record for new proposal for FederatedCatalog distribution (with Tractus-X) and the TargetNodeDirectory (TND).

The TND initial proposal was defined in [this](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556) PR and continue in the current one.

Related with https://github.com/eclipse-tractusx/tractusx-edc/issues/1585

[edit @paullatzelsperger]
Closes https://github.com/eclipse-tractusx/sig-release/issues/736

PR #1556, opened 2024-09-16
docs: Proposal for a TargetNodeDirectory in Tractus-X.

bmg13:
## WHAT

Decision Record for new proposal for FederatedCatalog TargetNodeDirectory with Tractus-X.

Related with https://github.com/eclipse-tractusx/tractusx-edc/issues/1585

### DIFF
--- a/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
+++ b/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
@@ -2,65 +2,29 @@
 
 ## Decision
 
-A new service that will have a db/cache containing the BPNL's and Connectors' URL's of each partner a member wants the offers from. Simply, **this new service works as the TargetNodeDirectory for the partner's FederatedCatalog**.
+A new service that will contain Connectors' URL's of each partner a member wants the offers from, acting as the **TargetNodeDirectory for the partner's FederatedCatalog**.
 
 
 ## Rationale
 
-While considering new interventions in the Federated Catalog, this discussion aims to propose a solution for the TargetNodeDirectory.
+While considering new interventions in the Federated Catalog, this decision aims to set the TargetNodeDirectory.
 From the [documentation](https://eclipse-edc.github.io/docs/#/submodule/FederatedCatalog/docs/developer/architecture/federated-catalog.architecture)
 > The Federated Catalog requires a list of Target Catalog Nodes, so it knows which endpoints to crawl. This list is provided by the TargetNodeDirectory. During the preparation of a crawl run, the ExecutionManager queries that directory and obtains the list of TCNs.
 
-So having a service containing the data of desired BPNL's and respective Connectors' URL's that a certain partner wants, but here the partner itself has the service in their scope and users are able to input (in the API of new service) the Participants.
+So having a service containing the data of Connectors' URL's that a certain partner wants and having it in their scope. Users are able to input (in the API of new service) the Connectors' URL's.
 
-This solution allows the member to choose the relevant data that interests them (offers from specific participants) enabling the reduction of storage cost and less network calls. Additionally, each member has control on how to host and manage this new service, service changes do not affect other parties (unless contract changes) and is able to be independently scalable.
+This solution allows the member to choose precisely the Target Catalog Nodes that interests them, resulting in reduced storage cost and network calls. Additionally, each member has control on how to host and manage this new service. Service changes do not affect other parties (unless contract changes) and is able to be independently scalable.
 
-
-Other solutions were also considered
+Other solution was also considered
 
 - File in a S3 bucket (or different cloud provider's solution)
-    - This solution was discarded due to One file for all instead of each partner having the data that respectively needs does not match the requirement and this solution would lock the usage of a proprietary tool (cloud provider) being harder to sustain in an open source community.
-
-- Centralized service with all participants data
-    - Solution was not considered due to presenting a single point of failure, the partner still needs to have a way to dynamically store on desired partners ids' and each partner must have the BPNL or the Connectors' URL's beforehand.
+    - This solution was discarded due to one file for all instead of each partner having the data that respectively needs does not match the requirement and this solution would lock the usage of a proprietary tool (cloud provider) being harder to sustain in the long run.
 
 ## Approach
 
-Using the Discovery Service, the FC is able to request the connetors' url's that it requires, providing the BPNL's. Since the Discovery Service has already built in this feature, the work of this flow is already relatively well established.
-However, the main question here arises: what if a new partner joins the space? The existing partners that want to obtain the offers of the new partner do not know of its existence. So an additional mechanism is needed. The default implementation is a static file, however a dynamic approach would be preferable (no downtime expected).
+The user is able to obtain the Connectors' URL's (through the Discovery Service, as an example) and store them in the new service through its API, allowing the Federated Catalog to request them. The API will allow to save a list of Connectors' URL's in bulk. These can later be retrieved and crawled by the Federated Catalog.
 
-Even with a new service registering the Discovery Service (basically making it available in the space to other partners) the existing partners that may want to retrieve the new partners' offers must establish the need to retrieve them by defining the new BPNL somewhere.
-
-
-A service containing the data of desired BPNL's and respective Connectors' URL's that a certain partner wants, but here the partner itself has the service in their scope and users are able to input (in the API of new service) the Participants. Simply, the new service would have a db/cache containing the BPNL's and Connectors' URL's of each partner this member wants the offers from. Again, **this new service works as the TargetNodeDirectory for the partner's FederatedCatalog**.
-
-
-Expected flow
-
-- Using the new service API, partner adds the participants that wants to obtain the offers from (BPNL's);
-- New service stores this data (db or cache);
-- Providing the BPNL, the FC obtains the Connectors' URL's from the DS API;
-- Using the new service API, FC passes the info from the DS to the new service;
-- New service stores this data (db or cache);
-- FC, with the ExecutionManager, periodically crawls the participants catalogs based on which ones are stored in the new service;
-- FC catalog cache is updated;
-- If new participant joins the space, user adds that participant's BPNL through new service API (like first step)
-
-![tnd_proposal](https://github.com/user-attachments/assets/5245f8af-845d-4707-8871-359457d581e3)
-
-
-⚠️  Any case that the Discovery Service (DS) is used, **a technical user account to access it must be requested**. Without it, the usage of the DS API is limited since a token is used in all relevant requests.
-
-Technically, the new service **would have a simple API** where the user could
-- Add a new BPNL
-- Remove a BPNL
-- List BPNL's (value and connectors associated with it)
-
-
-Regarding the service's Persistence Store, any suggestion is welcomed. Having a DB or a local Cache can suffice, but other options like persisting in a file could also be considered. Or a mixed between this options could be the best solution.
-
-ℹ️  Each partner must have the BPNL or the Connectors' URL's beforehand. If a new Partner registers and an existing partner would want their catalog, the BPNL or Connectors' URL's of the new partner must be obtain first.
 
 Finally, considering service deployment, a new chart can be created just for this new service (similar to the existing ones), being its usage only decided by the member. As so, a Dockerfile should exist to ease this approach (giving the user option of running it in a container or run a simple `jar`).
 
-The clear limitation of this approach is the overhead of a new service, specially one with a persistence store. This will increase costs in favour of faster responses.
+Limitations of this solution are that each partner must have the Connectors' URL's beforehand and deal with the overhead of a new service, specially one with a persistence store.
--- a/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
+++ b/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
@@ -23,7 +23,7 @@ Other solution was also considered
 ## Approach
 
 The user is able to obtain the Connectors' URL's (through the Discovery Service, as an example) and store them in the new service through its API. The API will allow to save a list of Connectors' URL's in bulk and the service is responsible to store that (in memory or in database). These can later be retrieved and crawled by the Federated Catalog.
-
+This solution improves on the default one of having the data in a static file since a dynamic approach would avoid downtime when a change is required.
 
 Finally, considering service deployment, a new chart can be created just for this new service (similar to the existing ones), being its usage only decided by the member. As so, a Dockerfile should exist to ease this approach (giving the user option of running it in a container or run a simple `jar`).
 
--- a/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
+++ b/docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
@@ -2,7 +2,7 @@
 
 ## Decision
 
-A new service that will contain Connectors' URL's of each partner a member wants the offers from, acting as the **TargetNodeDirectory for the partner's FederatedCatalog**.
+A new extension in the FederatedCatalog that will have a db/cache containing the BPNL's and Connectors' URL's of each partner a member wants the offers from, acting as the TargetNodeDirectory for the partner's FederatedCatalog.
 
 
 ## Rationale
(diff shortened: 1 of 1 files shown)


florianrusch:
Doesn't such a solution already exists with the BPN/EDC-Discovery? How would your service be different?

paullatzelsperger:
> Doesn't such a solution already exists with the BPN/EDC-Discovery? How would your service be different?

The `TargetNodeDirectory` is an internal service to the FederatedCatalog, that would source its information from the BPN/EDC-Discovery service.

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+While considering new interventions in the Federated Catalog, this discussion aims to propose a solution for the TargetNodeDirectory.
+From the [documentation](https://eclipse-edc.github.io/docs/#/submodule/FederatedCatalog/docs/developer/architecture/federated-catalog.architecture)
+> The Federated Catalog requires a list of Target Catalog Nodes, so it knows which endpoints to crawl. This list is provided by the TargetNodeDirectory. During the preparation of a crawl run, the ExecutionManager queries that directory and obtains the list of TCNs.
+
+So having a service containing the data of desired BPNL's and respective Connectors' URL's that a certain partner wants, but here the partner itself has the service in their scope and users are able to input (in the API of new service) the Participants.
+
+This solution allows the member to choose the relevant data that interests them (offers from specific participants) enabling the reduction of storage cost and less network calls. Additionally, each member has control on how to host and manage this new service, service changes do not affect other parties (unless contract changes) and is able to be independently scalable.
```
rafaelmag110, line 16:
```suggestion
This solution allows the member to choose precisely the Target Catalog Nodes that interests them reducing storage cost and network calls. Additionally, each member has control on how to host and manage this new service, service changes do not affect other parties (unless contract changes) and is able to be independently scalable.
```

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,66 @@
...
+However, the main question here arises: what if a new partner joins the space? The existing partners that want to obtain the offers of the new partner do not know of its existence. So an additional mechanism is needed. The default implementation is a static file, however a dynamic approach would be preferable (no downtime expected).
+
+Even with a new service registering the Discovery Service (basically making it available in the space to other partners) the existing partners that may want to retrieve the new partners' offers must establish the need to retrieve them by defining the new BPNL somewhere.
+
+
+A service containing the data of desired BPNL's and respective Connectors' URL's that a certain partner wants, but here the partner itself has the service in their scope and users are able to input (in the API of new service) the Participants. Simply, the new service would have a db/cache containing the BPNL's and Connectors' URL's of each partner this member wants the offers from. Again, **this new service works as the TargetNodeDirectory for the partner's FederatedCatalog**.
```
rafaelmag110, line 35: If you feel the need to put something in simpler terms, why do you keep both forms and not only the simpler one?

rafaelmag110 (changes requested):
Overall this decision record seems to be much more complicated than it actually needs to be. As I said in one of the comments, if you feel the need to use simpler terms, then avoid having repeated text. This is a straight forward decision to explain: 
-The FC crawls Target Catalog Nodes, which in essence are connector URLs.
-These nodes are resolved by the FC ExectionManager from the TargetNodeDirectory
-The current directory implementation is not ideal because ...
-The decision is to implement it the following way ...

rafaelmag110:
IMO I'd avoid making the `TargetNodeDirectory` extension dependent on the tractusx discovery service. It might come to a point where the association decides the DS is not longer needed and we would need to revisit this again.

A `TargetCatalogNode` is a connector URL, and representing it as a mapping between BPNL - Connector URL makes it unnecessarily complex.

We can allow the user of this new service API to input various `TargetCatalogNodes` (as in, a list of connector urls), in a single `POST` but how those URLs are resolved beforehand isn't the TargetNodeDirectory business.

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: Yes, nice touch, added [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/fae5e8bafbc033e16bcfce259fc95cb2ad7b06af).

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: Yeah. Removed duplicated content in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/fae5e8bafbc033e16bcfce259fc95cb2ad7b06af).

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+⚠️  Any case that the Discovery Service (DS) is used, **a technical user account to access it must be requested**. Without it, the usage of the DS API is limited since a token is used in all relevant requests.
+
+Technically, the new service **would have a simple API** where the user could
+- Add a new BPNL
+- Remove a BPNL
+- List BPNL's (value and connectors associated with it)
+
+
+Regarding the service's Persistence Store, any suggestion is welcomed. Having a DB or a local Cache can suffice, but other options like persisting in a file could also be considered. Or a mixed between this options could be the best solution.
```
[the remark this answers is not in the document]
bmg13: Removed the suggestion and focused only on proposed solution. Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/fae5e8bafbc033e16bcfce259fc95cb2ad7b06af).

bmg13:
> IMO I'd avoid making the `TargetNodeDirectory` extension dependent on the tractusx discovery service. It might come to a point where the association decides the DS is not longer needed and we would need to revisit this again.
> 
> A `TargetCatalogNode` is a connector URL, and representing it as a mapping between BPNL - Connector URL makes it unnecessarily complex.
> 
> We can allow the user of this new service API to input various `TargetCatalogNodes` (as in, a list of connector urls), in a single `POST` but how those URLs are resolved beforehand isn't the TargetNodeDirectory business.

Yeah, I was aiming at abstracting this action from the user and trying to allow the service to do it on its own, however, after reading that comment, I have to agree with you. Removing an external dependency (with some overhead even) from this service and allowing it to work only as a TargetNodeDirectory while giving the user that control is the best approach, specially in the long term. Changed the doc to reflect it.

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Decision
+
+A new service that will contain Connectors' URL's of each partner a member wants the offers from, acting as the **TargetNodeDirectory for the partner's FederatedCatalog**.
+
+
+## Rationale
+
+While considering new interventions in the Federated Catalog, this decision aims to set the TargetNodeDirectory.
+From the [documentation](https://eclipse-edc.github.io/docs/#/submodule/FederatedCatalog/docs/developer/architecture/federated-catalog.architecture)
```
ndr-brt, line 11: this link is not working, maybe it's better to point directly to the readme in the repo:
https://github.com/eclipse-edc/FederatedCatalog/blob/e733355c6991ff633ee009bd5f35ced61e941da6/docs/developer/architecture/federated-catalog.architecture.md

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,30 @@
...
+> The Federated Catalog requires a list of Target Catalog Nodes, so it knows which endpoints to crawl. This list is provided by the TargetNodeDirectory. During the preparation of a crawl run, the ExecutionManager queries that directory and obtains the list of TCNs.
+
+So having a service containing the data of Connectors' URL's that a certain partner wants and allow them to host it. Users are able to input (using the API of new service) the Connectors' URL's of the connectors they want the catalogs from.
```
ndr-brt, line 14: what do we mean here with "service"? a class? a runtime? an endpoint? 
could we give it a name as well?

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+This solution allows the member to choose precisely the Target Catalog Nodes that interests them, resulting in reduced network calls and latency. Additionally, each member has control on how to host and manage this new service. Service changes do not affect other parties (unless contract changes) and is able to be independently scalable.
```
ndr-brt, line 16:
```suggestion
This solution allows the member to choose precisely the Target Catalog Nodes that interests them, resulting in reduced network calls and latency. Additionally, each member has control on how to host and manage this new service. Service changes do not affect other parties (unless contract changes) and it can be scaled be independently.
```

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Other solution was also considered
+
+- File in a S3 bucket (or different cloud provider's solution)
+    - This solution was discarded due to one file for all instead of each partner having the data that respectively needs does not match the requirement and this solution would lock the usage of a proprietary tool (cloud provider) being harder to sustain in the long run.
+
+## Approach
+
+The user is able to obtain the Connectors' URL's (through the Discovery Service, as an example) and store them in the new service through its API. The API will allow to save a list of Connectors' URL's in bulk and the service is responsible to store that (in memory or in database). These can later be retrieved and crawled by the Federated Catalog.
```
ndr-brt, line 25: is this operation meant to be done manually?

ndr-brt (commented):
What are the reasons to deploy this service separatedly? It could be deployed together with the `FederatedCatalog`.

I also think it could implement a process to keep the directory in sync with the `Discovery Service`, and to provide an API to manage a "BPN Whitelist", that will permit to "enable" only specific BPNs.

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: Yes, with the goal of the removing the Discovery Service dependency from this TargetNodeDirectoryService the responsibility of obtaining and saving the the Connectors' URL's lies with the user, as sugggested [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#issuecomment-2376486527).

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: unsure why, but when created this doc that link pointed to correct documentation (that was the same as the readme.md file) and now it does not.
changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/a9cdd06d866b53b2ad56fbd0c92f0a67af751ae2).

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: the idea would be to have a runnable jar. a service able to run independently.
following a suggestion, maybe TargetNodeDirectoryService. 
Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/a9cdd06d866b53b2ad56fbd0c92f0a67af751ae2).

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
bmg13: Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556/commits/a9cdd06d866b53b2ad56fbd0c92f0a67af751ae2).

bmg13:
> What are the reasons to deploy this service separatedly? It could be deployed together with the `FederatedCatalog`.

The idea is to have this TargetNodeDirectoryService as a valid solution to any use, however if some user intends to define other TND they should not be locked at this Service, I believe.


> I also think it could implement a process to keep the directory in sync with the `Discovery Service`, and to provide an API to manage a "BPN Whitelist", that will permit to "enable" only specific BPNs.

Initially, something similar was proposed, the user inputs the BPNs from which they want the catalogs from (working as a whitelist) and the TargetNodeDirectoryService would collect the respective Connectors' URL's from the Discovery Service and store them. However, and rightly pointed out by @rafaelmag110 in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#issuecomment-2376486527), there is a way to ensure the user chooses the Connectors' URL's they want, whitelisting them, beforehand (manually by using the Discovery Service, per example) and not create that explicit dependency on TargetNodeDirectoryService.

docs/development/decision-records/2024-09-16-tx-target-node-directory-proposal/README.md
ndr-brt: that's because upstream we just switched to the new docs site (https://eclipse-edc.github.io), but FC documentation still needs to be moved

ndr-brt:
> > What are the reasons to deploy this service separatedly? It could be deployed together with the `FederatedCatalog`.
> 
> The idea is to have this TargetNodeDirectoryService as a valid solution to any use, however if some user intends to define other TND they should not be locked at this Service, I believe.
> 
> > I also think it could implement a process to keep the directory in sync with the `Discovery Service`, and to provide an API to manage a "BPN Whitelist", that will permit to "enable" only specific BPNs.
> 
> Initially, something similar was proposed, the user inputs the BPNs from which they want the catalogs from (working as a whitelist) and the TargetNodeDirectoryService would collect the respective Connectors' URL's from the Discovery Service and store them. However, and rightly pointed out by @rafaelmag110 in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#issuecomment-2376486527), there is a way to ensure the user chooses the Connectors' URL's they want, whitelisting them, beforehand (manually by using the Discovery Service, per example) and not create that explicit dependency on TargetNodeDirectoryService.

To me looks overkill to have another separated service just for the TargetNodeDirectory, I didn't see in this document a valid reason to do so.

I also don't see why it shouldn't interact with the Discovery Service, I mean, everything in software can change, but that's what we do, we adapt to update things to make them work altogether :) (plus I don't think that it will disappear from one day to the other).
One downside that I see by having this process as manual is that, for example, if a participant changes the URL, there will be the need of a manual intervention to update it.
BPNs instead are more stable data (it's, in fact, the current Catena-X participant identifier), shorter and easier to read than an URL.

Machines are made for automation, manual work is boring and prone to error, let's just avoid it.

ndr-brt (commented):
> > What are the reasons to deploy this service separatedly? It could be deployed together with the `FederatedCatalog`.
> 
> The idea is to have this TargetNodeDirectoryService as a valid solution to any use, however if some user intends to define other TND they should not be locked at this Service, I believe.
> 
> > I also think it could implement a process to keep the directory in sync with the `Discovery Service`, and to provide an API to manage a "BPN Whitelist", that will permit to "enable" only specific BPNs.
> 
> Initially, something similar was proposed, the user inputs the BPNs from which they want the catalogs from (working as a whitelist) and the TargetNodeDirectoryService would collect the respective Connectors' URL's from the Discovery Service and store them. However, and rightly pointed out by @rafaelmag110 in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#issuecomment-2376486527), there is a way to ensure the user chooses the Connectors' URL's they want, whitelisting them, beforehand (manually by using the Discovery Service, per example) and not create that explicit dependency on TargetNodeDirectoryService.

To me looks overkill to have another separated service just for the TargetNodeDirectory, I didn't see in this document a valid reason to do so.

I also don't see why it shouldn't interact with the Discovery Service, I mean, everything in software can change, but that's what we do, we adapt to update things to make them work altogether :) (plus I don't think that it would eventually disappear from one day to another).
One downside that I see by having this process as manual is that, for example, if a participant changes the URL, there will be the need of a manual intervention to update it.
BPNs instead are more stable data (it's, in fact, the current Catena-X participant identifier), shorter and easier to read than an URL.

Machines are made for automation, manual work is boring and prone to error, let's just avoid it.


EDIT: I posted this as a comment but I deleted it and re-proposed it as review :)

bmg13:
> I was suggesting to keep it embedded in the federated catalog, because I don't see why it should deployed in a separated runtime.

Okay, this make sense, will rewrite to include TND embedded in the federated catalog.


> I also don't see why it shouldn't interact with the Discovery Service, I mean, everything in software can change, but that's what we do, we adapt to update things to make them work altogether :) (plus I don't think that it would eventually disappear from one day to another).
One downside that I see by having this process as manual is that, for example, if a participant changes the URL, there will be the need of a manual intervention to update it.
BPNs instead are more stable data (it's, in fact, the current Catena-X participant identifier), shorter and easier to read than an URL.

I see. So we will suggest having a `DiscoveryServiceRetriever` (DSR) extension included in the Federated Catalog. This will allow to:
- Have and API for a member to add/delete the BPNL's (or connector URL's) of participants from which catalogs are wanted;
- Create a bearer token (after requesting from identity provider) to retrieve the connector url's associated with stored BPNL's;
- Store the data retrieved.

If by any chance a member does not aim at use the Discovery Service, a new extension can be create and replace the DSR.

This will allow to use the SQL (or InMem) store already present in the Federated Catalog.

ndr-brt (commented):
at this point my suggestion would be to rephrase this and #1555 in a single DR, because they target the same scope, it will be easier to follow for the community.

bmg13:
> at this point my suggestion would be to rephrase this and #1555 in a single DR, because they target the same scope, it will be easier to follow for the community.

I followed the idea of it being two separated decisions (despite same scope), but will try to merge them in the meantime.

bmg13:
Considering the [suggestion](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#pullrequestreview-2336701529), this PR is closed and existing proposal added to this [PR](https://github.com/eclipse-tractusx/tractusx-edc/pull/1555).

bmg13:
Considering the [suggestion](https://github.com/eclipse-tractusx/tractusx-edc/pull/1556#pullrequestreview-2336701529), this PR is closed and existing proposal added to this [PR](https://github.com/eclipse-tractusx/tractusx-edc/pull/1555).

PR #1558, opened 2024-09-17
feat(snapshot workflow): Allow snapshots from bugfix branches

### LINKED ISSUE #1557: Publish New Snapshot should work for bugfix branches
## WHAT

When running the publish-new-snapshot workflow manually from a bugfix branch, all published artifacts will have the same version has defined in the `gradle.properties` (usually should be the bugfix version). This leads to an unexpected release of the bugfix and not of a "-SNAPSHOT" version. 

## WHY

We should be able to publish snapshot artifacts for a bugfix, to allow pre-validation.

## HOW

Allow for a `version` input on the workflow.

## FURTHER NOTES

The publish-new-snapshot workflow also has a few quirks that could be adjusted.
- It references the `releases` branch which was removed in #1543 
- It references `hotfix` branches but nomenclature has changed in #1543 
- It runs when a new released is published. Can this be removed?

rafaelmag110:
## WHAT

This PR proposes a refactor to the `determine-version` job of the `publish-new-snapshot` workflow, that improves the job to resolve the correct snapshot version string from the gradle.properties every time. This version string is then used as input for the various publishing workflows.

As the publish-new-snapshot workflow can be manually activated, this improvement also allows for the creation of a snapshot from any branch.

## WHY

To allow snapshots from bugfix branches.

## FURTHER NOTES

In light of #1543, also took the opportunity to remove some unnecessary "listeners" from the run-all-tests workflow, making the file simpler.
- Removed any reference to the releases branch as its not needed anymore.
- Removed any reference to previews/* branches as they are not needed anymore.
- Checking if the `github.event_name == pull_request` is redundant, since that event can't trigger the workflow.
- We don't need to run all tests on tag push or release publication, since the tests were already executed a) on PR and b) on the branch from where the tag/release will be pushed/published.

Closes #1557

### DIFF
--- a/.github/workflows/publish-new-snapshot.yaml
+++ b/.github/workflows/publish-new-snapshot.yaml
@@ -23,20 +23,17 @@
 
 ---
 name: "Publish new snapshot"
+run_name: "Publish new snapshot from ${{github.ref_name}}"
 
 on:
   workflow_run:
     workflows: [ "Run-All-Tests" ]
     branches:
       - main
-      - releases
       - release/*
-      - hotfix/*
+      - bugfix/*
     types:
       - completed
-  release:
-    types:
-      - published
   workflow_dispatch:
 
   # periodic build triggers are defined in run-all-tests.yml, which triggers this workflow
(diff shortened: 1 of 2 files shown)


rafaelmag110:
@paullatzelsperger @wolf4ood @ndr-brt

paullatzelsperger (approved):

wolf4ood (approved):

ndr-brt (approved):

Issue #1577, opened 2024-09-23
`0.7.5` Duplicate Transfer-Data-Plane-Signaling Dependencies in `edc-controlplane-base` Module?


wolf4ood:
### Discussed in https://github.com/eclipse-tractusx/tractusx-edc/discussions/1576

<div type='discussions-op-text'>

<sup>Originally posted by **efiege** September 23, 2024</sup>
During our review of the dependency tree for the `edc-controlplane-postgresql-azure-vault` module, we observed that both the following dependencies are being included:

- org.eclipse.edc:transfer-data-plane-signaling:0.7.2
- project :edc-extensions:transfer-dataplane-signaling

It appears that the version included in this repository is a patched version of `org.eclipse.edc:transfer-data-plane-signaling:0.7.2`, addressing this issue: https://github.com/eclipse-tractusx/tractusx-edc/pull/1534.

However, having both dependencies included might result in the registration of two `DataPlaneSignalingFlowController` instances within the `DataFlowManager`, one with the fix and one without. This could potentially cause the EDC to load the unpatched version, depending on how it resolves these extensions.

The unpatched core EDC version is referenced here:

https://github.com/eclipse-tractusx/tractusx-edc/blob/75d1526115bf3922c204aea4ff8434a1b37eb514/gradle/libs.versions.toml#L233

https://github.com/eclipse-tractusx/tractusx-edc/blob/75d1526115bf3922c204aea4ff8434a1b37eb514/edc-controlplane/edc-controlplane-base/build.gradle.kts#L70
</div>

wolf4ood:
@bmg13 could you check on this ?

Thanks

bmg13:
hey @wolf4ood good point, please assign it to me, will try to address it in the meantime

rafaelmag110:
@wolf4ood this was merged. Can it be closed?

PR #1578, opened 2024-09-23
Fix: Remove duplicate transfer data plane signaling dependencies

bmg13:
## WHAT

This PR removes duplicate transfer data plane signaling dependencies.

## WHY

Avoid potentially having two instances of DataPlaneSignalingFlowController.

## FURTHER NOTES

Confirmed in new dependencies check.

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/1577

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -167,3 +167,8 @@ jobs:
         run: |
           ./gradlew compileJava compileTestJava
           ./gradlew -p edc-tests/edc-dataplane test -DincludeTags="CloudTransferTest"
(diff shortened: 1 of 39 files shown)


ndr-brt (commented):
0.7.5 has already been released, shouldn't this be done on a `bugfix/0.7.6` branch?

bmg13:
> 0.7.5 has already been released, shouldn't this be done on a `bugfix/0.7.6` branch?

yes, it should, saw that branch was now created. Will create the pr in a sec

PR #1584, opened 2024-09-26
fix: return 404 when a dataset doesn't have attached policies

ndr-brt:
## WHAT

Backport upstream fix https://github.com/eclipse-edc/Connector/pull/4490
to 0.7.6, by using a response filter instead of copy-paste all the module.

## WHY

## FURTHER NOTES


wolf4ood (approved):

atipijasin:
Amazing! Thx🚀

PR #1601, opened 2024-10-03
feature: Agreement Retirement

rafaelmag110:
## WHAT

Introduces a feature that enables a participant to retire an active contract agreement.

## WHY

There are cases where the cancellation or update of a contract agreement might be valid, i.e. as CatenaX use cases evolve the respective frame contracts also do, and legally speaking the old contracts are no longer valid after the new version is released. There should be a way to technically retire the old agreements, effectively terminating any ongoing transfer process and disabling any new transfer request on that agreement.

## FURTHER NOTES

Only in memory storage is provided for the AgreementsRetirementStore. An SQL based implementation will be provided in a following PR.

This feature only works from a provider perspective since no "agreement retirement message" is sent to the consumer. This only enables the technical retirement of an agreement after the official retirement is legally agreed upon by both parties.

Closes https://github.com/eclipse-tractusx/sig-release/issues/777


jimmarino (changes requested):
Please do not submit large features like this without first proposing a Decision Record and having it go through the review process to be discussed.

rafaelmag110:
Closed.
DR created: https://github.com/eclipse-tractusx/tractusx-edc/pull/1604

Issue #1611, opened 2024-10-07
deprecate `tractusx-connector-azure-vault`


ndr-brt:
## WHAT

We will deprecate first and remove later the `tractusx-connector-azure-vault` distribution

## WHY
We should focus on a single distribution that is interacting with open source services.

## HOW

DR will come

## FURTHER NOTES

// anything else you want to outline

_Please be sure to take a look at
our [contribution guidelines](https://github.com/eclipse-tractusx/tractusx-edc/blob/main/CONTRIBUTING.md) and
our [PR etiquette](https://github.com/eclipse-tractusx/tractusx-edc/blob/main/pr_etiquette.md)._

PR #1612, opened 2024-10-07
docs: DR for azure distrubution deprecation

ndr-brt:
## WHAT

Adds DR to document azure distribution deprecation decision.

## WHY

## FURTHER NOTES

Part of #1611

### DIFF
--- a/docs/development/decision-records/2024-10-07-azure-distribution/README.md
+++ b/docs/development/decision-records/2024-10-07-azure-distribution/README.md
@@ -1,0 +1,21 @@
+# Refactor of the Release Workflow
+
+## Decision
+
+We will stop to provide an azure flavoured distribution of the connector.
+
+## Rationale
+
+Two reasons are pushing toward this decision:
+- Tractus-X Connector is a reference implementation for Catena-X: so nothing should stop anybody to build and use their 
+  own Connector (also without using EDC), so there are no reasons for us to maintain more than one distribution.
+- Tractus-X is an open source software and we should focus on a distribution that involves OSS services.
+
+## Approach
+
+We will deprecate, 0.8.0:
+- `tractusx-connector-azure-vault` helm chart
+- `edc-controlplane-postgresql-azure-vault` docker image
+- `edc-dataplane-postgresql-azure-vault` docker image
+
+We will stop publishing them from version 0.9.0 on.


docs/development/decision-records/2024-10-07-azure-distribution/README.md
```diff
@@ -0,0 +1,21 @@
+# Refactor of the Release Workflow
+
+## Decision
+
+We will stop to provide an azure flavoured distribution of the connector.
+
+## Rationale
+
+Two reasons are pushing toward this decision:
+- Tractus-X Connector is a reference implementation for Catena-X: so nothing should stop anybody to build and use their
```
rafaelmag110, line 10: I feel like the point of this first item should not be towards the freedom dataspace participants have to create their own distribution, because this is true since the beginning of the project. What is different at this point, IMO, is how easy it is to create and release a new distribution using the existing one as reference.

ndr-brt: This point was meant to tell the community that we don't "have to" maintain such distribution because our goal is not to provide different distributions for different needs, but to provide one (the reference one) and everyone should feel free to customize it according to their needs.
I can rephrase a little, but that's one of the reasons why

wolf4ood (approved):

docs/development/decision-records/2024-10-07-azure-distribution/README.md
rafaelmag110: I understood what you meant. Just don't want this to be seen as a "you are free to develop your own stuff if you need it" but rather a "it's not our intention to do it and we made it as easy as possible for you to do it"

docs/development/decision-records/2024-10-07-azure-distribution/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+We will stop to provide an azure flavoured distribution of the connector.
+
+## Rationale
+
+Two reasons are pushing toward this decision:
+- Tractus-X Connector is the reference implementation for Catena-X: this project is not meant to satisfy different needs
+  but to show how an EDC Connector distribution for Catena-X can be assembled. Our focus should point toward maintaining
+  a single distribution.
+- Tractus-X is an open source software, and we should focus on a distribution that involves OSS services.
```
paullatzelsperger, line 13: I would also add two more reasons: 
- for the past year there were no resources to stand up an integration test environment, i.e. an actual Azure Vault instance
- there is no reason for favoritism toward Azure Vault over other services like AWS SecretsManager or similar

bmg13 (approved):

paullatzelsperger (approved):
🪓

PR #1633, opened 2024-10-17
fix: Synchronize EDR refresh

### LINKED ISSUE #1618: EDR Token Refresh auto refresh result is non deterministic
## Describe the bug

The GET [/edrs/{transferProcessId}/dataaddress](https://app.swaggerhub.com/apis/eclipse-tractusx-bot/tractusx-edc/0.7.6#/Control%20Plane%20EDR%20Api/getEdrEntryDataAddress) using `auto_refresh` as `true` should always return a non expired access and refresh token pair. For this to happen, the connector will verify the expiry of the token and trigger a refresh when its due. 

In practice, multiple clients can request the same EDR in an (quasi-)parallel manner, i.e. a job that spans to multiple workers and all require the same EDR. When such happens, and the requested EDR is in a expired state, all the requests will trigger the automatic refresh of same token. Here is where the problem occurs.

Considering the ideal requesting scenario:

Client A issues the refresh request, and token A' is created and stored in the server.
Client B issues the refresh request, and token B' is created and stored in the server.
Client A received token A' and stored it in the database.
Client B recevied token B' and stored it in the database.

Server has B' as the active token and the clients have B' as the active token. The system state is as expected.

However, the following can happen:

Client A issues the refresh request, and token A' is created and stored in the server.
Client B issues the refresh request, and token B' is created and stored in the server.
For some reason, Client B might received token B' first and store it in the database.
When client A receives token A', and stores it in the database, the system state will be inconsistent, as the server has B' as the valid token.

From this point onward, all token validations will fail, rendering the tokens useless and preventing any new token refresh on the same EDR.

## Context Information

Problem is found in EDC versions 0.7.X and up.

rafaelmag110:
## WHAT

Introduces a synchronization mechanism with database locks that only allows one request to update an expired token at a time. All other requests wait until the first request updates the token and releases the database.

## WHY

To fix the issue described in #1618

## FURTHER NOTES

PR to bugfix/0.7.7 first because it's a priority to have that merged.

Closes #1618


core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/service/EdrServiceImpl.java
```diff
@@ -74,12 +74,25 @@ private ServiceResult<DataAddress> autoRefresh(String id, DataAddress edr, Refre
...
+                            monitor.debug("Dont need to refresh. Will resolve existing.");
+                            var refreshedEdr = edrStore.resolveByTransferProcess(id);
+                            return ServiceResult.from(refreshedEdr);
+                        } else {
+                            monitor.debug("Token '%s' expired, need to refresh.".formatted(id));
+                            return tokenRefreshHandler.refreshToken(id, edr)
+                                    .compose(updated -> updateEdr(edrEntry, updated));
+                        }
+                    });
+            edrLock.releaseLock(id);
```
paullatzelsperger, line 90: please handle the result somehow, e.g.
```suggestion
            edrLock.releaseLock(id).onFailure(f -> monitor.warning("Error releasing lock: %s".formatted(f.getFailureDetail())
```

rafaelmag110: [42c1d78](https://github.com/eclipse-tractusx/tractusx-edc/pull/1633/commits/42c1d78d8ab47b6a92e845073a595c6fb42bc579)

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/DefaultEdrLockProviderExtension.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.edc.edr.spi.store.EndpointDataReferenceEntryIndex;
+import org.eclipse.edc.runtime.metamodel.annotation.Extension;
+import org.eclipse.edc.runtime.metamodel.annotation.Inject;
+import org.eclipse.edc.runtime.metamodel.annotation.Provider;
+import org.eclipse.edc.runtime.metamodel.annotation.Provides;
+import org.eclipse.edc.spi.system.ServiceExtension;
+import org.eclipse.edc.transaction.spi.TransactionContext;
+import org.eclipse.tractusx.edc.edr.spi.index.lock.EndpointDataReferenceLock;
+
+@Provides({EndpointDataReferenceLock.class})
```
paullatzelsperger, line 31: not needed - `@Provider` takes care of that.

spi/edr-spi/src/testFixtures/java/org/eclipse/tractusx/edc/edr/spi/testfixtures/index/lock/EndpointDataReferenceLockBaseTest.java
```diff
@@ -0,0 +1,92 @@
...
+    @Test
+    @DisplayName("Verify isExpired Returns true when expired")
+    void verify_isExpiredReturnTrueWhenExpired() {
+        var result = getStore().isExpired(edr("-1000"), edrEntry("mock", "mock"));
+        assertThat(result).isTrue();
+    }
+
+    @Test
+    @DisplayName("Verify isExpired Returns true when not expired")
+    void verify_isExpiredReturnTrueWhenNotExpired() {
```
lgblaumeiser, line 73: Just by chance, shouldn't this be ReturnFalseWhenNotExpired?

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+
+    @Override
+    public StoreResult<Void> releaseLock(String edrId) {
+        LOCK.writeLock().lock();
+        try {
+            var reentrantReadWriteLock = lockedEdrs.get(edrId);
+            if (reentrantReadWriteLock == null) {
+                return StoreResult.generalError("Could not release row-lock because it does not exist");
```
paullatzelsperger, line 71: not sure this is really an error.

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public StoreResult<Boolean> acquireLock(String edrId, DataAddress edr) {
+        LOCK.writeLock().lock();
+        try {
+            var rowLock = lockedEdrs.computeIfAbsent(edrId, k -> new ReentrantReadWriteLock());
+
+            LOCK.writeLock().unlock(); // unlocks global lock before waiting
+            rowLock.writeLock().lock();
+            LOCK.writeLock().lock();
```
paullatzelsperger, line 54: is acquiriing this `LOCK` really needed? the global `LOCK` synchronizes access to the `lockedEdrs` map, but between here and the end of the try block there is no more access to it, so `LOCK` could be freed after obtaining `rowLock`

core/edr-core/src/test/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLockTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.edc.junit.annotations.ComponentTest;
+import org.eclipse.edc.query.CriterionOperatorRegistryImpl;
+import org.eclipse.edc.transaction.spi.NoopTransactionContext;
+import org.eclipse.edc.transaction.spi.TransactionContext;
+import org.eclipse.tractusx.edc.edr.spi.index.lock.EndpointDataReferenceLock;
+import org.eclipse.tractusx.edc.edr.spi.testfixtures.index.lock.EndpointDataReferenceLockBaseTest;
+import org.junit.jupiter.api.BeforeEach;
+
+@ComponentTest
+class InMemoryEdrLockTest extends EndpointDataReferenceLockBaseTest {
```
paullatzelsperger, line 32: tests for `releaseLock` missing

paullatzelsperger:
Actually, regarding the inmem lock; if access to the hashmap is atomic, the global lock isn't needed anymore.

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
rafaelmag110: I don't have a strong opinion on this either. If we are to handle the result from the call in the `EdrServiceImpl`, my guess was that we could at least be informed that some thread is trying to release the lock when it does already exist...

spi/edr-spi/src/testFixtures/java/org/eclipse/tractusx/edc/edr/spi/testfixtures/index/lock/EndpointDataReferenceLockBaseTest.java
rafaelmag110: You are right. The test case was correct but the method and display names where off. Thanks!

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
rafaelmag110: You are right. Added the lock because the `finally` block would fail if the thread didn't had the global lock. But could had removed the try block instead...
Anyways, the global lock is gone.

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
rafaelmag110: If we handle it with a success, then it has no point in handling the result of it at all. The method doesn't have any return case that has to be handled...

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
```diff
@@ -45,43 +43,34 @@ public InMemoryEdrLock(EndpointDataReferenceEntryIndex entryIndex, TransactionCo
...
-        } finally {
-            LOCK.writeLock().unlock();
+        var reentrantReadWriteLock = lockedEdrs.get(edrId);
+        if (reentrantReadWriteLock == null) {
+            return StoreResult.generalError("Could not release row-lock because it does not exist");
         }
+        if (reentrantReadWriteLock.writeLock().isHeldByCurrentThread()) {
+            reentrantReadWriteLock.writeLock().unlock();
+            if (!reentrantReadWriteLock.hasQueuedThreads()) {
+                lockedEdrs.remove(edrId);
```
paullatzelsperger, line 69: access needs to be atomic here, i.e. `Map#computeIfPresent`

core/edr-core/src/test/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLockTest.java
rafaelmag110: [64c76a9](https://github.com/eclipse-tractusx/tractusx-edc/pull/1633/commits/64c76a920083f1bb113088b65264520b2d909fb3)

core/edr-core/src/main/java/org/eclipse/tractusx/edc/edr/core/lock/InMemoryEdrLock.java
rafaelmag110: [64c76a9](https://github.com/eclipse-tractusx/tractusx-edc/pull/1633/commits/64c76a920083f1bb113088b65264520b2d909fb3)

rafaelmag110:
> Actually, regarding the inmem lock; if access to the hashmap is atomic, the global lock isn't needed anymore. We either don't implement any locking in the inmem variant, or we use a concurrent hashmap w rowlevel locks, or a global lock. In sequence of (my personal) preference :)

Can't see how we can make it without row-level locks. Remember the row has to keep locked until the token is refreshed, that's how we keep its state consistent. I went for option two from your list.

paullatzelsperger:
> Can't see how we can make it without row-level locks. Remember the row has to keep locked until the token is refreshed, that's how we keep its state consistent. I went for option two from your list.

we could completely leave out locking altogether - this is the in-mem store, and it's not suited for production anyways.

without row-level locks (i.e. only a global one), we'd effectively synchronize on the EDR store, which to me would be fine because - again this is in-mem-land.

Option 2 is find though.

paullatzelsperger (approved):

wolf4ood (approved):

rafaelmag110:
all ✅ 
🚀

PR #1650, opened 2024-10-24
build(deps): bump EDC to 0.10.0

ndr-brt:
## WHAT

Bumps EDC to stable 0.10.0 version

## WHY

## FURTHER NOTES


lgblaumeiser (approved):
Like that :-)

PR #1688, opened 2024-11-25
feat: strict policy definition validation

### LINKED ISSUE #1685: Introduce strict policy validation
## WHAT

Upstream EDC now supports optional strict policy validation when posting a policy definition [ref](https://github.com/eclipse-edc/Connector/pull/4621).

We should enable it for improving user experience.

## WHY
Improves user experience 

## HOW

// if possible, outlines a solution proposal

## FURTHER NOTES

wolf4ood:
## WHAT

Enables strict policy definition validation by default. If necessary the validation can be disabled with helm configuration/ env variables.

Two additional Constraint functions were introduced:

- `ContractReferenceConstraintFunction`  (`cx-policy:ContractReference`)
- `UsagePurposeConstraintFunction` (`cx-policy:UsagePurpose`)

Both evaluate to `true` always, they are only there as placeholder for enabling the  validation of policy definitions.

## WHY

user experience

## FURTHER NOTES

The current validation only checks if there are functions that can be evaluated with the posted policy definition and the left operand is bound to at least a policy context.

Additional validation could be introduced in policy functions by implementing `AtomicConstraintRuleFunction#validate` which allows validating for example right operand values and operators supported. 

Closes #1685


edc-extensions/bdrs-client/src/main/java/org/eclipse/tractusx/edc/identity/mapper/BdrsClientImpl.java
```diff
@@ -120,7 +120,7 @@ private boolean isCacheExpired() {
     private Result<Void> updateCache() {
         var membershipCredToken = createMembershipPresentation();
         if (membershipCredToken.failed()) {
-            return membershipCredToken.mapTo();
+            return membershipCredToken.mapEmpty();
```
rafaelmag110, line 123: Any reason to use `mapEmpty` instead of `mapFailure`? I'd personally prefer the latter since IMO it's more indicative of a failure.

rafaelmag110 (commented):
Initial impressions.

edc-extensions/bdrs-client/src/main/java/org/eclipse/tractusx/edc/identity/mapper/BdrsClientImpl.java
wolf4ood: No reason :)

ndr-brt (approved):

rafaelmag110 (approved):

PR #1706, opened 2024-12-04
fix: add default empty array value for null policy profiles

ndr-brt:
## WHAT

Add migration to fix the problem of null `profiles` column.

## WHY

Avoid NPE

## FURTHER NOTES

- there's a workaround for this error that can be applied before an 0.8.1 is released. Added a description in the migration guide.


wolf4ood (approved):

docs/migration/Version_0.7.x_0.8.x.md
```diff
@@ -38,6 +38,9 @@ For those who are not using either the Helm Charts or the provided [`migration`
 - [federated catalog - init schema](../../edc-extensions/migrations/control-plane-migration/src/main/resources/org/eclipse/tractusx/edc/postgresql/migration/federatedcatalog/V0_0_1__Init_FederatedCatalogCache_Database_Schema.sql)
 - [policy monitor - create state index](../../edc-extensions/migrations/control-plane-migration/src/main/resources/org/eclipse/tractusx/edc/postgresql/migration/policy-monitor/V0_0_2__Alter_PolicyMonitor_CreateStateIndex.sql)
 - [policy definition - add profiles column](../../edc-extensions/migrations/control-plane-migration/src/main/resources/org/eclipse/tractusx/edc/postgresql/migration/policy/V0_0_5__Add_Profiles.sql)
+  - **WARNING**: in version 0.8.0 this could give runtime errors, eventually you can manually run the query contained in the
```
rafaelmag110, line 41:
```suggestion
  - **WARNING**: When upgrading from 0.7.X, if your policy definition table already has data, this migration will lead the PolicyDefinitionStore to throw run time errors. As a workaround you can manually run the query contained in the 
```

rafaelmag110 (approved):
Looks good. Just a nit.

PR #1707, opened 2024-12-06
docs(ContractDefinitions): Added description how to use private Properties in asset selectors

gerbigf:
## WHAT

Added description on how to use private properties as asset selectors 

## WHY

The docs are currently incomplete and the behaviour has changed between 0.5 and 0.7.

## FURTHER NOTES

Not using the "privateProperties." prefix will still show the asset in the catalogue but then it's not negotiatable. A separate defect will be opened on that.


gerbigf:
@wolf4ood could you take a close look if I specified the left operant correctly? Is it really "privateProperties.'yourproperty'" with those double quotation marks "''"?

wolf4ood (approved):
@gerbigf LGTM

PR #1719, opened 2024-12-17
docs(policies): Updated the always-pass policy

gerbigf:
## WHAT

Using "Policy" instead of "Set"will result in an error when trying to create a policy. Replacing the Policy with Set fixes the issue. Also the policy is now aligned with the other examples in the repo. 

## WHY

The current version was incorrect.


rafaelmag110 (approved):
Please update the deps file and we are good to go.

bmg13 (approved):

gerbigf:
@rafaelmag110  done - is that it?

gerbigf:
@rafaelmag110 tried that and took the latest dependencies from main - still failing...

PR #1721, opened 2024-12-19
docs(mgmt): add static header config

arnoweiss:
## WHAT

This PR documents a feature that allows a Data Provider to add constant headers attached by the Provider Data Plane request to the backend.

## WHY

This feature has been available for a long time and may be useful for non-productive scenarios where a backend API key is attached to the backend request.

## FURTHER NOTES

https://github.com/eclipse-edc/Connector/blob/243ebc9b64fb1ccc08d3a78ebe669c5569175548/spi/data-plane/data-plane-http-spi/src/main/java/org/eclipse/edc/connector/dataplane/http/spi/HttpDataAddress.java#L216


arnoweiss:
`DEPENDENCIES` are failing but not because they're outdated but because of restricted licences. how should I deal with this?

rafaelmag110:
Hey
The file has out-of-date dependencies. That is the reason why the check is failing. The restricted are not the issue here.

rafaelmag110 (approved):

PR #1742, opened 2025-01-13
chore: add generated files in 'bin' folders to gitignore

lgblaumeiser:
## WHAT

I had lots of generated files in my workspace, all of them in 'bin' folders. Added the folders to the .gitignore file to get rid of false proposals of unstaged files.

## WHY

To keep the stage area clean

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,4 +112,7 @@ buildNumber.properties
 **/*.lock
 **/*.tgz
 
+# Generated bin folders
+**/bin/
+
 edc-tests/miw-tests/src/test/resources/docker-environment/postgres_data/


ndr-brt (commented):
which kind of file where they? who was generating them? gradle puts everything into `build/`

ndr-brt:
> > which kind of file where they? who was generating them? gradle puts everything into `build/`
> 
> I did nothing special, only thing is, that I use Visual Studio Code as IDE

yeah, it looks to be an issue with `vscode-java`: https://github.com/redhat-developer/vscode-java/issues/634

.gitignore
```diff
@@ -112,4 +112,7 @@ buildNumber.properties
 **/*.lock
 **/*.tgz
 
+# Generated bin folders
+**/bin/
```
ndr-brt, line 116:
```suggestion
bin/
```

this is enough to tell "all the bin folders in root and all the subfolders"

.gitignore
```diff
@@ -112,4 +112,7 @@ buildNumber.properties
 **/*.lock
 **/*.tgz
 
+# Generated bin folders
```
ndr-brt, line 115: Given that we have a `# IntelliJ` section, it could deserve a `# vscode` one

bmg13 (approved):

ndr-brt (approved):

PR #1747, opened 2025-01-16
docs(usage): Add documentation about response channel usage

### LINKED ISSUE #1746: Documentation describing how to use response channels could be added
## WHAT

We could add a piece of documentation that motivates and describes the usage of the connector transfer process response channels feature, introduced here https://github.com/eclipse-edc/Connector/pull/4579.

## WHY

This is aligned with a feature request https://github.com/eclipse-tractusx/sig-release/issues/936 for release 25.03, so a detailed explanation can be found there.

rafaelmag110:
## WHAT

This PR adds a new document detailing the usage of EDC Connector transfer process response channels.

## WHY

This is aligned with a feature request https://github.com/eclipse-tractusx/sig-release/issues/936 for release 25.03.
It demonstrates tractusx-edc users how to leverage this feature in any use case.

## FURTHER NOTES

I did not add the necessary connector config directly to the helm charts because I realize it might not be of the interest of every connector operator to have it enabled. This can change in the future, so the best option IMO is to wait for user feedback.

Closes #1746


docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
+# Using transfer process response channels
+
+In certain cases, a dataspace consumer might need to send feedback messages to a transfer process counterpart.
+This bidirectional exchange of information requires two separate channels: one forward channel where data flows in the
+direction of the consumer from the provider, and a response channel (back channel) where data flows in the reverse
```
lgblaumeiser, line 5: stumbled above the original phrasing and it took a while to detect that it is not the wrong direction.
```suggestion
direction from the provider to the consumer, and a response channel (back channel) where data flows in the reverse
```

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+direction.
+The EDC Connector enables such bidirectional flow of information.
```
lgblaumeiser, line 7: I do not get this sentence, what is the EDC Connector? What does it mean, that this is enabled? Why is this information important here, isn't it explained in the remainder of the documentation?

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+
+As of writing of this document, a known workaround is in place to enable the same level of communication. The provider
+creates an additional dataset offer, offering its consumer access to an endpoint where messages can be received via HTTP
+POST requests. Albeit the creation of unnecessary offers, it is not advisable to use case architects setup their use
+cases in such manner, as per the dataspace reference architecture the flow of data should always happen in the direction
+of the consumer from the provider.
```
lgblaumeiser, line 13:
```suggestion
from the provider to the consumer.
```

lgblaumeiser: Actually, for me the reason is, that it disconnects the transfer from the feedback. The existing solution does not change this reversal, does it? Still information flows from the consumer to the provider.

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+
+Using a transfer process response channel guarantees compliance and eliminates unnecessary dataset offers and all the
+extra steps needed to obtain access to the underlying endpoints.
+
+## HTTP response channels
+
+The HTTP wire protocol is the only supported protocol for response data communication as of now. Given the bellow
```
lgblaumeiser, line 20:
```suggestion
The HTTP wire protocol is the only supported protocol for response data communication as of now. Given the below
```

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+mentioned considerations, and by following the bellow demonstrated process, a dataspace consumer can send response
```
lgblaumeiser, line 21:
```suggestion
mentioned considerations, and by following the below demonstrated process, a dataspace consumer can send response
```

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+
+- A `HttpData` type asset is offered by the provider, exposing the base url of the backend application. This asset
+  should enable the proxying of at least the request method, path, and body.
+- The backend application exposes an `/responseChannel` endpoint, where messages will be wired to by the dataplane.
+- A valid contract agreement exists for the mentioned offer.
+
+## Getting an EDR with response channel properties
+
+Given the above requirements are met, the consumer client is able to start a transfer process and indicate the intention
+to receive response channel information is its EDR. For this to happen, the transferType attribute of the transfer
```
lgblaumeiser, line 42:
```suggestion
to receive response channel information in its EDR. For this to happen, the transferType attribute of the transfer
```

lgblaumeiser (changes requested):
In general, I would shorten the introduction, as this is a documentation, the references to the current possibilities will outdate soon. I would simply state the use case that is provided with this feature, not what it replaces.

The explanation on what it replaces would be perhaps a mechanism for a blog post in the Tractus-X webpage or something similar, that addresses the transition from previous practices to the new feature.

docs/usage/management-api-walkthrough/09_responsechannels.md
bmg13: I get it that the tp disconnect exists and this workaround keeps it, but I would say that it is clearer to indicate the reversal. Although not "directly", the provider expects (or better, can expect) feedback from the consumer, as Rafael and you mentioned, so is good to keep that.

lgblaumeiser: My point is:

Current situation:
Provider offers a contract offer which provides the access to the asset
Provider offers a contract offer which provides access to a POST Rest endpoint to send feedback. This  "somehow" contradicts the idea that the participant offering data is providing the contract offer, right?

New situation:
Provider offers a contract offer which provides the access to the asset and the access to a POST Rest endpoint for sending feedback. So from a contract perspective, nothing has changed concerning the feedback, it is still based on a contract offer provided by the provider of the original data or did I miss something?

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
+# Using transfer process response channels
+
+In certain cases, a dataspace consumer might need to send feedback messages to a transfer process counterpart.
```
ndr-brt, line 3: The need is not from the perspective of who sends the feedback, but from who receives it, e.g.:
- provider push: the provider needs to get some feedback, so it defines a transfer with a backchannel
- consumer pull: the provider needs to get some feedback, so it defines a transfer with a backchannel

so it's always the provider who need to get feedback messages from the consumer.

docs/usage/management-api-walkthrough/09_responsechannels.md
ndr-brt: maybe EDC connector could be replaced by "Tractus-X EDC connector", or just "connector", but to be honest this sentence could also be removed altogether, as it's implicit in this piece of doc :)

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+  "@context": {
+    "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
+  },
+  "@type": "TransferRequest",
+  "assetId": "{{ASSET_ID}}",
+  "contractId": "{{CONTRACT_AGREEMENT_ID}}",
+  "counterPartyAddress": "{{CONNECTOR_ADDRESS}}",
+  "dataDestination": {
+    "type": "<supported-transfer-type>"
+  },
```
ndr-brt, line 59: `dataDestination` is not needed for `PULL` transfer types ar at least it should not :)

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+  "protocol": "dataspace-protocol-http",
+  "transferType": "HttpData-PULL",
+  "callbackAddresses": []
+}
+```
+
+Changing the `transferType` from `HttpData-PULL` to `HttpData-PULL-HttpData`, indicates the intention to receive
+response channel details. Other possible `transferType` combinations can be discovered within `dcat:distribution`
+when requesting a providers' catalog.
```
ndr-brt, line 68: This should not be a consumer decision, when the provider defines a backchannel for a specific asset, in the catalog there should be only the related `distribution`, and the consumer will be forced to consume it whether it want to set up a transfer.

docs/usage/management-api-walkthrough/09_responsechannels.md
gerbigf: That's correct but I guess Rafael wanted to express that not in all cases feedback is needed?

docs/usage/management-api-walkthrough/09_responsechannels.md
gerbigf: @lgblaumeiser indeed as long as we use the Consumer Push flow we need separate offers for the data and the feedback. Feedback Channels don't change anything as long as we don't have infinite provider push.

gerbigf: But it might be worth noting that using a response channel only makes sense for the regular Consumer Pull where the consumer afterwards wants to give feedback?

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+
+## HTTP response channels
+
+The HTTP wire protocol is the only supported protocol for response data communication as of now. Given the bellow
+mentioned considerations, and by following the bellow demonstrated process, a dataspace consumer can send response
+messages to a provider.
+
+## Requirements
+
+- Considering the EDC Connector operator has enabled the HTTP response channel for a certain connector installation, by
```
gerbigf, line 26: Same thing here. I'd call it connector operator.

docs/usage/management-api-walkthrough/09_responsechannels.md
gerbigf, line 42: But isn't the provider the one that provides information about the feedback channel in its EDR?

docs/usage/management-api-walkthrough/09_responsechannels.md
```diff
@@ -0,0 +1,108 @@
...
+
+Changing the `transferType` from `HttpData-PULL` to `HttpData-PULL-HttpData`, indicates the intention to receive
+response channel details. Other possible `transferType` combinations can be discovered within `dcat:distribution`
+when requesting a providers' catalog.
+
+## Sending messages to the response channel
+
+Once the right EDR has been identified using the EDR Management API, it should contain additional details with response
+channel information. We can then proceed with the same process as for any other transfer using an [EDR](07_edrs.md), but
+using `responseChannel-endpoint` instead of `endpoint` and `responseChannel-authorization` instead of `authorization`.
```
gerbigf, line 74: I would highlight that the initial transfer still works as is _and additionally_ after the data has been processed the information from the response channel endpoint can be used to send the feedback.

docs/usage/management-api-walkthrough/09_responsechannels.md
lgblaumeiser: Hi @gerbigf , that is still not my point. What @rafaelmag110 described here is, that there is this unholy pattern, that a data sink provides an asset to which data is sent and that the back channels would do something to prevent this pattern.

This is actually not the case, it just integrates the back channel usage into the original contract offer but that still means that the data flow is from the consumer to the provider.

Which brings me back to my proposed advantage, the back channels allow to bundle a data transfer with a back channel, so that there is a connection between the two which is not the case in the old fashion. Therefore, the unholy pattern is prevented because it is hidden in the original contract definition and not required explicitely. But it is still there.

docs/usage/management-api-walkthrough/09_responsechannels.md
rafaelmag110: Yes, that is what I wanted to mention. But Andrea is correct. I'll rephrase this so its agnostic of who actually needs it, but that the need exists.

docs/usage/management-api-walkthrough/09_responsechannels.md
rafaelmag110: I agree with your statement. I wrote this doc having the current implementation of the response channels as reference. In the current implementation, a provider can't set a response channel for a particular asset only. It's either enabled for all assets or it isn't, as the dataplane that exposes the response channel endpoint sends that information to the control plane selector and that is used to compute the possible distributions in the catalog.

What I wrote here might not be where we want to go, but is not invalid. Now is up to the consumer to define if the response channel information should be sent or not, as both distributions are part of the catalog it receives...

docs/usage/management-api-walkthrough/09_responsechannels.md
rafaelmag110: I added more information to the paragraph, to explain in the introduction how the connector allows the bidirectional flow of information.
@lgblaumeiser lmk if it is more clear this time around.

rafaelmag110:
@ndr-brt @lgblaumeiser @gerbigf @bmg13 
Thanks for your review and suggestions. I addressed the direct ones and resolved the open conversation. I left the ones are still ongoing, open.

I'd appreciate your thoughts on the changes.

docs/usage/management-api-walkthrough/09_responsechannels.md
lgblaumeiser: It should also not be on the developers, right 😄 , it is the providers decision to offer the same asset potentially with multiple transfer types, isn't it. But that is actually a general thought on how to provide assets, it is not necessary in this documentation, as here the point is to explain how a provider can set up a backchannel, right?

lgblaumeiser (approved):
I think, this is now in a pretty good shape, I already approve, although there is one open comment where I stated my opinion, but I could live with this changed or not.

docs/usage/management-api-walkthrough/09_responsechannels.md
bmg13: If is not the current approach and since this is a "limitation"

> a provider can't set a response channel for a particular asset only.

maybe we should not add to the doc the provider ability to expose the backchannel and leave the responsability for the consumer (for now, atleast).
The other approach would be to enable or disable backchannel for all assets. But I would prefer a case-by-case (asset-by-asset) approach and the consumer must be the one deciding it.

ndr-brt: > I agree with your statement. I wrote this doc having the current implementation of the response channels as reference. In the current implementation, a provider can't set a response channel for a particular asset only. It's either enabled for all assets or it isn't, as the dataplane that exposes the response channel endpoint sends that information to the control plane selector and that is used to compute the possible distributions in the catalog.

this is definitely something that needs to be tackled upstream, as it should be a provider decision to state if an asset will provide a back channel.
Opened an issue: https://github.com/eclipse-edc/Connector/discussions/4744

I would add this info to the documentation (currently it is like this, but it could become that)

bmg13 (approved):

ndr-brt (approved):

rafaelmag110:
@ndr-brt @lgblaumeiser
Added a new line after you approved, as we discussed during the weekly. Please have a look and merge if you think is ready to go.

PR #1748, opened 2025-01-16
feat: Update dataplane logs to allow better transfer process traceability

bmg13:
## WHAT

This PR adds logic to allow to filter a transfer process flow from a single id (TP ID) for dataplane by adding said id in the logs.

## WHY

This will improve troubleshooting scenarios of a transfer process and improve overall traceability.

## FURTHER NOTES

This PR does not add any breaking changes, however it does pull the entire dataplane-core module from Upstream to Tractus-X.

Logs updated are:
- Transfer from source to destination;
- Unknown data source type 
- Unknown data sink type

Tested with local distribution and the following are outcome logs containing the TP id in the DP transfer source-destination log.

highligHt for the log
```
provider-data-plane-1        | DEBUG 2025-01-16T12:51:47.161051717 Transferring from AmazonS3 to AmazonS3 for DataFlow 7926e16e-7357-4032-8078-db08ecd9d67f.
```

Screenshot 2025-01-16 at 12 56 10

Closes #https://github.com/eclipse-tractusx/tractusx-edc/issues/1749

### DIFF
--- a/edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java
+++ b/edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java
@@ -112,8 +112,8 @@ public class PipelineServiceImpl implements PipelineService {
 
         var source = sourceFactory.createSource(request);
         sources.put(request.getProcessId(), source);
-        monitor.debug(() -> format("Transferring asset %s from %s to %s for DataFlow %s.",
-                request.getAssetId(), request.getSourceDataAddress().getType(), request.getDestinationDataAddress().getType(), request.getProcessId()));
+        monitor.debug(() -> format("Transferring from %s to %s for DataFlow %s.",
+                request.getSourceDataAddress().getType(), request.getDestinationDataAddress().getType(), request.getProcessId()));
         return sink.transfer(source)
                 .thenApply(result -> {
                     terminate(request.getProcessId());
--- a/edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImpl.java
+++ b/edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/manager/DataPlaneManagerImpl.java
@@ -94,7 +94,7 @@ public class DataPlaneManagerImpl extends AbstractStateEntityManager<DataFlow, D
     @Override
     public Result<DataFlowResponseMessage> start(DataFlowStartMessage startMessage) {
         var dataFlowBuilder = DataFlow.Builder.newInstance()
-                .id(startMessage.getProcessId()) // todo: is this the tpi?
+                .id(startMessage.getProcessId())
                 .source(startMessage.getSourceDataAddress())
                 .destination(startMessage.getDestinationDataAddress())
                 .callbackAddress(startMessage.getCallbackAddress())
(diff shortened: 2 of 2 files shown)


edc-extensions/dataplane/dataplane-proxy/edc-dataplane-proxy-consumer-api/src/main/java/org/eclipse/tractusx/edc/dataplane/proxy/consumer/api/asset/ConsumerAssetRequestController.java
```diff
@@ -69,7 +69,6 @@ public class ConsumerAssetRequestController implements ConsumerAssetRequestApi {
     public static final String EDR_ERROR_MESSAGE = "No EDR for transfer process: %s : %s";
     private static final String ASYNC_TYPE = "async";
     private static final String HEADER_AUTHORIZATION = "header:authorization";
-    private static final String BEARER_PREFIX = "Bearer ";
```
bmg13, line 72: removed simply because was not being invoked

bmg13:
> please do not create PRs with a discussion/issue before, and they would need proper triaging

You're right, was still writing it. Updated the description accordingly.

edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/DataPlaneDefaultServicesExtension.java
```diff
@@ -0,0 +1,86 @@
...
+        return new InMemoryDataPlaneStore(clock, criterionOperatorRegistry);
+    }
+
+    @Provider(isDefault = true)
+    public AccessTokenDataStore defaultAccessTokenDataStore() {
+        return new InMemoryAccessTokenDataStore(criterionOperatorRegistry);
+    }
+
+    @Provider(isDefault = true)
+    public PipelineService pipelineService(ServiceExtensionContext context) {
```
rafaelmag110, line 71: From what I was able to see from this PR, the `PipelineServiceImpl` was the only class that really changed from the whole "override".
Since the default `PipelineService` is provided as default here, you can simply create an extension which implements a provider for this interface. Better for maintainability.

edc-dataplane/edc-dataplane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/pipeline/PipelineServiceImpl.java
```diff
@@ -0,0 +1,197 @@
...
+    @Override
+    public CompletableFuture<StreamResult<Object>> transfer(DataFlowStartMessage request, DataSink sink) {
+        var sourceFactory = getSourceFactory(request);
+        if (sourceFactory == null) {
+            return noSourceFactory(request);
+        }
+
+        var source = sourceFactory.createSource(request);
+        sources.put(request.getProcessId(), source);
+        monitor.debug(() -> format("Transferring from %s to %s for DataFlow %s.",
```
rafaelmag110, line 115: This is a nit, just to keep the log pattern.
It seems the EDC usually logs the ID first and then the message attached to that. So it would be something like:
`"Dataflow %s Transferring from %s to %s". `

rafaelmag110 (changes requested):
Please check my comment regarding the possibility to just make this as an extension that adds a `PipelineService` provider.
Also, I wasn't expecting these changes to be so few, on only to a couple of log messages. Maybe adding them directly upstream wouldn't make such a fuss?

ndr-brt (changes requested):
please close this PR so we can go through a correct `triage` phase before

PR #1761, opened 2025-01-21
feat: Add the folder path to the vault helm values

### LINKED ISSUE #1758: Add the folder path to the vault helm values
## WHAT

The folder path for the vault should be added to the vault configuration. The folder path specifies the folder structure after the VAULT_SECRET_DATA_PATH. 

In the EDC Dataspace Connector the folder path is added to the the secret URL, when it is not null.

```java
if (folderPath != null) {
    builder.addPathSegments(PathUtil.trimLeadingOrEndingSlash(folderPath));
}
```

For example with the following specs:
- secretPath: `v1/secret`
- entryType/VAULT_SECRET_DATA_PATH: `data`
- folderPath: `department-1`
- key: `token`

the secret url should be `v1/secret/data/department-1/token`.

## WHY
Currently, it is not possible to specify the folder path, so all edc secrets must be stored directly in the data path.

I want to specify access policies for specific vault paths and this is not possible with the current configuration. A workaround would be to add a second data path after the mandatory data path and my preferred folder structure, but this is not a good practice.

Example:

I have a sales department, which should only be able to read and update its own secrets. The secrets are stored under the path secret/data/sales, but with the current EDC configuration it is not possible to use this folder structure.

## HOW

The folder path is already implemented in the Eclipse EDC with the settings key `edc.vault.hashicorp.folder`. It needs only to be added to the helm chart for example as the value `vault.hashicorp.paths.folder` and to the environment variables of the deployments of the control plane and the data plane.

AndrYurk:
## WHAT

Add the vault.hashicorp.paths.folder to the helm chart

## WHY

To allow specify the folder path, so all edc secrets could be stored directly in the data path.

Closes #1758


lgblaumeiser (commented):
One question: Perhaps I miss something, but this change adds a configuration value to the helm charts, I miss the use of this configuration parameter somewhere. My understanding is, that the value has to be added to the construction of the access path to the vault in the connector, right? Does this already exist? Is this an upstream PR that is referenced?

AndrYurk:
@lgblaumeiser, yes, we already have this in the connector
Declaration:
```java
@Setting(description = "The path of the folder that the secret is stored in, relative to VAULT_FOLDER_PATH", required = false, key = "edc.vault.hashicorp.folder")
private String folderPath;
```
https://github.com/eclipse-edc/Connector/blob/08f913e81b56ffa474783cf279cabe5fc0756a4b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultSettings.java#L54-L56

Usage:
```java
var folderPath = settings.getFolderPath();

        var builder = baseUrl(settings)
                .newBuilder()
                .addPathSegments(PathUtil.trimLeadingOrEndingSlash(vaultApiPath))
                .addPathSegment(entryType);

        if (folderPath != null) {
            builder.addPathSegments(PathUtil.trimLeadingOrEndingSlash(folderPath));
        }
```
https://github.com/eclipse-edc/Connector/blob/08f913e81b56ffa474783cf279cabe5fc0756a4b/extensions/common/vault/vault-hashicorp/src/main/java/org/eclipse/edc/vault/hashicorp/client/HashicorpVaultClient.java#L280C13-L289

lgblaumeiser (approved):
lgtm

PR #1774, opened 2025-01-30
docs(dr): Add DR about empty assetSelector validator

rafaelmag110:
## WHAT

Adds decision record about adding a validator to block contract definitions with empty asset selector.

## WHY

Part of https://github.com/eclipse-tractusx/tractusx-edc/issues/1773. Please refer to that issue for more details.

PR #1786, opened 2025-02-05
feat: add contract definition empty asset selector validator

### LINKED ISSUE #1773: Block contract definitions with empty assetSelector
## WHAT

Tractusx-edc should provide a feature that blocks the creation/update of contract definitions to contain
an empty asset selector. 
This feature should have the option to be switched on/off, being off the default state.

## WHY
The `assetsSelector` property is a query (similar to a SQL SELECT statement) that returns a set of assets the 
contract definition applies to. When left empty, the default behavior of the connector is to apply the contract
definition to all registered assets.

Unwillingly, mostly due to lack of proper connector usage knowledge, users might expose datasets that shouldn't be
exposed.

rafaelmag110:
## WHAT

Adds a validator that replaces the existing contract definition validator for the data management API.

## WHY

Implementation of https://github.com/eclipse-tractusx/tractusx-edc/pull/1774

Closes #1773


edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
```diff
@@ -0,0 +1,60 @@
+# Contract Definitions Validator: Empty Asset Selector
+
+The goal of this extension is to provide a replacement validator for contract definition entities.
+It is used to validate requests that create or update contract definitions via the data management API endpoint.
+When enabled, it prevents contract definitions with no asset selector, or an empty one, from being created.
+
+This extension is included with the standard tractusx-edc distribution, but is disabled by default. To enable it,
```
gerbigf, line 7: I would a bit of text to explain why this is required - most people might not know. 
 
 
I.e. "When enabled, it prevents contract definitions with no asset selector, or an empty one, from being created which elsewise leads to all available assets being included in the Contract Definition"?

gerbigf:
Have we already discussed to enable that per default for Tractus-X? I would strongly support that...

bmg13:
> Have we already discussed to enable that per default for Tractus-X? I would heavily support that...

Believe during technical discussion it was set as default disabled to avoid adding breaking changes. Additionally as default it is seen as a "querying" approach, i e., a filtering view in each no value in asset selector means no filter. Changing this would be adopter's choice.

ndr-brt:
> Have we already discussed to enable that per default for Tractus-X? I would heavily support that...

In the DR we agreed on "off by default": https://github.com/eclipse-tractusx/tractusx-edc/tree/main/docs/development/decision-records/2025-01-30-empty-asset-selector-validator

BTW I don't have strong feelings here

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
```diff
@@ -0,0 +1,60 @@
...
+```json
+{
+  "@context": {
+    "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
+  },
+  "@type": "ContractDefinition",
+  "@id": "myContractDefinitionId",
+  "accessPolicyId": "myAccessPolicyId",
+  "contractPolicyId": "myContractPolicyId"
+  // Asset selector is missing
```
lgblaumeiser, line 23: I stumbled over these formulation, what is meant with the comment, is this the status message returned with the 400 status code, is this a remark, that in the json is something missing? A bit more context would make this easier to understand.

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
```diff
@@ -0,0 +1,60 @@
...
+{
+  "@context": {
+    "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
+  },
+  "@type": "ContractDefinition",
+  "@id": "myContractDefinitionId",
+  "accessPolicyId": "myAccessPolicyId",
+  "contractPolicyId": "myContractPolicyId",
+  "assetSelector": []
+  // Asset selector should contain at least one valid criterion
```
lgblaumeiser, line 39: Again, here a bit more context would support the reader to understand what is the issue here.

lgblaumeiser (approved):
I support Felix remark and have another documentation issue. In general, I think, this PR looks good to me, so I approve it already, although I would appreciate if you could improve the documentation remarks!

bmg13 (approved):

edc-extensions/validators/build.gradle.kts
ndr-brt, line 1: I don't see the need for this overarching `validators` extension, it just adds unnecessary complexity and indirection.
plus, it's never really used

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/EmptyAssetSelectorBlockerExtension.java
```diff
@@ -0,0 +1,64 @@
...
+    @Inject
+    Monitor monitor;
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @Override
+    public void start() {
```
ndr-brt, line 57: I got the need to effectively override the default validator, but this is definitely not the correct phase to be used.
I would suggest `prepare` instead.
But, generally speaking, if the way this `validatorRegistry` has been implemented doesn't permit to effectively set up a custom validator because of the extensions loading order, this is something that needs to be fixed upstream, and it could be done with different approaches (priority, validator composition, ...)

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/EmptyAssetSelectorBlockerExtension.java
```diff
@@ -0,0 +1,64 @@
...
+        if (blockerEnabled) {
+            monitor.info("Empty Asset Selector Blocker is enabled.");
```
ndr-brt, line 59: I'd be more verbose like "ContractDefinition validator that blocks empty assetsSelector has been enabled"

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/MandatoryIfNotBypass.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Optional;
+
+import static org.eclipse.edc.connector.controlplane.contract.spi.types.offer.ContractDefinition.CONTRACT_DEFINITION_PRIVATE_PROPERTIES;
+import static org.eclipse.edc.jsonld.spi.JsonLdKeywords.VALUE;
+import static org.eclipse.edc.spi.constants.CoreConstants.EDC_NAMESPACE;
+
+public class MandatoryIfNotBypass implements Validator<JsonObject> {
+
+    private static final String BYPASS_PROPERTY = EDC_NAMESPACE + "allowEmpty";
```
ndr-brt, line 36: we already have a toggle for this feature, I would avoid to overengineer this by adding an additional toggle on the data itself. It's either active or not at startup, or I am missing some pieces here?

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/EmptyAssetSelectorValidator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.edc.validator.jsonobject.validators.model.CriterionValidator;
+
+import static org.eclipse.edc.connector.controlplane.contract.spi.types.offer.ContractDefinition.CONTRACT_DEFINITION_ACCESSPOLICY_ID;
+import static org.eclipse.edc.connector.controlplane.contract.spi.types.offer.ContractDefinition.CONTRACT_DEFINITION_ASSETS_SELECTOR;
+import static org.eclipse.edc.connector.controlplane.contract.spi.types.offer.ContractDefinition.CONTRACT_DEFINITION_CONTRACTPOLICY_ID;
+
+public class EmptyAssetSelectorValidator {
+
+    public static JsonObjectValidator instance(CriterionOperatorRegistry criterionOperatorRegistry) {
+        return JsonObjectValidator.newValidator()
```
ndr-brt, line 35: this is pretty much duplicated from upstream and it makes me think that maybe there could be a better way to tackle that, like have the possibility to provide additional validators to the already existent validators through composition. Not for this PR, but this is maybe a topic that could be raised upstream.

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
```diff
@@ -0,0 +1,60 @@
...
+  "@type": "ContractDefinition",
+  "@id": "myContractDefinitionId",
+  "accessPolicyId": "myAccessPolicyId",
+  "contractPolicyId": "myContractPolicyId",
+  "assetSelector": [],
+  "privateProperties": {
+    "allowEmpty": "assetSelector"
+  }
+}
+```
```
ndr-brt, line 60: this looks unnecessary

edc-extensions/validators/build.gradle.kts
rafaelmag110: You mean the module? I added this with the expectation that more validators could be added here in the future. We can go the other way around and just add this when we have the need for another validator.

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/MandatoryIfNotBypass.java
rafaelmag110: My other comment already addresses this.

edc-extensions/validators/build.gradle.kts
ndr-brt: yes, the module. no issues with the `validators` folder, but the build file and the addition in the "settings" file could be avoided, also with multiple validators module I won't see the advantage of this intermediate module.

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
ndr-brt: I thought I added a comment on the `allowEmpty` thing, but I didn't :sweat_smile:  . I would say that feature is not needed, there's already a feature toggle in config to enable this validation, I don't see the point to add another toggle at the data level to permit to workaround that, because I can see people using it all the time.

If a runtime blocks empty assets selector, that's the behavior the final user will need to deal to

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/EmptyAssetSelectorBlockerExtension.java
rafaelmag110: Created a discussion upstream: https://github.com/eclipse-edc/Connector/discussions/4812

edc-extensions/validators/empty-asset-selector/src/main/java/org/eclipse/tractusx/edc/validators/emptyassetselector/README.md
rafaelmag110: But if people use it all the time, it means they are aware of what they are doing. And that is ok.

In our case, for example, one connector installation might have more than a single user. One user might be knowledgeable about what an empty asset selector means, but the other doesn't. Our decision as a company it that all connector installations prevent contract defs with empty asset selector by default. BUT, if one of our users wants to still workaround the block, he can do the bypass.

ndr-brt: > BUT, if one of our users still wants to workaround the block, he can do the bypass.

I don't see why this should happen, giving the user the possibility to avoid validations or certain logic to me looks like an antipattern, potentially dangerous for no real need

rafaelmag110: How strong are you against this? This was a requirement from Felix when thinking about it. I do understand your points but honestly not against having this.
I was going to say we can always remove it and add it later if there are proper requirements, but the requirement is there.

rafaelmag110: I understand. I'll remove it then.

rafaelmag110:
@ndr-brt 
I have removed the bypass. Please have a second look and merge it if it's good to go.

ndr-brt (approved):

PR #1796, opened 2025-02-10
chore: Add known issues section to readme file

### LINKED ISSUE #1772: Incompatibility with Hashicorp Vault 1.18.1, last working version is 1.16.1
## Describe the bug

As I recently found out, there is a general incompatibility with the Hashicorp Vault 1.18.1. The bug is actually on there side, but current connector versions are not able to work with this vault version. Last known version that works is 1.16.1.

Issue is, that the way, the upstream connector creates the url for a rest call, it uses encoded slashes, resp. %2F. According to the corresponding web standard a system should properly work with plain and encoded slashes, but version 1.18.1 does not. This way actually encountered in the test of the BDRS service, as the tests took the latest version of the vault which failed at some point consistently.

### To Reproduce

1. Configure the deployment tests, of e.g., the BDRS, to use a vault 1.18.1 (resp. the corresponding helm chart)
2. Try to run the tests
3. Observe stange IOExceptions which indicate that an html page was returned at some point instead of a JSON object

### Expected behavior

As the error is on the vault side, the only means is to use a working vault implementation

lgblaumeiser:
## WHAT

Add a note to the readme to indicate, that Hashicorp Vault 1.18.1 is incompatible with the EDC

## WHY

Discussed in the last weekly, see comment in #1772 

Closes #1772

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Build Tractus-X EDC together with its Container Images
 ./gradlew dockerize
 ```
 
+## Known Incompatibilities
+
+- Hashicorp Vault 1.18.1 is not compatible with the EDC due to a bug in the vault concerning path handling
+
 ## License
 
 Distributed under the Apache 2.0 License.
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ Build Tractus-X EDC together with its Container Images
 ## Known Incompatibilities
 
 - Hashicorp Vault 1.18.1 is not compatible with the EDC due to a bug in the vault concerning path handling
+  - [Internal Issue](https://github.com/eclipse-tractusx/tractusx-edc/issues/1772)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,10 @@ Build Tractus-X EDC together with its Container Images
 ./gradlew dockerize
 ```
 
+## Known Incompatibilities
+
+- Hashicorp Vault 1.18.1 is not compatible with the EDC due to a bug in the vault concerning path handling
```
ndr-brt, line 59: it would be great to have some reference to an explaination of this bug or its resolution in next versions of Hashicorp Vault

lgblaumeiser: Added links to our discussion and the upstream issue I have raised.

ndr-brt (approved):

Issue #1783, opened 2025-02-04
bug: catalog generated by Tractus-X EDC 0.8.0 provide substantial changes that are making it incompatible with internal apps developed against the catalog returned by version 0.7.7


suh-rao:
## Describe the bug

When Catalog call made to connector version from 0.8.0 then we get offers response in which we get
"  http://www.w3.org/ns/dcat#distribution : [ ]" , outside of dataset.

### To Reproduce

From 0.7.* connector make a catalog call to 0.8.* connector.
Catalog response looks like below.

```json
{
  "@id": "9f44e306-ad41-4939-97d3-4c3f609d17be",
  "@type": [
    "http://www.w3.org/ns/dcat#Catalog"
  ],
  "http://www.w3.org/ns/dcat#dataset": [
    {
      "@id": "blob-test-asset",
      "@type": [
        "http://www.w3.org/ns/dcat#Dataset"
      ],
      "http://www.w3.org/ns/odrl/2/hasPolicy": [
        {
          "@id": "Y29udHJhY3QtZGVmaW5pdGlvbi1ibG9iLXRlc3QtYWx3YXlzLXRydWU=:YmxvYi10ZXN0LWFzc2V0:Y2IzMGUxOGMtMjg1OS00MWEyLWE3M2YtOTBjNzE5ZTRlYjVm",
          "@type": [
            "http://www.w3.org/ns/odrl/2/Offer"
          ],
          "http://www.w3.org/ns/odrl/2/permission": [
            {
              "http://www.w3.org/ns/odrl/2/action": [
                {
                  "@id": "use"
                }
              ]
            }
          ],
          "http://www.w3.org/ns/odrl/2/prohibition": [],
          "http://www.w3.org/ns/odrl/2/obligation": []
        }
      ],
      "http://www.w3.org/ns/dcat#distribution": [
        {
          "@type": [
            "http://www.w3.org/ns/dcat#Distribution"
          ],
          "http://purl.org/dc/terms/format": [
            {
              "@id": "AzureStorage-PUSH"
            }
          ],
          "http://www.w3.org/ns/dcat#accessService": [
            {
              "@id": "1750127b-a7c9-4b54-94e7-68684d2a247f",
              "@type": [
                "http://www.w3.org/ns/dcat#DataService"
              ],
              "http://www.w3.org/ns/dcat#endpointDescription": [
                {
                  "@value": "dspace:connector"
                }
              ],
[...]

PR #1797, opened 2025-02-12
chore: Add decision record concerning downward compatibiity

lgblaumeiser:
## WHAT

Decision Record on Downward Compatibility

## WHY

As we had currently downward compatibility issues which lead to applications to fail to continue to work, a decision needs to be done about the expectation of the provided service implementation.

### DIFF
--- a/docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
+++ b/docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
@@ -1,0 +1,20 @@
+# Handling of downward compatibility for a DSP version
+
+## Decision
+
+DSP assets, like the catalog, which are forwarded directly to a caller through the management api, have to be kept downward compatible as long as the DSP version through which the asset is retrieved is supported. There is no compatibility requirement between two different DSP versions concerning such assets.
+
+Relevant assets are:
+- Catalog returned by...
+- ...
+
+## Rationale
+
+DSP assets which are directly received by a caller of a management api endpoint is data that is transfered between two dataspace participants and creates a direct relationship between the calling application and the provider connector. As this is the case, the breaking change requirements of Catena-X has to be fulfilled. The requirement states that a consumer can continuously consume data from the provider without the need to be forced to update his system due to a change of the provider service stack. This requires that such assets must not change in a way, that existing properties vanish from the returned data.
+
+As a consumer participant is capable to ensure the usage of a connector with a certain set of DSP versions supported, he is capable to use applications that are capable to support all potential data formats received by the different supported DSP versions. Therefore, the asset data models supported in two different DSP versions are not related and therefore no compatibility requirement is needed between those.
+
+## Approach
+
+- The compatibility tests are enriched by tests that ensure the detection of changes in the corresponding DSP assets.
+- If changes are detected a manual check that the downward compatibility is kept is executed.


docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,20 @@
+# Handling of downward compatibility for a DSP version
+
+## Decision
+
+DSP assets, like the catalog, which are forwarded directly to a caller through the management api, have to be kept downward compatible as long as the DSP version through which the asset is retrieved is supported. There is no compatibility requirement between two different DSP versions concerning such assets.
```
rafaelmag110, line 5: > "have to be kept downward compatible as long as the DSP version through which the asset is retrieved is supported"
Can this statement be explained in different terms?

rafaelmag110 (commented):
I left a comment inline, but I'd like to provide my perspective on why that statement could be explained in different terms.

What we should try to achieve here is a decision on how we handle protocol versions, as in, how many versions a certain connector version should support and how we guarantee no changes are introduced to a currently supported version.

In my perspective there is no point in guaranteeing two distinct DSP versions are downwards compatible. Two achieve this you would need to keep deprecated properties in newer versions, which you shouldn't need in the first place because you also support the previous (deprecated) version. In this sense, keeping up with deprecation is more simple because adopters can look for deprecated DSP versions and not deprecated properties in recent versions.

Also, the requirement from CatenaX is that communication between two participants is not broken due to stack upgrades. If two participants communicate over a specific version of a protocol, one of those participants can only upgrade its stack to a latter version IF the new version also supports the previous protocol version. I believe this is possible with the multi protocol support from the Connector. TractusX responsability then is to define which DSP protocol versions have to be supported for each TractusX release.

In terms of testing, in my perspective tractusx-edc should only guarantee that the protocol data model never changes for an already supported version. This is not really testing for downward compatibility, but for consistency.

rafaelmag110:
Nevertheless, many thanks @lgblaumeiser  for pushing this decision record.

bmg13:
Just leaving my 2 cents here :)

> The requirement states that a consumer can continuously consume data from the provider without the need to be forced to update his system due to a change of the provider service stack. 

As mentioned, the `Deprecation` exists for this reason. To ensure no imediate adoption of a new version is not needed but must considered and eventually changed until the definitve removal. 
So if a Provider bumps the connector version it would not affect the Consumer since deprecated values are supported.

> TractusX responsibility then is to define which DSP protocol versions have to be supported for each TractusX release.

This is one of the reasons the migration guides include potential breaking changes and why I believe "sustaining" version for multiple iterations "indefinatly" a dangerous path, since we should never live in a idea that late adoption of a new version is great. If consistency with all versions are to be kept indeterminably than no one would bump versions.

> This requires that such assets must not change in a way, that existing properties vanish from the returned data.

I do not understood that this was being considered and still think it wasn't. It was the consistency between supported versions and not a "vanishing" from a simple one-sided bump.

That said, really cool this discussion and helps define the tests; as long as supported versions are clearly defined, the tests should cover any of this concerns to happen. Regarding these:
> If changes are detected a manual check that the downward compatibility is kept is executed.

IMO, is very hard to enforce any kind of manual check. The proposed tests should suffice, right?

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
ndr-brt, line 5: "DSP assets" don't exist.
Asset is a concept internal to the EDC domain, in DSP it could be reflected in the "Dataset" term (from DCAT spec)

ndr-brt: yes, we need to be more specific, as here we're not talking about supporting a DSP version, but supporting the outcome of a specific EDC version based on a DSP version. 

e.g. the EDC connector versions subsequent to 0.7.7 need to being able to operate with EDC connector at version 0.7.7 as well as other third party app that were built to operate with version 0.7.7.
Specifying that the connector version implemented by the 0.7.7 connector is a "special" DSP 0.8 ([descibed here](https://github.com/International-Data-Spaces-Association/ids-specification/tree/v0.8)), not 100% compliant with the specs

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Relevant assets are:
+- Catalog returned by...
+- ...
+
+## Rationale
+
+DSP assets which are directly received by a caller of a management api endpoint is data that is transfered between two dataspace participants and creates a direct relationship between the calling application and the provider connector. As this is the case, the breaking change requirements of Catena-X has to be fulfilled. The requirement states that a consumer can continuously consume data from the provider without the need to be forced to update his system due to a change of the provider service stack. This requires that such assets must not change in a way, that existing properties vanish from the returned data.
```
ndr-brt, line 13:
```suggestion
DSP datasets which are obtained by a consumer third party app through the management api endpoint create a direct relationship between the calling application and the provider connector. As this is the case, the breaking change requirements of Catena-X has to be fulfilled. The requirement states that a consumer can continuously consume data from the provider without the need to be forced to update his system due to a change of the provider service stack. This requires that such assets must not change in a way, that existing properties vanish from the returned data.
```

this need to be rephrased:
> As this is the case
which case?

plus, a link to the "breaking change requirements of Catena-X" will be necessary

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+As a consumer participant is capable to ensure the usage of a connector with a certain set of DSP versions supported, he is capable to use applications that are capable to support all potential data formats received by the different supported DSP versions. Therefore, the asset data models supported in two different DSP versions are not related and therefore no compatibility requirement is needed between those.
```
ndr-brt, line 15:
```suggestion
As a consumer participant is capable to ensure the usage of a connector with a certain set of DSP versions supported, it is capable to use applications that are capable to support all potential data formats received by the different supported DSP versions. Therefore, the asset data models supported in two different DSP versions are not related and therefore no compatibility requirement is needed between those.
```

looks a bit convoluted, could this be simplified? Like "changes between different protocol versions are expected"

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Approach
+
+- The compatibility tests are enriched by tests that ensure the detection of changes in the corresponding DSP assets.
```
ndr-brt, line 19: please explain the approach further

ndr-brt (commented):
format wise, I suggest to keep lines length around 120 chars, it will facilitate reading and reviewing

lgblaumeiser:
Hi @rafaelmag110 

> In my perspective there is no point in guaranteeing two distinct DSP versions are downwards compatible. To achieve this you would need to keep deprecated properties in newer versions, which you shouldn't need in the first place because you also support the previous (deprecated) version. In this sense, keeping up with deprecation is more simple because adopters can look for and remove deprecated DSP versions and not deprecated properties in recent versions.

I totally agree and actually thought I expressed that in the text. Backward compatibility is only within a DSP version and there are requirements coming from Catena-X about how long a version is valid within the data space. Whether this requirement is realistic or not is another discussion. Backward compatibility with a new DSP version is provided by supporting the old version as well, using the multi-version support of the EDC.

> Also, the requirement from CatenaX is that communication between two participants is not broken due to stack upgrades. If two participants communicate over a specific version of the DSP protocol, one of those participants can only upgrade its stack to a latter connector version IF this new version also supports the previous protocol version. I believe this is possible with the multi protocol support from the Connector. TractusX responsibility then is to define which DSP protocol versions have to be supported for each TractusX release.

Exactly, and we have to learn over time, how and potentially even how the requirement from Catena-X about the duration can be met.

> In terms of testing, in my perspective tractusx-edc should only guarantee that the protocol data model never changes for an already supported version. This is not really testing for downward compatibility, but for consistency.

Ok, for me this is downward compatibility between versions that support the same DSP protocol version concerning the implementation of the protocol version. That is downward compatibility on another layer compared to the compatibility between an EDC version that supports an older DSP version and an EDC supporting a newer DSP version but also providing means to communicate on the level of the older versions. Both are aspects of downward compatibility.

lgblaumeiser:
Hi @bmg13 ,

> > The requirement states that a consumer can continuously consume data from the provider without the need to be forced to update his system due to a change of the provider service stack.
> 
> As mentioned, the `Deprecation` exists for this reason. To ensure no imediate adoption of a new version is not needed but must considered and eventually changed until the definitve removal. So if a Provider bumps the connector version it would not affect the Consumer since deprecated values are supported.

There is a limitation to this specified by Catena-X, but in principle there is a requirement on the tools used in the Catena-X dataspace, I will add a link to that information.

> 
> > TractusX responsibility then is to define which DSP protocol versions have to be supported for each TractusX release.
> 
> This is one of the reasons the migration guides include potential breaking changes and why I believe "sustaining" version for multiple iterations "indefinatly" a dangerous path, since we should never live in a idea that late adoption of a new version is great. If consistency with all versions are to be kept indeterminably than no one would bump versions.

That is not the question, it is defined, each implementation of a DSP version must not have breaking changes over its lifetime. If something has to be changed in a breaking way, it must be aligned with a DSP version to keep the expected behavior distinguishable.

> 
> > This requires that such assets must not change in a way, that existing properties vanish from the returned data.
> 
> I do not understood that this was being considered and still think it wasn't. It was the consistency between supported versions and not a "vanishing" from a simple one-sided bump.

I think we all agree, that adding properties are not considered as breaking changes and a tool must be implemented in a flexible way, i.e., it should ignore fields it does not need for its functionality. My understanding of the current cases is, that something have been removed from the returned catalog which was there in the earlier version. That was, what I meant with vanishing.

> > If changes are detected a manual check that the downward compatibility is kept is executed.
> 
> IMO, is very hard to enforce any kind of manual check. The proposed tests should suffice, right?

The test methodology I propose makes sense in cases, where it is really hard to create a correct expected object to which you can compare the data object you receive from the test object. In JavaScript the objects returned from some libraries are extremely large and you would not want to maintain the "correct" object by hand. So what they do, they record one test run, assuming that the returned object is right. Afterwards, they compare the test object against the previously recorded test object. This way, you will get an issue, whenever the test object implementation changes in a way, that the data object changes in any way.

The question that needs to be answered is, whether the change is due to an intended change in the implementation or a regression done by accident. The manual check would involve to compare the changes of the previously recorded object from the currently returned object. The alternative is, that you define the minimum set of properties that you expect in the returned value and do a comparison that the data object returned from the test object is a superset of the minimum test data which is a valid approach but the comparison is a bit more difficult to implement.

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,38 @@
+# Handling of downward compatibility for a DSP version
+
+## Decision
+
+For a given DSP protocol version *N*, there is no restriction concerning backward compatibility towards the previous
+versions (*N-1*) implementation of the DSP protocol.
```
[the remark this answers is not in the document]
lgblaumeiser: Ok, my point was, that this is a clear statement towards the Tractus-X community, that they cannot do any assumption at that point and have to explicitely deal with the version information. This might be implied by DSP, but by explicitely stating that in the DR decision part from us Tractus-X EDC provider the implication from upstream cannot be misunderstood.

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+During the support of a certain DSP version *N*, the implementation has to be strictly backward compatible, i.e.,
+for two successive connector versions *X-1* and *X* that support both the DSP version *N*, if one participant changes
+from connector version *X-1* to version *X*, another dataspace participant using any compatible connector version and
+third party tools that use this connector can continue to transfer data to and from that participant without the need
+to do any changes in his setup.
```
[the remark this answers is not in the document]
lgblaumeiser: But that is actually the requirement we phase currently. There are data structures that are forwarded from another EDC to third-party apps and the issue is, that they can expect backward compatibility. That is, why I explicitely excluded additions to the data model as breaking change.

That two EDC versions are interoperable concerning a defined DSP version is one thing, but it must not happen, that someone operates his app with his connector and the other day, he cannot use it because the data he receives from another connector has changed in a way, that expected fields are missing.

If he fails because some fields have been added, we can consider that as a bug in the app implementation, but not if the data structure is changed in a way, that expected fields are gone. That is the whole reason for this decision record.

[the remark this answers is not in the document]
ndr-brt: Stating that there won't be any breaking change (eventually only additions) in a protocol version is more than enough, including that we will make third party app working as well will put us in a weird situation like the one happened in #1783 where the change was not breaking but that third party app had issue. 
We cannot guarantee that, in any case.

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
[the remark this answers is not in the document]
lgblaumeiser: I have adapted the rationale below. For me it is a part of the decision, you are right the DSP protocol works this way, but we could alternatively decide that we handle this with an layer in between which tries to cope with such breaking changes coming from the protocol implementation. As this makes no sense due to the reasons given in the rationale, we deliberately decide, that we forward the missing compatibility from upstream, as this can be handled with version awareness. This is our decision, inherited from upstream, but still our decision.

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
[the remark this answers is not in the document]
lgblaumeiser: I hope this new formulation makes the decision clearer

ndr-brt (approved):
I still think the first statement in the `Decision` can be removed as it is not a decision we're making, but, I don't want to block this anymore

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Decision
+
+For a given DSP protocol version *N*, there is no restriction concerning backward compatibility towards the previous
+versions (*N-1*) implementation of the DSP protocol.
+
+During the support of a certain DSP version *N*, the implementation has to be strictly backward compatible, i.e.,
+for two successive connector versions *X-1* and *X* that support both the DSP version *N*, if one participant changes
+from connector version *X-1* to version *X*, any interaction with the new connector version *X* has no breaking
+behavior changes. Additions to exchanged data models are in general not seen as breaking changes.
```
rafaelmag110, line 11:
```suggestion
from connector version *X-1* to version *X*, any interaction with connector version *X-1* is kept intact and has no breaking
behavior. Additions to exchanged data models are in general not seen as breaking changes.
```

docs/development/decision-records/2025-02-12-dsp-downward-compatibility/README.md
```diff
@@ -0,0 +1,41 @@
...
+[Operating model](https://catenax-ev.github.io/docs/operating-model/how-life-cycle-management). The requirement states
+that a dataspace participant can continuously exchange data with another particpant without the need to be forced to
+update his system due to a change of the other participants service stack. This requires that data sent between
+consumer and provider must not change in a breaking fashion.
+
+A backward compatibility contraint between DSP versions is neither needed nor possible over time, as
+
+1. a dataspace participant is able to control the connector version to use and thereby the DSP version offered in the
+   dataspace independently from other participants. The participant can choose a connector compatible to his
+   environment, so there is no need for a backward compatibility constraint.
```
rafaelmag110, line 25:
```suggestion
1. a dataspace participant should be able to control the connector version to use in his stack independently
 from other participants, as long as that version implements the DSP version(s) required by the dataspace 
 at that point. If all required DSP version(s) are supported, there is no need for a backward compatibility constraint. 
 Communication is still possible through previous DSP version(s).
```

rafaelmag110 (approved):
This is in a much better shape than the initial proposal, imo.
Left just a couple of nits, but approve the overall content.

Issue #1743, opened 2025-01-14
Get rid of the dependency check


ndr-brt:
## WHAT

...and provide it before the release

## WHY
In [TRG 7-04](https://github.com/eclipse-tractusx/eclipse-tractusx.github.io/blob/3c416db1293acd006f8a5e909b7a06764fcfeeea/docs/release/trg-7/trg-7-01.md#dependencies-file) is stated that it's not necessary to keep the DEPENDENCIES in the repository and have it checked on every commit.
The check can happen once right before the release, and the file can be published on github pages for reference.

This needs to be done also on other repos under the edc subdomain: bdrs, compatibility tests and so on...

## HOW

- [x] add dependencies check right before the release
- [x] commit the DEPENDENCIES file during the release
- [x] generate and publish the DEPENDENCIES file to github pages on every commit on main branch
- [x] change NOTICE file to point to the github pages published DEPENDENCIES file
Bonus:
- [ ] add automatic job that nightly (but also on every commit!) request IPlab review for restricted/rejected dependencies (so the check before release will likely pass all the time)

## FURTHER NOTES

inspiration can be taken from the upstream EDC process:
- [dependencies check during release preparation (notice the "strict" flag)](https://github.com/eclipse-edc/.github/blob/b1129698357874b3973201ff449941c1a19c9be1/.github/workflows/core-prepare-release.yml#L31)
-  [generate dependencies file action](https://github.com/eclipse-edc/.github/blob/b1129698357874b3973201ff449941c1a19c9be1/.github/actions/generate-dependencies-file/action.yml)
- [publish dependencies file to github pages on every push to main](https://github.com/eclipse-edc/.github/blob/b1129698357874b3973201ff449941c1a19c9be1/.github/workflows/publish-dependencies.yml)

bmg13:
hey everyone, can I be assigned this, please?

ndr-brt:
@bmg13 there's still a little task to do: get rid of the DEPENDENCIES file from the repository :)

bmg13:
> [@bmg13](https://github.com/bmg13) there's still a little task to do: get rid of the DEPENDENCIES file from the repository :)
> 
> EDIT: already opened a PR: [#1866](https://github.com/eclipse-tractusx/tractusx-edc/pull/1866)

Thank you, @ndr-brt !

pratapipatelbcone:
[...]

PR #1818, opened 2025-02-27
feat: Include dependency check before release

bmg13:
## WHAT

Perform the dependency check before release in main branch instead on all branches in PR targetting main and publish the DEPENDENCIES file to GitHub pages.

## WHY

Considering this dependency validation can happen only before release, this is an overhead removed in PRs and additionally provides the dependencies information in GH Pages as well.

## FURTHER NOTES

Some reasoning behind the decisions proposed:
- The creation, validation and publishing of dependencies information is done on three situations: every night, on each new commit in main and by workflow call. This promotes having a valid DEPENDENCIES file, i.e. with no rejected or restricted dependencies, by the time a release is to be done.
- The preexisting [run-dash action in sig-infra](https://github.com/eclipse-tractusx/sig-infra/blob/main/.github/actions/run-dash/action.yaml) is called to promote reusability of this check on other TX repos as well.

Closes #1817


ndr-brt (changes requested):
This PR is doing too many things, please stick with the plan reported in the issue #1743 . Following the PR title the only changes needed here are the ones to add the check before the release, no publish on gh-pages, no replacement in the NOTICE file.
The process could sound slow but it will permit to have consistent and reasonable contributions instead of a big one that does thing not really needed (e.g. the "schedule" thing).

bmg13:
> This PR is doing too many things, please stick with the plan reported in the issue #1743 . Following the PR title the only changes needed here are the ones to add the check before the release, no publish on gh-pages, no replacement in the NOTICE file. The process could sound slow but it will permit to have consistent and reasonable contributions instead of a big one that does thing not really needed (e.g. the "schedule" thing).

This makes sense, will break this PR in smaller ones with specific scopes.

PR #1831, opened 2025-03-10
feat: Add workflow to check dependencies before release.

bmg13:
## WHAT

Perform the dependency check before release in main branch (instead on all branches in PR targetting main) and commits the DEPENDENCIES file.

## WHY

Considering this dependency validation can happen only before release, this is an overhead removed in PRs.

## FURTHER NOTES

The preexisting [run-dash action in sig-infra](https://github.com/eclipse-tractusx/sig-infra/blob/main/.github/actions/run-dash/action.yaml) is called to promote reusability of this check on other TX repos as well.

For more information please check this PR:https://github.com/eclipse-tractusx/tractusx-edc/pull/1818.

Relates to https://github.com/eclipse-tractusx/tractusx-edc/issues/1743


.github/workflows/check-dependencies.yaml
```diff
@@ -39,24 +36,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-java
```
rafaelmag110, line 37: If the order on the other file is non important, we can remove the `setup-java` action here.

.github/workflows/check-dependencies.yaml
```diff
@@ -18,16 +18,13 @@
...
 
 
-name: "3rd Party dependency check (Eclipse Dash)"
+name: "Generate and check DEPENDENCIES"
 
 on:
   workflow_dispatch:
   push:
     branches:
       - main
```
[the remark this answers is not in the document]
ndr-brt: I would keep it for the time being until we actually push the dependencies file on github pages and we make it available for consultation

.github/actions/generate-and-check-dependencies/action.yml
```diff
@@ -0,0 +1,59 @@
...
+inputs:
+  run:
+    description: How the dependencies check must be applied. The "standard" will fail only for 'rejected' dependencies, "strict" will fail also for 'restricted' ones.
+    required: false
+    default: standard
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v4
```
ndr-brt, line 33: generally I would not put checkout in a composite action because it could lead to unexpected behavior (like overriding changes made by other actions), it's always better to have it called on the root workflow

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
rafaelmag110: We don't have any benefit of having it active since we won't be looking into the result, and it will probably be just laying around failing on every push until we decide to fix it or do a release...

We can always run the check manually if we want to have a preview of how the deps are.

We can also run this in the weekly build instead of on every push...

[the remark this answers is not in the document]
ndr-brt: well, the one that merges the PR will get a notification, that could be read before reaching the release day and it would make somebody aware of the existence, e.g. of a rejected dependencies (worst case tho).

@bmg13 In any case maybe for the moment I would not remove the triggering of this on `pull_request` until we have the file published on github. It's maybe nitpicking, but I would prefer having the actual desired process (file published and available, check active on release) before removing the current one, so this PR should only be about the check on release, and not removing the check for PRs

.github/workflows/draft-release.yaml
```diff
@@ -109,7 +111,7 @@ jobs:
       - name: Commit manifest files
         id: make-commit
         run: |
-          git add gradle.properties $(find charts -name Chart.yaml) $(find charts -name README.md)
+          git add DEPENDENCIES gradle.properties $(find charts -name Chart.yaml) $(find charts -name README.md)
```
ndr-brt, line 114: why don't replace it with `git add .`? Any modification to local files should be done for release purpose, and adding everything will avoid to have anybody to think about adding an eventual new file (like `DEPENDENCIES`)

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
lgblaumeiser: I would suggest to not use strict in here, as we do not want to break the build in case of restricted dependencies on this frequent basis. Then the build only breaks in cases, where we have to react, like rejected dependencies.

.github/workflows/draft-release.yaml
```diff
@@ -84,6 +84,8 @@ jobs:
         run: |
           git config user.name "eclipse-tractusx-bot"
           git config user.email "tractusx-bot@eclipse.org"
+      - name: Check dependencies before release
```
lgblaumeiser, line 87: Actually, in case of a release, running the check in strict mode is important, as we actually should not release a version with restricted dependencies.

lgblaumeiser (changes requested):

.github/workflows/draft-release.yaml
```diff
@@ -84,6 +84,8 @@ jobs:
...
+        uses: ./.github/workflows/check-dependencies.yaml
       - uses: ./.github/actions/setup-java
```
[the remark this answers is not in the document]
bmg13: Good point, done before in draft-release and removed in other usages. Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/7d314d64101efbf822bdf9b977d6ab8821fd779d).

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
bmg13: Bringing this comment also to this thread and I will try and explain my reasoning.
> Actually, in case of a release, running the check in strict mode is important, as we actually should not release a version with restricted dependencies.

I believe that the strict should be used both on release prepare and on any run of the action. If we wait only to a release to find restricted dependencies and if they exist, than the run will fail and that can be prevented. I think finding out the restricted dependencies as soon as possible is the best approach.

> @bmg13 In any case maybe for the moment I would not remove the triggering of this on pull_request until we have the file published on github. 

I agree, rollbacked in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/7d314d64101efbf822bdf9b977d6ab8821fd779d) to run on pr as well. In next PR's related with this issue this may be revisited.

> My opinion is we should remove this listener from here. This means the CI check will fail on main most of the time because we are not actively updating the file. It's not necessary.

Will keep it for now to, as suggested, focus this pr on check the deps before release and leave the rest (triggers per example) as they are.

.github/workflows/check-dependencies.yaml
bmg13: Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/7d314d64101efbf822bdf9b977d6ab8821fd779d).

.github/actions/generate-and-check-dependencies/action.yml
bmg13: Good point, thank you. Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/7d314d64101efbf822bdf9b977d6ab8821fd779d).

.github/workflows/draft-release.yaml
bmg13: It makes sense to include everything. Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/7d314d64101efbf822bdf9b977d6ab8821fd779d).

.github/workflows/draft-release.yaml
bmg13: Replied in here.

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
lgblaumeiser: Hi @bmg13, the point with restricted dependencies is, that this is a temporary status which basically means, that the compliance check of the Eclipse Foundation has not been finished so far. This can change a few days later. The point with release check is, that when we want to release, we might not be able to wait these days, but in normal development, there is no reason to not continue.

Actually, we get real trouble with dependabot warnings this way, as the proposal of the dependabot pr typically leads to a restricted dependency, as this updated version is simply so new that it has not been checked. But since the previous version typically was accepted, the risk of getting the new version rejected is minimal.

.github/workflows/draft-release.yaml
lgblaumeiser: Ok, @bmg13 , in the answer referenced, you state that it should run in strict mode for release preparation, but I do not see that you set the parameter to strict, but stay on the default value, which is not strict, right?

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
ndr-brt: I'd also avoid `strict` on every commit on main branch because otherwise the build will be 90% broken, it's good to have the possibility what is the status of the dependencies on every commit but that should not stop the development cycle.

The goal is to have:
- on every commit the DEPENDENCIES built and published
- on release a proper strict check

.github/workflows/check-dependencies.yaml
```diff
@@ -37,26 +37,7 @@ jobs:
...
 
-      - name: generate dependency list
-        run: |
-          ./gradlew allDependencies | grep -Poh "(?<=\s)[\w.-]+:[\w.-]+:[^:\s\[\]]+" | sort | uniq > dependency-list
-          cat dependency-list
-
-      - name: Run dash
-        id: run-dash
-        uses: eclipse-tractusx/sig-infra/.github/actions/run-dash@main
+      - uses: .github/actions/generate-and-check-dependencies
```
ndr-brt, line 41: please note that you removed `setup-java` both from the `generate-and-check-dependencies` action and here as well, so this workflow will fail

.github/workflows/draft-release.yaml
```diff
@@ -85,6 +85,8 @@ jobs:
           git config user.name "eclipse-tractusx-bot"
           git config user.email "tractusx-bot@eclipse.org"
       - uses: ./.github/actions/setup-java
```
ndr-brt, line 87: same here, setup-java needs to be called somewhere. I'd let it in the root workflow like here

.github/workflows/draft-release.yaml
ndr-brt: yes, this call should be `strict`

ndr-brt (changes requested):

.github/workflows/draft-release.yaml
bmg13: Yeah, you are right, will update it.

.github/workflows/check-dependencies.yaml
bmg13: Updated to ensure the setup-java is done before calling the action in here (left removed in the action).

.github/workflows/check-dependencies.yaml
bmg13: You're right. Rollbacked in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/a0343f885b79364f8d09844e34e58a0e7e8874d7) to include it in the `check-dependencies` workflow. So, now, each workflow will certainly setup it up before calling the `generate-and-check-dependencies`. Left on both workflows since they are independent from each other and will run independently.

.github/workflows/draft-release.yaml
bmg13: Updated in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/a0343f885b79364f8d09844e34e58a0e7e8874d7).

.github/workflows/draft-release.yaml
bmg13: Applied the strict enforcement on draft release only in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/a0343f885b79364f8d09844e34e58a0e7e8874d7).

.github/workflows/check-dependencies.yaml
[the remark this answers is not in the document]
bmg13: Updated in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1831/commits/a0343f885b79364f8d09844e34e58a0e7e8874d7) so that only failing on restricted if is a draft release workflow call. On other scenarios left the default (`standard`) use.

lgblaumeiser (approved):
lgtm

ndr-brt (approved):
LGTM

PR #1840, opened 2025-03-14
Add dataplane proxy public api extension module

### LINKED ISSUE #1780: HTTP Proxy implementation
## WHAT

The current HTTP proxy implementation upstream has some shortcomings. Major issue is, that depending on the situation, masking a 4xx error coming from the backend by translating it into a 500 status code can be the right thing to do as it indicates a configuration error on the provider side, or wrong, as the status code was motivated by a user error. Currently, in the proxy masks all 4xx status codes to 500.

Based on a request to upstream, a discussion arose which will lead to a decision of the upstream committer group soon. Based on this decision, downstream activities will be needed to keep the functionality of a HTTP proxy with improved status code handling.

## WHY

To keep the HTTP proxy functionality stable although upstream changes will occur.

## HOW

The downstream implementation should be based on the upstream code base.

farhin23:
## WHAT

Adds `dataplane-public-api-v2` in `dataplane` extensions.

## WHY

As the `data-plane-public-api-v2` will be deprecated in the upstream, this PR adds an extension for this proxy implementation.

## FURTHER NOTES

Closes #1780


ndr-brt (approved):
looks good to me, please fix the dependencies

bmg13 (approved):

PR #1863, opened 2025-03-25
feat: publish dependencies to GitHub pages

bmg13:
## WHAT

Publishes the new DEPEDENCIES file to the GitHub pages.

## WHY

To keep deps information up to date, once the DEPENDENCIES file is updated the new version should be published in GH pages.

## FURTHER NOTES

The change regarding PR to main triggering workflow is done now since it avoids publishing on new PR's and since is a change to be done related with the issue, was made here.

A following PR to update the NOTICE file will be proposed, was not included here to keep the focus on GH pages.

Relates to https://github.com/eclipse-tractusx/tractusx-edc/issues/1743

### DIFF
--- a/.github/workflows/generate-and-publish-dependencies.yaml
+++ b/.github/workflows/generate-and-publish-dependencies.yaml
@@ -18,7 +18,7 @@
 #################################################################################
 
 
-name: "Generate and check DEPENDENCIES"
+name: "Generate DEPENDENCIES file and publish to GitHub pages"
 
 on:
   workflow_dispatch:
@@ -39,3 +39,14 @@ jobs:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup-java
       - uses: ./.github/actions/generate-and-check-dependencies
+      - name: Prepare to publish
+        shell: bash
+        run: |
+          mkdir public
+          cp DEPENDENCIES public/
+      - name: Publish to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: public
+          keep_files: true
--- a/.github/workflows/generate-and-publish-dependencies.yaml
+++ b/.github/workflows/generate-and-publish-dependencies.yaml
@@ -25,9 +25,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    branches:
-      - main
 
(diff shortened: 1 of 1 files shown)


ndr-brt (approved):

lgblaumeiser (changes requested):
Hi @bmg13 , just that I get it, you proposal is to cancel the ongoing checks and only do the check on release preparation and deliberately by explicitely starting this workflow. I have to admit, I would like to see a report on pull requests, just to be notified pretty early, when we use rejected dependencies. That could be awkward when we detect them only when we want to do a release.

I would propose to store the file after a successful check during the release potentially in the release preparation and keep this workflow as it is.

bmg13:
> Hi @bmg13 , just that I get it, you proposal is to cancel the ongoing checks and only do the check on release preparation and deliberately by explicitely starting this workflow. I have to admit, I would like to see a report on pull requests, just to be notified pretty early, when we use rejected dependencies. That could be awkward when we detect them only when we want to do a release.
> 
> I would propose to store the file after a successful check during the release potentially in the release preparation and keep this workflow as it is.

hey @lgblaumeiser. This workflow is triggered by any commit to main, so any pr (to main) after being merged will trigger this. So, the dependencies check and the publishing to github pages is done more frequently than just when preparing a release. You are right, the check in the PR is removed, however not having this check on every commit is the expected behavior of the [issue](https://github.com/eclipse-tractusx/tractusx-edc/issues/1743).

lgblaumeiser:
> > Hi @bmg13 , just that I get it, you proposal is to cancel the ongoing checks and only do the check on release preparation and deliberately by explicitely starting this workflow. I have to admit, I would like to see a report on pull requests, just to be notified pretty early, when we use rejected dependencies. That could be awkward when we detect them only when we want to do a release.
> > I would propose to store the file after a successful check during the release potentially in the release preparation and keep this workflow as it is.
> 
> hey @lgblaumeiser. This workflow is triggered by any commit to main, so any pr (to main) after being merged will trigger this. So, the dependencies check and the publishing to github pages is done more frequently than just when preparing a release. You are right, though, the check in the PR is removed, however not having this check on every commit is the expected behavior of the [issue](https://github.com/eclipse-tractusx/tractusx-edc/issues/1743).

HI @bmg13 sorry, I oversaw the push to main trigger only saw the removal of the pr trigger. I am fine with this then, a misunderstanding from my side.

PR #1883, opened 2025-04-03
refactor(test): avoid duplicate postgres/in-memory e2e testing

### LINKED ISSUE #1879: proposal: avoid double in-memory / postgresql e2e testing
## WHAT

While looking at the EndToEnd tests, I noticed that for a lot of tests we are doing double setting to permit them to run both in-memory and with postgres.

But since our artifacts are mainly postgresql-backed solutions, I was wondering why are we double-testing these features.

The proposal is to avoid this double persistence testing and go with just PG. 

## WHY
Most of the code we're covering there is already tested upstream, while in the Postgres testing we're verifying that the runtime is working with the db migrations we're providing.

## HOW

My proposal would be to:
- switch all the tests that are using the test runtime `runtime-memory" to use the `runtime-postgresql`
- delete the `runtime-memory` runtime
- switch the `@PostgresqlIntegrationTest` annotations in the `edc-tests` folder to `@EndToEnd`, this will also speed up tests execution (that is currently dependent by the "huge" `postgres-tests` job, that's taking 8-10 min approx.

## FURTHER NOTES

I will came up with a DR to outline the proposal and to find an agreement

ndr-brt:
## WHAT

Avoid duplicated in-memory/postgresql e2e testing by choosing the latter.
Get rid of the `runtime-memory` runtime

## WHY

Test efficacy

## FURTHER NOTES

- the iatp tests are still using runtimes that are in memory, I can foresee a situation where they get integrated in the other e2e tests, for the moment just left them as they are because an eventual switch would have been out of scope
- I discovered a bug on the `provision-azure` module ([already reported upstream](https://github.com/eclipse-edc/Technology-Azure/issues/338)), curious that nobody spotted it already here, it's part of the runtime since 0.8.0...). I will open an issue in this repo as well, the test execution in the meantime has been mitigated with a workaround.
- time to run `Run-All-Tests` passed from approx 10-11 mins to 4-5

Closes #1879


edc-tests/e2e/policy-tests/src/test/java/org/eclipse/tractusx/edc/tests/policy/PolicyDefinitionEndToEndTest.java
```diff
@@ -48,38 +46,44 @@
...
+    private static final RuntimeExtension PROVIDER_RUNTIME = pgRuntime(PROVIDER, POSTGRES);
+
+    @DisplayName("Policy is accepted")
+    @ParameterizedTest(name = "{1}")
+    @ArgumentsSource(ValidContractPolicyProvider.class)
+    void shouldAcceptValidPolicyDefinitions(JsonObject policy, String description) {
+        PROVIDER.createPolicyDefinition(policy);
+    }
 
+    @DisplayName("Policy is accepted")
```
bmg13, line 82: nit: I know is pre-exisitng, but this display name does not seem okay and could be updated.

bmg13 (approved):
this is awesome!
just left a nit

lgblaumeiser (commented):
lgtm, I think, there are quite some DisplayNames who need some polishing, but that is something out of scope for this PR

lgblaumeiser (approved):
I meant to approve, sorry

ndr-brt:
generally speaking, `@Display` are really easy to get outdated, I'd avoid those unless really necessary

Issue #1881, opened 2025-04-02
Version hasn't been bumped after latest release


ndr-brt:
## Describe the bug

Version in `gradle.properties` it's still `0.9.0`, instead of `0.10.0-SNAPSHOT`

ndr-brt:
This happened because the `publish` job failed for unknown reasons, but packages got published on maven central anyway.
https://github.com/eclipse-tractusx/tractusx-edc/actions/runs/13792692823

So the current "mitigation" solution has been to extract a manually executable `bump-version` workflow, but the long term solution in my opinion would be better to separate the release from main branch, in a Trunk Based approach, like:

- prepare the release by creating the release branch, already bump the version on main
- do the release on the release branch.

that's it. we're using the same approach on EDC upstream project quite successfully.

PR #1886, opened 2025-04-04
build: extract bump-version workflow

ndr-brt:
## WHAT

Extracts `bump-version` workflow, that can be used to bump version manually in case the publish on maven central failed (but the artifacts actually got published)

## WHY

Bump the version to the next snapshot :)

## FURTHER NOTES

inlined the `publish-tag-and-release` as it was used once

Part of #1881


lgblaumeiser (approved):
lgtm

PR #1907, opened 2025-04-16
refactor(test): improve IATP tests assertions

ndr-brt:
## WHAT

I just discovered that the IATP tests didn't test the interaction with the catalog correctly, they were hitting the federated catalog instead of the /catalog endpoint on management api, that was responding 404 (`not(200)` was ok).

Now they make more sense

## WHY

## FURTHER NOTES


rafaelmag110 (approved):
Looks good.
Wouldn't it be better to be more specific in the assert? `not(200)` is valid in a lot of cases and as it was the case here the test was green but the real behavior was failing...

rafaelmag110:
@ndr-brt adjust checkstyle and we're ready to go

Issue #1898, opened 2025-04-08
Allow retrieval of all BPN groups through specific endpoint


bmg13:
## WHAT

Update the BPN Group API to allow a user to directly retrieve all the BPN groups.

## WHY

Currently, to list all the BPN Groups available, it is needed to retrieve all the BPNs and iterate over the groups they are assigned to. This is a very consuming approach and exposing an endpoint returning all the groups would simplify client usability.

## HOW

Query the ```edc_business_partner_group``` table and expose a new GET endpoint that returns all the BPN groups in the ```BusinessPartnerGroupApiV3```.

Before implementation, a **Decision Record must be done** explaining the approach to be followed. Once it is merged, an implementation can be PR.

## FURTHER NOTES

[This](https://github.com/eclipse-tractusx/tractusx-edc/pull/1867) Decision Record can be checked due to similarity of flow.

Issue #1806, opened 2025-02-20
Implementation of a BPN Group Member Get-Method


MarcMaron:
## WHAT

This ticket aims to provide a straightforward method for accessing the members of a BPN Group directly.

## WHY

Currently, the EDC allows for the definition of BPN Groups and checking the different group memberships of a single BPN.
However, this process is cumbersome for the inverse use-case: checking which BPNs are included in a BPN-Group.
At present, one would have to scour every BPN to determine which members are part of which group.
Implementing a method to access group membership details would resolve this issue.

## HOW

We propose the implementation of a simple GET-method to access the members of a BPN-Group.
This method would need a BPN-Group ID as input and would provide all BPNLs of the members it entails.

lgblaumeiser:
First make a proposal as DR to make the changes more concrete

lgblaumeiser:
Hi, @MaximilianHauer @MarcMaron , we discussed and assigned this item to Rafael. Please check, if you want to stay assigned, as this is for us someone,working actively on the issue. Thanks!

MaximilianHauer:
thx for the info lars we will remove us from the assignment but if we can contribute give us a sign

PR #1908, opened 2025-04-17
docs: Create DR to retrieve all BPN Groups

bmg13:
## WHAT

Adds a Decision Record to expose all the BPN groups in a new endpoint.

## WHY

Right now to get all BPN groups stored the user needs to retrieve all BPNs and their related groups and iterate over them. This can be time consuming and adding a new endpoint that returns all the BPN groups would simplify API usage.

Relates to #1898

### DIFF
--- a/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
+++ b/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
@@ -9,6 +9,8 @@ TractusX-EDC will expose an endpoint that will provide all the BPN groups with t
 Currently, listing all available BPN Groups requires [retrieving all BPNs](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.9.0/#/Business%20Partner%20Group/resolveV3) in the `BusinessPartnerGroup` api and iterating through their assigned groups.
 This approach is consuming for the user and providing an endpoint to return all BPN groups with respective BPN assigned would improve usability since a single request can be performed. Using the existing BPN Group API keeps consistency.
 
+Since the response of this new endpoint includes the same information provided from the `/group/{group}` GET request and this one was not published on any release, it will be removed.
+
 ## Approach
 
 1. In `BusinessPartnerGroupApiV3` create a new POST endpoint, appending `/request`. This returns filtered BPNs (according to a particular query sent as a QuerySpec in the body of the request) each containing the list of groups it is assigned to.
@@ -35,4 +37,5 @@ The endpoint will accept a request body with the `querySpec` containing the filt
   }
 ]
 ```
+5. Delete the `/group/{group}` GET endpoint from the `BusinessPartnerGroupApiV3` interface and remove the implementation from `BusinessPartnerGroupApiControllerV3`.
 
--- a/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
+++ b/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
@@ -13,60 +13,26 @@ This approach is consuming for the user and providing an endpoint to return all 
 
 1. In `BusinessPartnerGroupApiV3` create a new POST endpoint, appending `/request`. This returns filtered BPNs (according to a particular query sent as a QuerySpec in the body of the request) each containing the list of groups it is assigned to.
 
-The endpoint will accept a request body with the `querySpec` like the following.
+The endpoint will accept a request body with the `querySpec` containing the filtering BPN's. If no BPN is sent as filter, the response should be all BPNs and their assigned groups.
 
-```json
-{
-  "filterExpression": [
-    {
-      "operandLeft": "BusinessPartnerNumber",
-      "operator": "IN",
-      "operandRight": ["BPN000000001", "BPN000000002", "BPN000000003"]
-    }
-  ],
-  "limit": 50,
-  "offset": 0,
-  "sortOrder": "ASC",
-  "sortField": "BusinessPartnerNumber"
-}
-```
-
-If the `operandRight` is empty, the response should be all BPNs and their assigned groups.
-
-The response for the above request will be similar to the next block.
+2. Create a query targeting the `edc_business_partner_group` table to evaluate and extract all BPNs and respective groups.
+3. Using the `BusinessPartnerStore` include a new operation that returns the BPNs and BPN groups and the change is reflected in the two existing implementations (In Memory and SQL).
+4. The response will include a list of filtered BPNs and their assigned groups and should be similar to the next block.
 
 ```json
 [
-   {
-      "@id": "BPN000000001",
-      "tx:groups": [
-         "group4",
-         "group3",
-         "group5",
-         "group2",
-         "group1"
-      ]
-   },
-   {
-      "@id": "BPN000000002",
-      "tx:groups": [
-         "group1",
-         "group2",
-         "group5"
-      ]
-   },
-   {
-      "@id": "BPN000000003",
-      "tx:groups": [
-         "group1"
-      ]
-   }
+  {
+    "bpn": "BPNL000000000001",
+    "groups": ["group1","group5","group4"]
+  },
+  {
+    "bpn": "BPNL000000000002",
+    "groups": ["group1","group5","group4"]
+  },
+  {
+    "bpn": "BPNL000000000003",
+    "groups": ["group1"]
+  }
 ]
 ```
-   
-  - The `query` object should contain the type of object to be queried and the criteria for filtering.
-  - The response should include a list of BPNs and their assigned groups.
-2. Create a query targeting the `edc_business_partner_group` table to evaluate and extract all BPNs and respective groups.
-3. Using the `BusinessPartnerStore` include a new operation that returns the BPNs and BPN groups and the change is reflected in the two existing implementations (In Memory and SQL).
-4. Response of the new endpoint is a key-value map with the key being the BPNs that meet the query constraints and the value containing the corresponding groups.
 
--- a/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
+++ b/docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
@@ -2,16 +2,71 @@
 
 ## Decision
 
-TractusX-EDC will expose an endpoint that will provide all the BPN groups that have a BPN assigned to them.
+TractusX-EDC will expose an endpoint that will provide all the BPN groups with the BPN assigned to them.
 
 ## Rationale
 
-Currently, listing all available BPN Groups requires [retrieving all BPNs](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.9.0/#/Business%20Partner%20Group/resolveV3) in the `BusinessPartnerGroup` api and iterating through their assigned groups. This approach is consuming for the user and providing an endpoint to return all BPN groups would improve usability. Using the existing BPN Group API keeps consistency.
+Currently, listing all available BPN Groups requires [retrieving all BPNs](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.9.0/#/Business%20Partner%20Group/resolveV3) in the `BusinessPartnerGroup` api and iterating through their assigned groups.
+This approach is consuming for the user and providing an endpoint to return all BPN groups with respective BPN assigned would improve usability since a single request can be performed. Using the existing BPN Group API keeps consistency.
 
 ## Approach
 
-1. In `BusinessPartnerGroupApiV3` create a new POST endpoint, appending `/groups/request`, that returns all groups according to a particular query (sent as a QuerySpec in the body of the request).
-2. Create a query targeting the `edc_business_partner_group` table to evaluate and extract all BPN groups.
-3. Using the `BusinessPartnerStore` include a new operation that returns the BPN groups and the change is reflected in the two existing implementations (In Memory and SQL).
-4. Response of the new endpoint is a list containing all the groups that meet the query constraints.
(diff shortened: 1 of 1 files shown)


docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Rationale
+
+Currently, listing all available BPN Groups requires [retrieving all BPNs](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.9.0/#/Business%20Partner%20Group/resolveV3) in the `BusinessPartnerGroup` api and iterating through their assigned groups. This approach is consuming for the user and providing an endpoint to return all BPN groups would improve usability.
+
+Creating a new specific endpoint that returns all BPN groups provides a simplified approach. Using the existing BPN Group API keeps consistency.
+
+## Approach
+
+1. In `BusinessPartnerGroupApiV3` create a new GET endpoint that returns all groups, appending `/groups`.
```
rafaelmag110, line 15: I'm not a fan of this approach. 
For other DMA endpoints, the EDC has established a pattern to query a certain store, which also allows other properties of the `QuerySpec` to be changed. We should apply this pattern here too to maintain consistency.

Also, the response message can be a simple list with the groups. No need for a "complex" JSON response.

rafaelmag110 (changes requested):
Left some comments inline.

docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
```diff
@@ -0,0 +1,32 @@
+# Expose all Business Partner Groups
+
+## Decision
+
+TractusX-EDC will expose an endpoint that will provide all the BPN groups.
```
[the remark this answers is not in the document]
bmg13: Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1908/commits/3a6204f31d110613cef5a2b9089bfef59a173c1b).

docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
bmg13: Yeah, makes more sense, even updating the path to include "/request". Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1908/commits/3a6204f31d110613cef5a2b9089bfef59a173c1b).

rafaelmag110:
Well, I gave this topic a second though and I believe we are heading in the wrong path here, as we are pretty much creating a new `groups` API over the `business-partner-groups` api. One API has the BPNL as the central entity and the other as the Group, and that is a distinction we do not want to have.

The first mistake we made was introducing the `group/{bpn}` endpoint , as that IMO directly responds with business logic that in this case should be derived differently. The endpoint we should have introduced is the ` POST business-partner-groups/request` which should return a list of bpn group entries as they are actually persisted ( as in the list of all bpn entries and their corresponding groups).

```
[
    {
        "@id": "BPN000001234",
        "tx:groups": "group1,group3"
    },
    {
        "@id": "BPN000005678",
        "tx:groups": "group1,group2"
    }
]
```

Then, if a client requires a list of all BPNs that a _specific_ group contains, it should filter by group using a `QuerySpec`.

```
# Considering the previous example, when using a QuerySpec containing a filter by group2
# the API should return

[
    {
        "@id": "BPN000005678",
        "tx:groups": "group1,group2"
    }
]
```

BUT the actual list of BPNs (as in, the list of strings), should then be resolved by the client and not directly returned by the API.

With the `POST /request` endpoint, _returning the existing distinct groups_ (what this PR is proposing) shouldn't be required, as that response is computed by the client too.

@lgblaumeiser @ndr-brt What is your view on this?

bmg13:
Uncertained if we should delete since was just added or to deprecate it.

bmg13:
> Hi @bmg13 I would approve in general, but shouldn't we document also that we revert the previous DR introducing the /groups/ endpoint?

@lgblaumeiser makes sense
Uncertained if we should delete since was just added or to deprecate it. I would deprecate if this new endpoint is not in time of next release

lgblaumeiser:
> > Hi @bmg13 I would approve in general, but shouldn't we document also that we revert the previous DR introducing the /groups/ endpoint?
> 
> @lgblaumeiser makes sense Uncertained if we should delete since was just added or to deprecate it. I would deprecate if this new endpoint is not in time of next release

Is the endpoint in the 0.9.0 release? If so, we have to deprecate it, but I have not found it in th openapi spec, so we can basically remove it imho. Anyway, we should document what will happen with the old endpoint in the DR.

bmg13:
> > > Hi @bmg13 I would approve in general, but shouldn't we document also that we revert the previous DR introducing the /groups/ endpoint?
> > 
> > 
> > @lgblaumeiser makes sense Uncertained if we should delete since was just added or to deprecate it. I would deprecate if this new endpoint is not in time of next release
> 
> Is the endpoint in the 0.9.0 release? If so, we have to deprecate it, but I have not found it in th openapi spec, so we can basically remove it imho. Anyway, we should document what will happen with the old endpoint in the DR.

Yeah, it was not added to any release, so I think it can be deleted, will update the DR. Thanks!

lgblaumeiser (approved):
lgtm

docs/development/decision-records/2025-04-17-expose-all-bpn-groups/README.md
```diff
@@ -0,0 +1,41 @@
+# Expose all Business Partner Groups
+
+## Decision
+
+TractusX-EDC will expose an endpoint that will provide all the BPN groups with the BPN assigned to them.
+
+## Rationale
+
+Currently, listing all available BPN Groups requires [retrieving all BPNs](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.9.0/#/Business%20Partner%20Group/resolveV3) in the `BusinessPartnerGroup` api and iterating through their assigned groups.
```
gerbigf, line 9: I'm still not satisfied with the proposal. In the "rationale" we say that listing all groups requires iterating through the assigned groups of BPNs. With the new /request/ endpoint we'll do exactly that (again). 

The user has no easy way of getting the "which BPNs are in group X" information. Just as previously, he would need to iterate over all JSON objects from the /request response and in each object iterate over the array of "groups". That might work for 15 BPNs and seven groups, but it will surely not work for 1000+ BPNs (which we should design the solution for). 

Furthermore, the user still can't "create a new group", "add member to group", "delete all members from a group". So I would question if we want to fix something that's broken per design or if we want to come up with a new version /v4/business-partner-group that solves those issues by introducing a group centric view.

gerbigf: Exactly. We learned that both our onboarding teams as well as individual users would like to manage group just the way you described it. They specify the group that they want to release the asset for (access policy) and afterwards they would want to manage the group with exactly the operations you described.

gerbigf: If we would redesign the API, we could also think about the current behaviour, that one can accidentally but very easily remove all group assignments with a simple PUT: 

Consider that a BPNL has three groups assigned: [A,B,C] - the user now wants to add group D. With the current implementation he first needs to do a GET on the BPNL group assignment, take the array and add D to it and then PUT the new array [A,B,C,D]. We've already had cases were users thought it's enough to PUT the array with [D] - leading to the loss of all current group assignments. I see this as an easy source of error that's hard to recover from as you can't "undo" the operation that easily. 

Being able to add/remove members from a group would be a much safer approach. If you try to add a member to the group that's already existing, an error could be thrown. If the user accidentally removes a member, all other member stay untouched...

bmg13: Thanks for those, I will try to give my 2 cents here as well.

1. Regarding the "/request" endpoint, I do not find any objection from any side in adding it, so we can keep the idea of having it in the v3. This request keeps consistency with similar entities and allows to have a more "predictable" api.
2. I understood a redesign of the outcome of the request in this DR, with BPNGroup being the identifier instead of the BPN. This can be achieved logically, no problem, my doubt is if it is in line with what the API now represents, since is more of BPN as id.
3. The usage of this request may be the point of major discord, since removing the recently added one for this would require logic on the client side. I think it is okay (as @rafaelmag110 explained above) and within an acceptable api usage. The part that is particular (at least to me) is that the BPN-BPNGroup relationship is the "real" entity here which is unusual (when compared with other entities like Asset, Policy, etc.) resulting in a trickier way of exposing the data.
4. I agree with redesigning the API if concrete requirements are presented that can justify that change. The current v3 api allows to achieve any outcome we discussed, however the usability of the api is not always great, specially from a BPNGroup-first view. Example of this is the potential accidental de-assignment @gerbigf mentioned, the need to delete a BPNGroup being so cumbersome, the listing of all BPNGroup's is not straightforward, etc.

That said, I would suggest middle ground here:
- Have the request with BPNGroup as id as shown in @lgblaumeiser comment, as opposed to what now is proposed in this pr. Can't say I am a fan, considering the current BPN first api view, but can understand;
- Delete the recently added endpoint (listing the BPNs in a BPN Group). If not accepted to be deleted, at least remove the "/group" path for consistency;
- Common acceptance of a possible new v4 (or new API) that tackles outlined requirements with a more BPNGroup-first view if needed.

gerbigf: I could go with your proposal @bmg13! This way we could ensure the initial request from https://github.com/eclipse-tractusx/tractusx-edc/issues/1806 is still fulfilled. Remember that we wanted to give users access to a) all BPNLs in a group and b) all groups.

A redesign of the BPN Group API would require more design work and could come with 25.09 earliest anyways.

lgblaumeiser: OK, two more cents from my side 😄 

1. Question: concerning the /request output, should we stick to the current proposal or the one I gave above? The current proposal is more in alignment with the semantics of the api, but it only adds value if I filter for exactly one group. If not, it adds the computation overhead you described. My proposal would fulfill the requirement get all groups better, imho.
2. What you describe with PUT is a major drawback of this function. But in EDC PUT was used consistently, so it does not make sense to introduce PATCH for now, which I like more then PUT. Nonetheless, it is a question of the user interface, as the REST PUT semantics is one thing, passing it directly to the user to edit the resource is another thing, right?

gerbigf: I'd go with fulfilling the requirements that were 
- List all BPNs in a specific group 
- List al groups that there are 

first. If we can do that with the /request endpoint that allows filtering for a group, fine. If the implementation in the back will be a bit hacky, I think we can also live with that. 

We can then take time to redesign the BPN Group feature and discuss if we want to improve usability at the cost of deviating from the "standard" upstream API. I think we have good reasons for that: The Group to BPNL assignment is a different type of business object than an asset or a policy. It's something that is meant to be updated and changed frequently and in different ways than an asset is changed. One doesn't frequently add or remove properties from an asset as part of a business process...

ndr-brt: Sorry for not joining the party earlier :')
I see this as a view/endpoints issue, not modeling one.

I don't see any issues in adding "view" endpoints that provides way to get the information in a way for which it addresses business need.
I would not go down in the rabbit hole of re-modeling this bpn groups domain part just because we cannot represent the endpoint in a "pure" REST. 
Let's try to be pragmatic.

Following up @gerbigf comment:
> List all BPNs in a specific group

The `POST /request` endpoint or also a simple `GET /?group=<name>` will make it

> List all groups that there are

we already have the `GET /group/<group>` endpoints (it should have been `/group**s**/<group> but... :smiling_face_with_tear: ), we can add a `GET /group` or `GET /groups` that will return all the groups


About the issue of being able to add a new group without having to do a full update of the BPN (I don't think this is related to these requirements but I felt it like a desirable feature):
`PUT /business-partner-groups/<BPN>/groups/<group name>` to add it. 
`DELETE /business-partner-groups/<BPN>/groups/<group name>` to delete it.

let's keep a complete refactoring for a moment in which there will be enough requirements to make it mandatory, at the moment I don't see the value for it.

rafaelmag110: Well, the rabbit hole for me is adding "views" to the API. We have different perspectives here.
And not adding specific business requirements to an API is something that I learned from you @ndr-brt.

To all the involved in the discussion, I think we all had time to understand and reflect on these requirements. Let's get to a conclusion tomorrow on our weekly so we can get this PR to closure.

ndr-brt: I mean, it's a matter of finding the best compromise here right? 

I'm not in favor of adding too much view logic because every piece of logic is a piece of code, and every piece of code needs to be maintained, and some uses cases could be too domain specific to be maintained into the connector.
Clients should be able to do simple data manipulation like browsing BPNs and aggregate the information, that's a matter of fact. 

But, if to address a simple requirement we are discussing about completely re-modeling the domain, considering that adding the `POST /request` and the "not adherent" `GET /groups` endpoint will do the job, well, that makes the choice really easy to take.

bmg13:
Discussed today during the Tractus-X Weekly meeting and aligned to revert to original proposal. Future work to update the API will need specific requirements and further discussion.
Change was added in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/1908/commits/36ab305e565911f0aa945ef39f1b0332bab8380d).

gerbigf (approved):
lgtm!

rafaelmag110 (approved):

lgblaumeiser (approved):
lgtm

ndr-brt (approved):

PR #1915, opened 2025-04-22
fix(config) - added missing opentelemetry options

### LINKED ISSUE #1905: Metrics endpoint not exposed
# Bug Report

## Describe the Bug

In my Prometheus I want to be able to scrape the metrics from the EDC. However, if I have the metrics endpoint enabled in my values.yaml as well as the opentelemetry agent, the deployed EDC does not expose the metrics endpoint (anymore?).
Connecting to the pods and running `netstat -tulpen` does not show the 9090 connection.

Image

The pod definition contains the metrics endpoint:
```
        - name: metrics
          containerPort: 9090
          protocol: TCP
```

`opentelemetry.properties` in the pod are also correct:

Image

### Expected Behavior

Proper exposure of the metrics endpoint to scrape the metrics with Prometheus.
(EDC deployed with helm chart)

### Observed Behavior

The metrics endpoint is missing. However, the configuration is correct.

## Steps to Reproduce

1. Deploy EDC with metrics endpoint and otel java agent enabled
2. Connect to the pod
3. Run `netstat -tulpen` or try to curl/wget the metrics endpoint
5. Optionally verify the configuration of the pod to verify if everything is correct 

## Context Information

- Used version: 0.9.0

kkotowiczz:
## WHAT
remapped port from default 9464 to 9090, and overriden exporter format

## WHY
In order to be able to scrap metrics with prometheus, when otel javaagent is enabled

## FURTHER NOTES
N/A

Closes #1905

### DIFF
--- a/charts/tractusx-connector/values.yaml
+++ b/charts/tractusx-connector/values.yaml
@@ -334,7 +334,7 @@ controlplane:
   opentelemetry: |-
     otel.javaagent.enabled=false
     otel.javaagent.debug=false
-    # By default prometheus uses port 9464
+    # Alignment of edc /metrics port and OTEL prometheus configuration (By default OTEL is exposing prometheus metrics at 9464)
     otel.exporter.prometheus.port=9090
     otel.metrics.exporter=prometheus
 
@@ -579,7 +579,7 @@ dataplane:
   opentelemetry: |-
     otel.javaagent.enabled=false
     otel.javaagent.debug=false
-    # By default prometheus uses port 9464
+    # Alignment of edc /metrics port and OTEL prometheus configuration (By default OTEL is exposing prometheus metrics at 9464)
     otel.exporter.prometheus.port=9090
     otel.metrics.exporter=prometheus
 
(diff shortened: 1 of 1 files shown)


charts/tractusx-connector/values.yaml
```diff
@@ -333,6 +334,9 @@ controlplane:
   opentelemetry: |-
     otel.javaagent.enabled=false
     otel.javaagent.debug=false
+    # By default prometheus uses port 9464
```
lgblaumeiser, line 337: One slight proposal on this comment and the same below, @kkotowiczz. I do not understand the point. Is it, that this config is only needed, because otel and prometheus are not aligned concerning port? Please adapt the comment in a way, that the issue better. Thanks!

lgblaumeiser (changes requested):

charts/tractusx-connector/values.yaml
ndr-brt: If it's just a configuration issue why is this change needed to happen here? tbh I don't see why we should have a particular exporter configured by default when that should be done by the user (in fact, the java agent is disabled by default, see line 335)

ndr-brt (commented):
Discussion on the PR aside, the PR should come with an automated way to test it, or a reasonable motivation if that's not possible

charts/tractusx-connector/values.yaml
ndr-brt: Maybe we can just provide documentation to tell users how they can have the `/metrics` endpoint exposed

kkotowiczz: Closing PR then

ndr-brt:
@kkotowiczz the documentation could have been done in this PR :)

Issue #1926, opened 2025-04-29
Prepare 0.10.0 release candidate for 25/06 Tractus-X release


lgblaumeiser:
- Take an upstream snapshot 0.13.0-Snapshot
- Finish the tasks necessary for the release
- Release a rc1 on May, 5th latest

ndr-brt:
https://github.com/eclipse-tractusx/tractusx-edc/releases/tag/0.10.0-rc1

PR #1940, opened 2025-05-05
refactor: avoid deferenced var may be null warning

ndr-brt:
## WHAT

Avoid security scanning warning for potential deferenced var may be null.
(no need to add coverage as the case was already covered, but refactored in a better way so codeql won't complain anymore)
Ref. https://github.com/eclipse-tractusx/tractusx-edc/security/code-scanning/603

## WHY

## FURTHER NOTES

Part of #1926


lgblaumeiser (approved):
lgtm

Issue #1952, opened 2025-05-08
End-to-end tests in the `Verify` workflow are not always executing.


AndrYurk:
## Describe the bug

When End-to-End tests are run by the `Verify` workflow, they do not always execute because the Gradle develocity cache is enabled. However, these tests must run to generate the Allure report.

### To Reproduce

Manually trigger the `Verify` workflow and observe that the End-to-End tests are not executed.

### Expected behavior

End-to-End tests should execute every time the `Verify` workflow runs.

### Screenshots/Error Messages

![Image](https://github.com/user-attachments/assets/3fde82ab-8647-4143-8253-040eb510ab46)

## Possible Implementation

As a potential solution, run the tests with the `--no-build-cache` amd `--rerun-tasks` options.

PR #1953, opened 2025-05-08
fix: avoid using gradle cache for end-to-end tests execution

AndrYurk:
## WHAT

This PR avoid using gradle cache when executing end-to-end tests in the `Verify` workflow

## WHY

To ensure that end-to-end tests always run and allow the generation of an accurate Allure report.

Part of #1952

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -130,8 +130,7 @@ jobs:
       - name: Run E2E tests (${{ matrix.dir }})
         run: |
           ./gradlew compileJava compileTestJava
-          ./gradlew -p edc-tests/e2e/${{ matrix.dir }} test -DincludeTags="EndToEndTest" -PverboseTest=true \
-          --rerun-tasks --no-build-cache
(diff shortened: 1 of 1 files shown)


.github/workflows/verify.yaml
```diff
@@ -130,7 +130,8 @@ jobs:
       - name: Run E2E tests (${{ matrix.dir }})
         run: |
           ./gradlew compileJava compileTestJava
-          ./gradlew -p edc-tests/e2e/${{ matrix.dir }} test -DincludeTags="EndToEndTest" -PverboseTest=true
+          ./gradlew -p edc-tests/e2e/${{ matrix.dir }} test -DincludeTags="EndToEndTest" -PverboseTest=true \
+          --rerun-tasks --no-build-cache
```
ndr-brt, line 134: I think we can just use one of those 2, `--no-build-cache` is pretty explanatory and in this case for our needs it is sufficient.

format nit: if we want to use `\` to go newline, the newline should have at least 2 leading whitespaces, so will be clear at a glance that's continuation of the previous line and not a different command

AndrYurk: Corrected

PR #1954, opened 2025-05-08
feat: make release workflow manual

### LINKED ISSUE #1652: The release workflow should be manual

rafaelmag110:
## WHAT

Makes the release workflow available for manual trigger _only_.

## WHY

Implementation of DR that was forgotten.

Implements #1666 
Closes #1652


rafaelmag110:
@ndr-brt @lgblaumeiser 
I removed the fail-on: error input because of the issue with the analyze job since I though it was the problem. But turns out that is actually doing nothing at all, so I kept its removal.

.github/workflows/release.yml
```diff
@@ -21,23 +21,18 @@
...
 
 
 jobs:
   # Gate
   validation:
-    name: "Check if repository is not fork AND head is release OR base is bugfix"
+    name: "Check if repository is not fork AND ( base is bugfix/* OR main )"
     runs-on: ubuntu-latest
-    if: ${{ github.repository == 'eclipse-tractusx/tractusx-edc' && (startsWith(github.ref_name, 'bugfix/') || startsWith(github.event.pull_request.head.ref, 'release/')) }}
+    if: ${{ github.repository == 'eclipse-tractusx/tractusx-edc' && (startsWith(github.ref_name, 'bugfix/') || git.ref_name == 'main') }}
```
ndr-brt, line 35: release flow should not be triggerable from main, as we want to pass through "draft release" first

rafaelmag110: I adapted this according to our current workflow, and in that we are releasing from main - after the PR that's created from the draft is merged.

If we want to change the process after, then it's another thing...

ndr-brt: If we're ok with setting up an halfway solution - that hopefully will be solved later - I'd also add a check that ensures that the version we are trying to release is not a "-SNAPSHOT", so we will force the draft process at least.

Issue #1962, opened 2025-05-12
Catalog Request: The provided AccessToken could not be validated


nicoprow:
## Describe the bug

When performing a catalog request between two EDCs (0.10.0-rc1) we receive a 500 internal server error response.

The logs inside the requesting EDC's control plane show a bad request error is being thrown somewhere during the process of BDRS communication (BdrsClientImpl) which in turn seems to be the cause of the 500 returned to the requestor. The error logs are further down.

### To Reproduce

_Steps to reproduce the behavior:_

1. Deploy an EDC 0.10.0-rc1 with configuration analogous to [here](https://github.com/SujitMBRDI/product-bpdm-env/blob/main/int/edc-test-sharing-member/values.yaml)
2. Perform a v3 catalog request to another EDC 0.10.0-rc1
3. See the 500 error

### Expected behavior

I expect the catalog request to return an empty catalog response.

### Screenshots/Error Messages

```
SEVERE 2025-05-08T10:35:14.708459881 JerseyExtension: Unexpected exception caught
org.eclipse.edc.spi.EdcException: Presentation Query failed: HTTP 400, message: {"timestamp":"Thu, 08 May 2025 10:35:14 GMT","status":400,"message":"Presentation Query: The provided AccessToken could not be validated.","error":"BAD_REQUEST","path":"/api/v1.0.0/iatp/presentations/query"}
	at org.eclipse.tractusx.edc.identity.mapper.BdrsClientImpl.lambda$resolve$0(BdrsClientImpl.java:105)
	at org.eclipse.edc.spi.result.AbstractResult.orElseThrow(AbstractResult.java:119)
	at org.eclipse.tractusx.edc.identity.mapper.BdrsClientImpl.resolve(BdrsClientImpl.java:105)
	at org.eclipse.tractusx.edc.identity.mapper.BdrsClientAudienceMapper.resolve(BdrsClientAudienceMapper.java:42)
	at org.eclipse.edc.protocol.dsp.http.dispatcher.DspHttpRemoteMessageDispatcherImpl.dispatch(DspHttpRemoteMessageDispatcherImpl.java:116)
	at org.eclipse.edc.runtime.core.message.RemoteMessageDispatcherRegistryImpl.dispatch(RemoteMessageDispatcherRegistryImpl.java:48)
	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogServiceImpl.requestCatalog(CatalogServiceImpl.java:45)
[...]

PR #1964, opened 2025-05-13
refactor(test): use actual bdrs-server in iatp e2e tests

ndr-brt:
## WHAT

uses an actual `bdrs-server-memory` instance in the `iatp` tests
to do this, other than a `BdrsExtension` I added a `DidServerExtension`, that takes care to hold and expose DIDs for all the participants (issuer included).

**NOTE**: this PR does not fix the bug, but it demonstrated that the communication with BDRS works, and likely the problem is in the interaction with the credential service.

## WHY

cover the BDRS interaction

## FURTHER NOTES
- use `implementation` for referencing internal modules to trigger automatic compilation on e2e test execution
- catch potential exceptions thrown by the BDRS resolution

Part of #1962


edc-extensions/bdrs-client/src/test/java/org/eclipse/tractusx/edc/identity/mapper/BdrsClientImplComponentTest.java
```diff
@@ -94,8 +94,9 @@ class BdrsClientImplComponentTest {
     private static final String ISSUER_DID = "did:web:%s:%s".formatted(NGINX_CONTAINER_NAME, ISSUER_NAME);
     private static final String HOLDER_DID = "did:web:%s:%s".formatted(NGINX_CONTAINER_NAME, HOLDER_NAME);
     private static final Network DOCKER_NETWORK = Network.newNetwork();
+
     @Container
-    private static final GenericContainer<?> BDRS_SERVER_CONTAINER = new GenericContainer<>("tractusx/bdrs-server-memory")
+    private static final GenericContainer<?> BDRS_SERVER_CONTAINER = new GenericContainer<>("tractusx/bdrs-server-memory:0.5.4")
```
lgblaumeiser, line 99: How can we ensure this hard coded reference gets updated?

edc-tests/e2e/iatp-tests/src/test/java/org/eclipse/tractusx/edc/tests/transfer/extension/BdrsServerExtension.java
```diff
@@ -0,0 +1,90 @@
...
+import static io.restassured.http.ContentType.JSON;
+import static java.util.Map.entry;
+import static org.eclipse.edc.util.io.Ports.getFreePort;
+
+public class BdrsServerExtension implements BeforeAllCallback, AfterAllCallback {
+
+    private final LazySupplier<URI> directoryEndpoint = new LazySupplier<>(() -> URI.create("http://localhost:%d%s".formatted(getFreePort(), "/directory")));
+    private final LazySupplier<URI> managementEndpoint = new LazySupplier<>(() -> URI.create("http://localhost:%d%s".formatted(getFreePort(), "/management")));
+
+    private final GenericContainer<?> bdrsServer = new GenericContainer<>("tractusx/bdrs-server-memory:0.5.4")
```
lgblaumeiser, line 44: Same here, how do we ensure the updates when new service versions are created. With now two occurences, this is even more critical.

lgblaumeiser (changes requested):
Small issue, but we need a safety net for the bdrs service version, so that the tests get updated with the current version in future as well.

edc-extensions/bdrs-client/src/test/java/org/eclipse/tractusx/edc/identity/mapper/BdrsClientImplComponentTest.java
ndr-brt: at the moment we don't have a way to have the testcontainers used images updated, the same is happening for postgres. I'd address this topic in a different issue because the solutions could be different, and just have this merged in the meantime

lgblaumeiser (approved):
lgtm

PR #1937, opened 2025-05-05
refactor: cleanup dependencies

ndr-brt:
## WHAT

leverage on upstream BOM modules to build our runtimes instead of manually adding every module needed

## WHY

reduce upgrade efforts, keep the build up-to-date easily

## FURTHER NOTES

- I cleaned up some tests in the [previous PR](https://github.com/eclipse-tractusx/tractusx-edc/pull/1930) but when I pushed the changes the PR was already merged. Small changes that makes them more readable and less likely to randomly fails (less statically instantiated ports)
- discovered a significant design flaw, described in this discussion: https://github.com/eclipse-tractusx/tractusx-edc/discussions/1938

Part of #1926

PR #1968, opened 2025-05-15
fix: exclude sts client extension from controlplane-base

ndr-brt:
## WHAT

Exclude `org.eclipse.edc:identity-trust-sts-remote-client` from the controlplane-base module.

## WHY

Fix bug

### Why this happened?
Using upstream BOM files ([PR](https://github.com/eclipse-tractusx/tractusx-edc/pull/1937)) we also inherited that module, that as shown in logs we received from users, it was overriding 2 services:
```
WARNING 2025-05-13T01:58:53.310725006 A service of the type org.eclipse.edc.iam.identitytrust.sts.remote.StsRemoteClientConfiguration was already registered and has now been replaced with a StsRemoteClientConfiguration instance.
WARNING 2025-05-13T01:58:53.316061488 A service of the type org.eclipse.edc.iam.identitytrust.spi.SecureTokenService was already registered and has now been replaced with a RemoteSecureTokenService instance.
```

### Why we didn't catch that before?
Unfortunately in some e2e runtimes we were actually doing exclusions instead of relying on the pure production runtimes, that's why this PR is actually moving the exclusion up to the base runtime definition 

### How to prevent that in the future?
E2E tests should be as similar as possible to the production environment, trying to avoid mocks, stubs, dummies and fakes.

Closes #1968

### DIFF
--- a/edc-controlplane/edc-controlplane-base/build.gradle.kts
+++ b/edc-controlplane/edc-controlplane-base/build.gradle.kts
@@ -26,6 +26,9 @@ plugins {
 configurations.all {
     // edr-cache-api excluded due to edr controller signature clash with tx-edr-api-v2 that provides same functionality with token auto_refresh capability
     exclude(group = "org.eclipse.edc", module = "edr-cache-api")
+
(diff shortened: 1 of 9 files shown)


lgblaumeiser (approved):
lgtm

Issue #1833, opened 2025-03-11
Improve release process to allow the release of a certain git commit (commit sha, tag)


lgblaumeiser:
## WHAT

Adopt the release pipelines in a way, that allows to start a release from a given git commit by providing, e.g., the commit sha or a tag that reflects the commit.

## WHY

In the Tractus-X release process that includes a test phase, the way of releasing is by providing a release candidate that is used during the testing and to "promote" the last release candidate to be the release if testing was successful. Since development at that point already progressed, a mechanism is needed to start the release not from the current HEAD but from a defined commit, e.g., the tag of the release candidate. This should be enabled by this feature.

## HOW

The idea is, that the release build gets a parameter that can be set when the build is started. The parameter reflects the checkout parameter for the git checkout command that is used to get the source code version that is to be released. The parameter defaults to HEAD, so that if the build is started without editing the parameter or automatically, the standard behavior is still executed.

### Outputs
- Decision Record reflecting the decision
- Changed release action

AndrYurk:
Hello, dear colleagues @lgblaumeiser, @ndr-brt, @rafaelmag110, @bmg13.

I’d like to share my thoughts on a possible solution for our release process, hope you will find it useful.
Below is an overview of the current release process (non-essential steps have been omitted):
![Image](https://github.com/user-attachments/assets/a7d5d81d-daee-46f3-be4d-a26b5317a15f)

Here is my proposal for improvement:
![Image](https://github.com/user-attachments/assets/64970c09-976d-4273-9a20-2ee227de3976)

Benefits that I see of this approach:
- The `release/*` branch will always reflect the exact code of the created release.
- All tests will run on the pull request, giving us full control over test execution.
- There will be a clear linkage between passed tests and the corresponding release.
- Harmonization with the hotfix release process.

Please feel free to share your feedback!

ndr-brt:
@AndrYurk thanks for the overview.

I think we can avoid the "merge back on main" at all. It's really not necessary and it only adds complexity.
We are currently following a sort of customized git-flow, without having any necessity.
By adopting a [Trunk Based approach](https://trunkbaseddevelopment.com/) this issue will be solved automatically by having 2 simple manual workflows:
- prepare -> starting from main or from a release branch:
  - it creates a new release/hotfix branch, prepare versions and bump version in the trunk (`main`)
- release: starting from a release/hotfix branch
  - it launches tests, then publishes artifacts, create tag and creates github release.

[...]

PR #1997, opened 2025-05-25
docs:  DR for release process changes

AndrYurk:
## WHAT

This PR introduces a decision record for the new release process

## WHY

To discuss and document the decision

Part of #1833

### DIFF
--- a/docs/development/decision-records/2025-05-26-release-process/README.md
+++ b/docs/development/decision-records/2025-05-26-release-process/README.md
@@ -1,0 +1,46 @@
+# Release Process Changes
+
+## Decision
+
+To streamline our release process, we will split it into two manually triggered workflows:
+### Prepare for Release ("Draft Release" workflow):
+- **Trigger:** Can be started from a specific commit SHA, the `HEAD` of `main`, the `HEAD` of any branch or tag.
+- **Actions:**
+  1. Create a new `release/*` or `hotfix/*` branch.
+  2. Update dependencies in both the release/hotfix branch and in `main`.
+  3. On the release/hotfix branch:
+     - Bump the Gradle project version and the Helm chart version based on the workflow’s input parameters.
+  4. On the `main` branch:
+      - For a real release, bump the Gradle version to the next `-SNAPSHOT`.
+      - For release candidates (RC) or hotfixes, leave the Gradle version unchanged.
+      - Update the Helm chart version according to the workflow’s input parameters.
+### Execute Release ("Release" workflow):
+- **Trigger:** Can be started from the `HEAD` of a `release/*`, `hotfix/*` branch.
+- **Actions:**
+    1) Run automated tests.
+    2) Publish artifacts (Maven, Docker, Helm).
+    3) Create the Git tag for this release.
+    4) Generate a GitHub Release entry.
+    5) Publish the OpenAPI UI spec to GitHub Pages.
+    6) Update release notes with a link to the Allure test-report.
+
+## Rationale
+
+- **Flexibility:** We can now prepare a release branch from any commit SHA, not just `main`.
+- **Clarity:** All tests and changes for a given release live in the release branch, making them easy to find.
+- **Simplicity:** Eliminates unnecessary pull requests against `main` and run tests prematurely during the GitHub Release.
+
+
+## Approach 
+
+1. **`draft-release.yaml`**
+    - Update according to the “Prepare for Release” workflow steps.
+
+2. **`release.yaml`**
+    - Update according to the “Execute Release” workflow steps.
+
+3. **New Action:** `update-version-and-charts`
+    - Automates version bumps in `build.gradle` (or `gradle.properties`) and `Chart.yaml`.
+
+4. **`verify.yaml`**
+    - Extend to generate the Allure report for any `release/*` or `hotfix/*` branch.
--- a/docs/development/decision-records/2025-05-26-release-process/README.md
+++ b/docs/development/decision-records/2025-05-26-release-process/README.md
@@ -40,7 +40,7 @@ To streamline our release process, we will split it into two manually triggered 
     - Update according to the “Execute Release” workflow steps.
 
 3. **New Action:** `update-version-and-charts`
-    - Automates version bumps in `build.gradle` (or `gradle.properties`) and `Chart.yaml`.
+    - Automates version update in `gradle.properties` and `Chart.yaml`
 
 4. **`verify.yaml`**
     - Extend to generate the Allure report for any `release/*` or `hotfix/*` branch.


docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+    2. Update DEPENDENCIES file in both the release/hotfix branch and in `main`.
+    3. On the release/hotfix branch:
+        - Bump the project version in `gradle.properties` and the Helm chart version based on the workflow’s input parameter.
+    4. On the `main` branch:
+        - For a real release, bump the project version in `gradle.properties` to the next `-SNAPSHOT`.
+        - For release candidates (RC) or hotfixes, leave version in `gradle.properties` unchanged.
+        - Update the Helm chart version according to the workflow’s input parameter.
+
+2. **`release.yaml`**
+- **Trigger:** Can be started from the `HEAD` of a `release/*`, `hotfix/*` branch.
```
bmg13, line 36:
```suggestion
- **Trigger:** Can be started from the `HEAD` of a `release/*` or `hotfix/*` branch.
```

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+- **Actions:**
+    1) Run automated tests.
+    2) Publish artifacts (Maven, Docker, Helm).
+    3) Create the Git tag for this release.
+    4) Generate a GitHub Release entry.
+    5) Publish the OpenAPI UI spec to GitHub Pages.
+    6) Update release notes with a link to the Allure test-report.
+
+3. **New Action:** `update-version-and-charts`
```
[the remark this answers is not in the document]
AndrYurk: Yes, `bump-version.yaml` will be part of it, but runs only for the `main`

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+## Approach 
+
+1. **`draft-release.yaml`**
+- **Trigger:** Can be started from a specific commit SHA, the `HEAD` of `main`, the `HEAD` of any branch or tag.
+- **Inputs:**
+    1) commit SHA (if empty, will use `HEAD`)
+    2) release version (required)
+- **Actions:**
+    1. Create a new `release/*` or `hotfix/*` branch.
+    2. Update DEPENDENCIES file in both the release/hotfix branch and in `main`.
```
ndr-brt, line 27: in `main` there should be no `DEPENDENCIES` file as we're publishing it on github pages.
on preparation we need to run the dash check in strict mode, so if it fails, we cannot release

rafaelmag110:
@AndrYurk thanks for this. Some points from my side.

- When you trigger a release preparation with the draft-release, how do we decide if it is an RC or an official release? How is the version bump decided? Does it actually make sense to differentiate?
- Once you branch off from main, say to create an RC, what happens to the development flow on main? Between the release of an RC and an official release, I'm guessing fixes required to an RC branch are done on main and then cherry picked to a new release branch based off of the first one, right?
- What is the requirement for releasing from any commit SHA then? Releasing from the HEAD from either release or bugfix branches should be enough.
- Why is the version required as input? In my opinion this is error prone as we are introducing human input.

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+    3. On the release/hotfix branch:
+        - Bump the project version in `gradle.properties` and the Helm chart version based on the workflow’s input parameter.
+    4. On the `main` branch:
+        - For a real release, bump the project version in `gradle.properties` to the next `-SNAPSHOT`.
```
lgblaumeiser, line 31: Shouldn't this be done in the `release.yml` file, as the release process could be stopped. In that case, we do not want the version to be changed in main, right?

docs/development/decision-records/2025-05-26-release-process/README.md
lgblaumeiser: What is the consequence of a failing dependency check in strict mode. Strict mode adds only the check for restricted dependencies, i.e., temporary not curated dependencies. Imho, it is a question of time, so we need a mechanism to trigger the dependency check again on the relevant branch.

For some reason, we have a DEPENDENCIES file again in the main branch, this seems to be a bug in the current release process. I agree, that we should not provide a DEPENDENCIES file in the main branch.

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Approach 
+
+1. **`draft-release.yaml`**
+- **Trigger:** Can be started from a specific commit SHA, the `HEAD` of `main`, the `HEAD` of any branch or tag.
+- **Inputs:**
+    1) commit SHA (if empty, will use `HEAD`)
+    2) release version (required)
+- **Actions:**
+    1. Create a new `release/*` or `hotfix/*` branch.
```
lgblaumeiser, line 26: Is there a need to separate between release and hotfix branches. I know we do this historically, but basically, it is a release, right. It starts from a different commit/branch/tag but I do not see the need of a different branch naming to distinguish a release 0.11.0 from 0.10.1, the version number and the commit history imply the type anyway.

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+        - Bump the project version in `gradle.properties` and the Helm chart version based on the workflow’s input parameter.
+    4. On the `main` branch:
+        - For a real release, bump the project version in `gradle.properties` to the next `-SNAPSHOT`.
+        - For release candidates (RC) or hotfixes, leave version in `gradle.properties` unchanged.
+        - Update the Helm chart version according to the workflow’s input parameter.
+
+2. **`release.yml`**
+- **Trigger:** Can be started from the `HEAD` of a `release/*` or `hotfix/*` branch.
+- **Actions:**
+    1) Run automated tests.
```
lgblaumeiser, line 38: Shouldn't this be done in draft-release? My suggestion is, that after the draft-release process we are in a situation that we can decide whether we can release or not, i.e., all checks are through and green. If not, we have to decide how to deal with the situation.

The test report should be created in this step, that is important. Question is, whether it could be build out of information from the draft-release step.

lgblaumeiser (commented):
Some discussion points added

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: I think we need to change the version in `main` immediately after creating the official release branch, as it will bring us a clear commit history and structure. The official release process will not be stopped, as it should be done after the successful RC release.

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: The key distinction is that we make commits in `hotfix/*` branches but not in `release/*` branches, and it serves as the identifier for marking a release as “latest.”

AndrYurk:
> @AndrYurk thanks for this. Some points from my side.
> 
> * When you trigger a release preparation with the draft-release, how do we decide if it is an RC or an official release? How is the version bump decided? Does it actually make sense to differentiate?

We’ll draft successive RCs until they meet our needs, otherwise, we’ll create an official release. The version in main is only bumped when running draft-release from main without the -RC suffix.

> * Once you branch off from main, say to create an RC, what happens to the development flow on main? Between the release of an RC and an official release, I'm guessing fixes required to an RC branch are done on main and then cherry picked to a new release branch based off of the first one, right?

I believe the simplest approach is to commit changes to main and then create a new RC. Cherry-picking should be reserved for hotfixes.

> * What is the requirement for releasing from any commit SHA then? Releasing from the HEAD from either release or bugfix branches should be enough.

The commit SHA input is included only in the `draft-release` workflow, it lets us select the exact point in main from which to create a release branch. Including a commit SHA makes no sense in the `release` workflow, since it always runs from the HEAD of the `release/*` or `hotfix/*` branches.

> * Why is the version required as input? In my opinion this is error prone as we are introducing human input.

This enables us to identify whether a branch corresponds to an RC, an official release, or a hotfix.

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: That makes sense to me, but it also makes sense to run tests immediately before the release, since there is a gap between `draft-release` and `release` (during which additional commits could be made), to avoid releasing a version without passing tests. The report could be built out of information from the draft-release step.

docs/development/decision-records/2025-05-26-release-process/README.md
lgblaumeiser: No, what I mean is, that we start releasing lets say 0.10.0. But then we find out, that there is an issue with the release, e.g., the tests fail or the dependency check fails for some reason. I.e., we have to fix something and start with the 0.10.0 release process again. WIth the current approach, we do not now, if we work on 0.10.0 or 0.11.0 after the draft release process, only when we trigger the release action, it is clear that we go for the release. That is my point.

docs/development/decision-records/2025-05-26-release-process/README.md
lgblaumeiser: That is the question, coming back to my issue on stopping a release process, the ability to also do a quick fix to a main release branch could make sense, as, e.g., if it is a real minor issue, we could just fix it and go for a second test.

I do not get your remark concerning "latest", which release would you say is the latest then?

docs/development/decision-records/2025-05-26-release-process/README.md
lgblaumeiser: Right, if we develop on the release branch (e.g., in the hotfix case) we need a mechanism, that runs the tests again and again. My point is, that the decision to press the release button should be based on a successful test run. So, it does not hurt to run the tests again, but it must not be the case, that these tests fail (except our random failures of course), because they succeeded prior to the decision to finalize the release.

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: I thought we would first create a 0.10.0-RC1 (which would not update the version in main). If that satisfies our needs, we would then create an official release from the same commit SHA. Now, I’m not sure—do we still want to use *-RC releases?

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: Corrected

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: Corrected

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,49 @@
...
+- **Actions:**
+    1) Create a new `release/*` branch.
+    2) Run `generate-and-check-dependencies` action in strict mode.
+    3) Run automated tests.  
+    4) On the release branch:
+        - Bump the project version in `gradle.properties` and the Helm chart version based on the workflow’s input parameter.
+    5) On the `main` branch:
+        - For the official release, bump the project version in `gradle.properties` to the next `-SNAPSHOT`.
+        - For release candidates (RC) or hotfixes, leave version in `gradle.properties` unchanged.
+        - Update the Helm chart version according to the workflow’s input parameter.
```
lgblaumeiser, line 34: What does this mean? Isn't that the same in helm charts as will gradle.properties?

lgblaumeiser: Ah, and this means, that the helm charts are set to what in the main branch?

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Decision
+
+To harmonize the release process, all releases will be based solely on `release/*` branches. We will no longer create or 
+use separate `bugfix/*` branches for release purposes, nor will we cut releases directly from `main`. In other words, 
+there will be no distinction between bugfix and release branches — everything intended for a release must reside in a 
+`release/*` branch. 
+
+Additionally, draft releases can now be triggered from a specific commit SHA, and the version type will be selected from 
+predefined options (`RC` as the default, `bugfix`, or `official release`) instead of entering a free-text version.
```
rafaelmag110, line 11: I see why you wrote this statement, but do we really need the ability to release from a specific commit SHA?
RCs will be drafted from main and prepared in their own branch. Once the RC is flagged for official release, we can simply draft the official release from the latest RC branch. In all cases, we are always drafting from the HEAD of any branch.

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Rationale
+
+- **Flexibility:** We can now prepare a release branch from commit SHA, not just `HEAD`.
+- **Cohesion:** Every release — whether a bugfix or a official version — will start from a `release/*` branch, providing 
+   a single branching pattern for all release work.
+- **Simplicity:** The draft releases workflow will now use predefined version-type options (`RC`, `bugfix`, or `official release`) 
+   instead of a free-text entry, minimizing confusion and input errors.
```
rafaelmag110, line 19: Still don't think the way we are describing the value added into the simplicity aspect is fair. The better option might be to just remove the line.

Another thing we must not leave out is the single reason why we though about doing this in the first place: We want a way to promote RCs into official releases. That should somehow be part of the rationale.

docs/development/decision-records/2025-05-26-release-process/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+## Approach 
+
+Branch flows:
```
rafaelmag110, line 24: Within the diagram, please consider that with the changed flow we should not release the official release from main but from an RC branch, that is where the release preparation happened and where the "release management tested" version is.

rafaelmag110 (changes requested):
We are getting there. A couple more points.

docs/development/decision-records/2025-05-26-release-process/README.md
AndrYurk: Corrected

rafaelmag110 (approved):
Thanks for the changes.

PR #2014, opened 2025-06-02
feat: Include edc.dcp.v08.forced property as true

### LINKED ISSUE #1982: DCP 1.0 functionality introduced by accident
## Describe the bug

The connector 0.10.0 requires a new config parameter:

edc.dcp.v08.forced=true

This indicates, that the 0.10.0 connector has the capability of another version of DCP which is set as default one. This is most likely a prototype for DCP 1.0. DCP 1.0 should not be part of the connector in version 0.10.0 as this is used in an environment, that uses the existing protocol version 0.8 only.

### To Reproduce

n/a

### Expected behavior

DCP 1.0 functionality is not part of the connector distribution

Alternative: No configuration is needed by end users to make the connector use 0.8 as standard.

bmg13:
## WHAT

Sets the `edc.dcp.v08.forced` property value as `true` in the charts.

## WHY

Since the 0.8 [is not the default version](https://github.com/eclipse-edc/Connector/blob/ae835abbcae75d2e6235e1cd927918565e48519b/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java#L101) as the current connector version forces the usage of the dcp 1.0 version, there is need to ensure the usage of the 0.8 in the current environment.

Closes #1982

### DIFF
--- a/charts/tractusx-connector/templates/deployment-dataplane.yaml
+++ b/charts/tractusx-connector/templates/deployment-dataplane.yaml
@@ -170,8 +170,6 @@ spec:
               value: {{ include "txdc.dataplane.url.control" . }}
             - name: "EDC_DPF_SELECTOR_URL"
               value: {{ include "txdc.controlplane.url.control" . }}/v1/dataplanes
-            - name: "EDC_DCP_V08_FORCED"
-              value: "true"
 
             #######
             # AWS #


charts/tractusx-connector/templates/deployment-dataplane.yaml
```diff
@@ -170,6 +170,8 @@ spec:
               value: {{ include "txdc.dataplane.url.control" . }}
             - name: "EDC_DPF_SELECTOR_URL"
               value: {{ include "txdc.controlplane.url.control" . }}/v1/dataplanes
+            - name: "EDC_DCP_V08_FORCED"
```
lgblaumeiser, line 173: Do we need this in the dataplane? I doubt that, because the dataplane should not communicate with the wallet or does it?

lgblaumeiser (commented):
One question

lgblaumeiser (approved):
lgtm, please remove from dataplane

charts/tractusx-connector/templates/deployment-dataplane.yaml
bmg13: You're right, removed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/2014/commits/6ebed6ff6e9b8b529ea9ff7a34c0820b3a1211af).

PR #2019, opened 2025-06-04
docs: add non finite provider push DR

correiaafonso12:
## WHAT

Adds a DR proposal for the non-finite provider push feature

## WHY

Define how non-finite data may be transferred using the Provider-PUSH flow

## FURTHER NOTES

This feature was previously discussed in this GH discussion: https://github.com/eclipse-tractusx/tractusx-edc/discussions/1994

### DIFF
--- a/docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
+++ b/docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
@@ -1,0 +1,37 @@
+# Non Finite Provider Push
+
+## Decision
+
+Provide explicit support for non-finite data transfers using the Provider-PUSH flow type.
+
+## Rationale
+
+Non-finite data is data that is defined by an infinite set or has no specified end. Currently, Provider-PUSH transfers transition the Transfer Processes and Data Flow to final states after the first data chunk is transferred, closing the communication channel between the Consumer and the Provider. This behavior prevents the exchange of non-finite data, as it incorrectly indicates that the transfer is complete, even though additional data will become available over time.
+
+## Approach
+
+With non-finite data, Transfer Processes will continue indefinitely until either the Consumer or Provider explicitly terminates the transmission. By doing so, the Transfer Process correctly reflects the state of the data transfer. The implemented solution will allow both finite and non-finite data to be transferred using the Provider-PUSH flow. Additionally, a triggering mechanism is required to enable the transfer of subsequent data chunks.
+
+1. Add an "isNonFinite" property to the `Asset`'s `DataAddress`:
+    - The `DataAddress` is schemaless, so it can already receive any custom property
+    - This property will indicate the data's finiteness and guide the `DataFlow` lifecycle.
+
+2. Perform the data transfer:
+    - Add a new extension that implements `PipelineService`, supporting both finite and non-finite data transfers.
+        - The registration of `DataSourceFactory` and `DataSinkFactory` remains the same as in the existing `PipelineServiceImpl`.
+        - The `transfer` method executes the data transfer as usual, and then evaluates the "isNonFinite" property:
+            - If the property is present, set to `true` and the transfer succeded, a `CompletableFuture` that never completes is returned, keeping the Data Flow *STARTED*.
+            - If the property is present, set to `true` but the transfer failed, a completed `CompletableFuture` is returned with the failure result.
+            - If the property is absent or set to `false`, the `CompletableFuture` produced by the data transfer is returned, preserving the current behavior for finite transfers.
+
+3. Add a new dataflow API to the management API context:
+    - This API will be available in the dataplane.
+    - This API will provide an endpoint to trigger a data transfer by a dataflow ID.
+
+4. Add a new `DataFlowService`:
+    - The service will include a `trigger` method that accepts a dataflow ID.
+    - This method will get the correspondant `DataFlow` and apply the following validations:
+        - The `DataFlow` must be of *PUSH* flow type.
+        - The `DataFlow` must contain the "isNonFinite" property.
+        - The `DataFlow` must be in the *STARTED* state.
+    - If all validations succeed, the `DataFlow` transitions to *RECEIVED*, restarting the data transfer.


correiaafonso12:
@lgblaumeiser regarding the [immediate push transfer](https://github.com/eclipse-tractusx/tractusx-edc/discussions/1994#:~:text=should%20think%20through%3A-,Can%20we%20avoid%20the%20immediate%20push%20transfer%20after%20the%20transfer%20is,it%20to%20me.%20There%20is%20no%20immediate%20data%20requested%20by%20concept.,-How%20does%20the), for now I've left the original approach, because from a technical standpoint there is still uncertainty on how to tackle it. If a new approach is decided, I'll update the DR accordingly

rafaelmag110 (changes requested):
This goes in the direction of the discussion. My suggestion for the implementation is to split this work into different issues and tackle them one at the time, otherwise the reviews will get complex and take more time.

The discussion about the use of an incomplete CF was not yet resolved, but in my perspective that does not impact the merge of the overall decision.

I'll suggest the removal of any reference to implementation details in this decision record. The record should reflect only the decision taken towards the addition/change of component in a higher level. Concrete implementation details should be avoided since those can be altered during or after the implementation.

With this I'm not indicating that you should proceed to implementation before fully clarifying implementation details and resolving the ongoing discussions.

correiaafonso12:
> This goes in the direction of the discussion. My suggestion for the implementation is to split this work into different issues and tackle them one at the time, otherwise the reviews will get complex and take more time.

Makes sense to me. I believe that the most logical split would be to have this DR stating the complete approach, a PR implementing how to [identify and handle non-finite data](https://github.com/eclipse-tractusx/tractusx-edc/discussions/1994#:~:text=UC1%20%2D%20Identify%20Non%2DFinite%20Assets), and another PR for the [trigger endpoint](https://github.com/eclipse-tractusx/tractusx-edc/discussions/1994#:~:text=UC2%20%2D%20Trigger%20a%20Data%20Transfer).

> I'll suggest the removal of any reference to implementation details in this decision record. The record should reflect only the decision taken towards the addition/change of component in a higher level. Concrete implementation details should be avoided since those can be altered during or after the implementation.

I'll adapt the DR to abstract the approach 👍

ndr-brt (changes requested):
> I'll suggest the removal of any reference to implementation details in this decision record. The record should reflect only the decision taken towards the addition/change of component in a higher level. Concrete implementation details should be avoided since those can be altered during or after the implementation.

I agree that we shouldn't mention variable names, but some code examples and technical overview will be needed to have an hi-level agreement on how to proceed, as it will be not easy to spot points of attention during the specific implementation PRs.
 
The DR as it is is too vague

correiaafonso12:
> The DR as it is is too vague

Added more detail to the approach chapter, balancing a deeper technical overview without stating concrete implementation details.

bmg13 (approved):

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Provide explicit support for non-finite data transfers using the Provider-PUSH flow type.
+
+## Rationale
+
+Non-finite data is data that is defined by an infinite set or has no specified end. Currently, Provider-PUSH transfers transition the Transfer Processes and Data Flow to final states after the first data chunk is transferred, closing the communication channel between the Consumer and the Provider. This behavior prevents the exchange of non-finite data, as it incorrectly indicates that the transfer is complete, even though additional data will become available over time.
+
+## Approach
+
+The first step is to enable Providers to indicate the finiteness of their data. This will be done during Asset registration by explicitly declaring the data's finiteness in the data address as a new property.
```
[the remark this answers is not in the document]
correiaafonso12: Added "... as a new property named `<new_property_name>`", which also state the property name 👍

rafaelmag110 (approved):

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+Provide explicit support for non-finite data transfers using the Provider-PUSH flow type.
+
+## Rationale
+
+Non-finite data is data that is defined by an infinite set or has no specified end. Currently, Provider-PUSH transfers transition the Transfer Processes and Data Flow to final states after the first data chunk is transferred, closing the communication channel between the Consumer and the Provider. This behavior prevents the exchange of non-finite data, as it incorrectly indicates that the transfer is complete, even though additional data will become available over time.
+
+## Approach
+
+The first step is to enable Providers to indicate the finiteness of their data. This will be done during Asset registration by explicitly declaring the data's finiteness in the data address as a new property named `isNonFinite`.
```
ndr-brt, line 13: providers won't indicate the "finiteness", but its contrary. 
please rephrase in a more schematic way, like:
"Provider will be enabled to specify non-finite `DataAddress` by setting a `isNonFinite` property as `true`"

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The implemented solution will provide an implementation of the `PipelineService` that allow both finite and non-finite data to be transferred using the Provider-PUSH flow. The behavior for finite data transfers will remain unchanged. With non-finite data, Transfer Processes will continue indefinitely until either the Consumer or Provider explicitly terminates the transmission, therefore correctly reflecting the state of the data transfer. This will be possible by guaranteeing that the `transfer` method returns a `CompletableFuture` that never completes in case of a successful, non-finite transfer.
```
ndr-brt, line 15: It would be nice to see some more design details, and also read how a non-finite data flow can be suspended/terminated, technically speaking.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The implemented solution will also include a triggering mechanism that enables the transfer of subsequent data chunks, allowing the active Transfer Processes and Data Flow to be reused. This mechanism will be exposed as a Provider-side endpoint on the Data Plane, enabling data transfers to be restarted on demand. A new `DataFlowApi` holding this endpoint will be added to the management API context. Additionally, a new `DataFlowService` will also be created, handling the logic to enable triggering a data transfer.
```
ndr-brt, line 17: More details please on the designed solution, class/method names, how the components interact...
Also it's not clear where the "subsequent data chunks" are coming from, I guess we are taking for granted that the source called is the same as the one defined in the source `DataAddress`, but I'd like to have it written for clarity.

rafaelmag110: I still don't understand what is the point of providing such level of detail in a decision record. The decision record should provide a detailed explanation of the decision taken by the group and the rational behind it. It should also provide an high level description of what has to change and/or what approach shall be taken from that point onward to honor the decision. But once you get into the finner implementation details, then I don't see the point.

What i'm saying is that we should care for finner details, yes, to understand exactly what taking this decision means and even to maintain a certain level of cooperation between the group. It will definitely help once the implementation PR comes and one has to review it. 

But clarifying all these details in written format is probably not the most efficient approach? I'd request Afonso to speak up in the next weekly and explain there how he plans to execute the decision. Otherwise this will go on forever because each new topic will generate an endless discussion on its own...

rafaelmag110: And i'm not saying your points are not valid, because they are very valid. However, we still have room to improve our collaboration has a group for it to be more efficient...

ndr-brt: well, the DR is the artifact that permits the committers to agree on the design choices, especially when the change is _that_ impactful, but's hard to agree on a design that's not explained, what usually happens with shallow DRs is that surprises then pop up during implementation and at that stage design discussions will be harder and slower (and less cost effective).

Presentations during weekly are ok to gather some general agreement, but details need to be written down "on paper" because in some months, or years when somebody will ask "why this has been made like this?", they will find the answer in the DR.

I also provided an example of good shaped DR, I could link more of them if you need.

TL;DR: we are not deciding whether or not to do this thing (this has already be decided), but **how** to do it.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,17 @@
+# Non Finite Provider Push
+
+## Decision
+
+Provide explicit support for non-finite data transfers using the Provider-PUSH flow type.
+
+## Rationale
+
+Non-finite data is data that is defined by an infinite set or has no specified end. Currently, Provider-PUSH transfers transition the Transfer Processes and Data Flow to final states after the first data chunk is transferred, closing the communication channel between the Consumer and the Provider. This behavior prevents the exchange of non-finite data, as it incorrectly indicates that the transfer is complete, even though additional data will become available over time.
```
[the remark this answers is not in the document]
correiaafonso12: Implemented at 8405b2b

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: Saying that something is non-finite is still describing its finiteness. It's still making a claim about the property finiteness, just that the property does not hold.

Nevertheless, I've rephrased the paragraph to make it more schematic, adding a lot more technical detail. Implemented at 8405b2b

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: Added more details at 8405b2b

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This feature consists of three main parts:
+
+- Identify Non-Finite Data Flows;
+- Handle Non-Finite Data During Transfers;
+- Trigger Additional Data Transfers.
+
+### Identify Non-Finite Data Flows
+
+A new `FinitenessEvaluator` interface will be added exposing the methods `isNonFinite(DataFlow dataflow)` and `isNonFinite(DataFlowStartMessage message)`. This model allows for custom implementations of ways to identify non-finite data. A default implementation will also be available, enabling Providers to indicate the finiteness of their data during Asset registration by setting a new property, `isNonFinite`, in the data address. By default, data will be considered finite and will only be treated as non-finite if `isNonFinite` is set to `true`.
```
ndr-brt, line 21: It's not clear in which context this `FinitenessEvaluator` will be used.

EDIT: it's shown in the code after. this explanation can be moved down below

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Handle Non-Finite Data During Transfers
+
+With non-finite data, Transfer Processes will continue indefinitely until either the Consumer or Provider explicitly terminates the transmission. Additionally, the implemented solution will keep the corresponding Data Flow in the *STARTED* state after a data transfer, thereby preventing the completion of the Transfer Process. Participants will continue to be able to terminate the Transfer Process using the [current termination mechanism](https://github.com/eclipse-edc/Connector/blob/v0.13.0/extensions/control-plane/api/management-api/transfer-process-api/src/main/java/org/eclipse/edc/connector/controlplane/api/management/transferprocess/v3/TransferProcessApiV3.java#L125).
```
ndr-brt, line 25: "Transfer Processes will stay in the `STARTED` state until it gets explicitly terminated or suspended by the Provider or the Consumer"

ndr-brt, line 25: also, could you add line breaks around 120 chars or so? that will help readability and review as well

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+To achieve this, a new implementation of the `PipelineService` will be provided, allowing both finite and non-finite data to be transferred using the Provider-PUSH flow. This implementation will be similar to upstream's `PipelineServiceImpl`, with the following changes to the `transfer` method:
```
ndr-brt, line 27: why the current `PipelineServiceImpl` cannot be used to transfer finite data and the new implementation used only for the non-finite ones? please motivate this choice properly.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- For finite data transfer, return a `CompletableFuture` with the transfer result (current behavior);
+- For failed non-finite data transfer, return a `CompletableFuture` with the failed transfer result;
+- For successful non-finite data transfer, return a `CompletableFuture` that never completes, preventing subsequent state transitions.
```
ndr-brt, line 31: it's not clear how this flow can be stopped by a termination or a suspension

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,71 @@
...
+            });
+
+    // Keep current behavior for finite data transfers
+    if (!finitenessEvaluator.isNonFinite(request)) {
+        return futureTransferResult;
+    }
+
+    // Complete transfer only if transfer result failed for non-finite data
+    var returnedFuture = futureTransferResult.newIncompleteFuture();
+    futureTransferResult.thenAccept(result -> {
```
ndr-brt, line 51: I'm not sure that  the whole non-finite flow should be stopped by an error in the first transmission, e.g. if the source is not available maybe the user would like to being able to trigger it again instead to have to re-create the transfer from scratch.

If we want to keep this behavior (stop non-finite transfer if first transmission doesn't work) we should think about having also the subsequent transmissions triggered to be able to interrupt the flow in case of error

ndr-brt (changes requested):
definitely in a better shape, thanks!
I still have some comments.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: I've decided to keep this section here as I find that it aligns better with the user journey (identify non-finite data -> transfer non-finite data). Nevertheless, I've added a sentence referring that the `FinitenessEvaluator` usage will be described below.

Please let me know if this is clear.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: Implemented at 05fb7a8

correiaafonso12: Couldn't comply fully with the 120 chars limit, but added some line breaks where needed :)

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: While many `TransferService` implementations may exist simultaneously, registering them in the `TransferServiceRegistry` and resolving the appropriate one when needed, [only one](https://eclipse-edc.github.io/documentation/for-contributors/runtime/dependency-injection/#7-limitations:~:text=No%20multiple%20registrations%3A%20registering%20two%20implementations%20for%20an%20interface%20will%20result%20in%20the%20first%20registration%20being%20overwritten%20by%20the%20second%20registration.%20If%20both%20providers%20have%20the%20same%20topological%20ordering%20it%20is%20undefined%20which%20comes%20first.%20A%20warning%20is%20posted%20to%20the%20Monitor) `PipelineService` implementation must be available at runtime, since it is explicitly [injected](https://github.com/eclipse-edc/Connector/blob/v0.13.0/core/data-plane/data-plane-core/src/main/java/org/eclipse/edc/connector/dataplane/framework/DataPlaneFrameworkExtension.java#L115-L116).

As this approach relies on the functionalities brought by the `PipelineService`, e.g. the ability to register n `DataSourceFactory` and `DataSinkFactory`, and only one implementation must be present, this `PipelineService` implementation must support both finite and non-finite data.

Having a single `PipelineService` implementation was a previous design decision. Should it be mentioned in this DR? If yes, should it be something similar to the explanation in this comment?

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: Nice catch, I was missing changes on the `terminate` method. Added at 05fb7a8

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
correiaafonso12: The decision to adopt a "fail-fast" approach, i.e. terminate the Data Flow if an error occurs, was taken to ensure visibility to transfer failures. Due to the asynchronous nature of data transfers, if an error occurred without terminating the Data Flow, the Provider would not have any way to detect it or know its cause.

> If we want to keep this behavior (stop non-finite transfer if first transmission doesn't work) we should think about having also the subsequent transmissions triggered to be able to interrupt the flow in case of error

Not sure I got this point. Triggered transfers will behave like other transfers, meaning that the Data Flow will terminate if an error occurs.

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Trigger Additional Data Transfers
+
+The implemented solution will also include a triggering mechanism that enables the transfer of subsequent data chunks, allowing the active Transfer Processes and Data Flow to be reused.
+
+This mechanism will be based on the `PipelineService` transfer behavior, retrieving data from the defined source data address and transfering it to the defined data destination.
+
+The trigger mechanism will be exposed as a Provider-side endpoint on the Data Plane, allowing data transfers to be restarted on demand.
+
+A new `DataFlowApi` and its corresponding controller will be added to the management API context. This API will expose a "trigger endpoint" at `POST /api/v1/dataflows/{dataflowId}/trigger`. When invoked, this endpoint will call a new component, the `DataFlowService`, responsible for handling the logic to trigger a data transfer. The `DataFlowService` will contain a `trigger` that applies the following validations:
```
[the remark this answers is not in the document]
correiaafonso12: Applied at 650bdc9

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
ndr-brt: I think a journey is more easy to understand if it goes outside-in, and not inside-out, but, in any case, I'm ok with the current shape in any case

docs/development/decision-records/2025-06-04-non-finite-provider-push/README.md
ndr-brt: I didn't said to propose this feature upstream as a whole, but to have a refactoring there that would avoid us to re-implement and maintain the whole PipelineService, because I don't think it will be beneficial to have full maintenance in this round.
At least it worth mention this in the DR.

correiaafonso12: Added a "Future Work" section mentioning this at 8421ce0

ndr-brt (approved):
Looks really good

PR #2030, opened 2025-06-13
fix: restored copyright to all eclipse contributors

### LINKED ISSUE #2029: Invalid License at EDC Helm Charts & main license
## Description

For some reason the EDC has included a incorrect license at their charts:

https://github.com/eclipse-tractusx/tractusx-edc/blob/4dc2d6d6b705d436a7893421441f8cab5edb2f05/charts/tractusx-connector-memory/LICENSE#L189

https://github.com/eclipse-tractusx/tractusx-edc/blob/4dc2d6d6b705d436a7893421441f8cab5edb2f05/charts/tractusx-connector/LICENSE#L189

And in the main license too!
https://github.com/eclipse-tractusx/tractusx-edc/blob/4dc2d6d6b705d436a7893421441f8cab5edb2f05/LICENSE#L189

It was referenced for Catena-X at 2022 and that is not the case anymore, since Catena-X can't hold copyright for what is in Tractus-X.

Heads up for @CDiezRodriguez which found this bug in the Identity Hub and I checked from where it came from and it come from here...

matbmoser:
## WHAT

- Removed Catena-X Copyright and substituted with empty copyright owner

## WHY

As specified in this issue #2029 the Licenses used are pointing to Catena-X and that is not true anymore since the consortia died. Or you add the Catena-X Automotive Network e.V.  Association or you remove it so it valid for all the eclipse contributors.

I am not sure if it should be the eclipse foundation.

I am removing it so it would be valid for everyone like in other repos.

## FURTHER NOTES

Closes #2029

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2022 Catena-X
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/charts/tractusx-connector-memory/LICENSE
+++ b/charts/tractusx-connector-memory/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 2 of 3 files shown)


ndr-brt (approved):
thx

PR #2064, opened 2025-07-01
[TAP 7.2] - Event Logging Subscriber

### LINKED ISSUE #1949: Provide event information in logs
## WHAT

The Tractus-X connector has an embedded event mechanism that informs potential listeners about specific incidents, like finalized contract negotiations. These events should be routed using the standard log mechanism to provide the information for further processing, e.g., in Dashboards

## WHY
Better transparency of domain activities within the connector

## HOW

An event listener will be implemented that receives the internal events and transforms them into a log entry. This is related to the issue https://github.com/eclipse-tractusx/sig-release/issues/1280, as the log entry would then be logged in a structured way, that allows further processing steps to work with semantically richer data.

kkotowiczz:
## WHAT
Added optional extension which logs all the domain events to OTEL Collector by direct HTTP call.
## WHY
To explicitly gather information from control and dataplane in grafana / loki stack

## FURTHER NOTES
N/A - new extension

Closes #1949


ndr-brt:
removed myself from the reviewers

PR #2066, opened 2025-07-01
feat(logging): Event subscriber

### LINKED ISSUE #1949: Provide event information in logs

kkotowiczz:
## WHAT
Added optional extension which logs all the domain events to OTEL Collector by direct HTTP call.
## WHY
To explicitly gather information from control and dataplane in grafana / loki stack

## FURTHER NOTES
N/A - new extension

Closes #1949


th0masgl:
> In addition, and a very important step since this adds functionality to our distribution: this needs a decision record! ...

I created the PR for the decision record here : https://github.com/eclipse-tractusx/tractusx-edc/pull/2084

edc-extensions/event-subscriber/src/main/java/org/eclipse/tractusx/edc/eventsubscriber/EventSubscriberExtension.java
```diff
@@ -0,0 +1,66 @@
...
+import org.eclipse.edc.spi.system.ServiceExtensionContext;
+import org.eclipse.edc.spi.types.TypeManager;
+
+import static org.eclipse.tractusx.edc.eventsubscriber.EventSubscriberExtension.NAME;
+
+@Extension(value = NAME)
+public class EventSubscriberExtension implements ServiceExtension {
+
+    public static final String NAME = "Tractus-X Event Subscriber Extension";
+    @Setting(required = false, key = "tx.edc.otel.logs.endpoint", defaultValue = "http://localhost:4318/v1/logs")
```
rafaelmag110, line 38:
```suggestion
    @Setting(required = false, key = "tx.edc.otel.events.endpoint", defaultValue = "http://localhost:4318/v1/logs")
```

Small nit since we might require an endpoint for logs in the future. This is for events.

rafaelmag110 (approved):

PR #2085, opened 2025-07-09
doc: add decision record for logging and metrics framework

### LINKED ISSUE #1745: Metric framework decision
## WHAT

A decision record is created that describes, how metrics are created and provided within the connector. In addition, documentation describes that usage of the framework. Metrics are related to the domain level, i.e., business level metrics like number of contracts, number of exchange partners, transfer processes and on the other hand to a technical level to manage the operations like certain resource metrics.

## WHY

As operator of a Tractus-X connector, it is crucial to get information on the status of the connector in order to manage the operation. As domain expert providing a certain use case, also information about certain activities are needed, e.g., to gete information relevant in certain audits. There should be a general approach how the metrics should be realized and how they are forwarded to backend systems, so that it can be processed further according to the reporting needs.

## HOW

There are two potential solution proposals for now:
1. Provide metrics using a family of metrics endpoints
2. Use the OpenTelemetry infrastructure that is already included to provide the metrics

The decision record should outline the intended solution. A documentation should give insides on how to use the infrastructure to create metrics. If necessary, implementation tasks should be created that describe necessary actions to get to a realized framework for the metrics.

lgblaumeiser:
## WHAT

A decision record that describes the overall decision concerning how logging information and further events are handled by the connector and the approach to compute metrics for better operations of a connector.

## WHY

The framework described allows a more standardized and streamlined operation of connectors. Major information is made available in a way, that is easily processable and which allows to get an overview over current activities within a connector on a technical level as well as on a domain level.

Fixes #1745

### DIFF
--- a/docs/development/decision-records/2025-07-09-logging_and_metrics_framework/README.md
+++ b/docs/development/decision-records/2025-07-09-logging_and_metrics_framework/README.md
@@ -1,0 +1,36 @@
+# Logging/Eventing/Metrics Framework
+
+The purpose of this decision record is to agree on the general strategy concerning how log information and domain
+event information is processed in the connector and in background services that take the information and process
+it further.
+
+## Decision
+
+1. The connector provides log information in a structured way in the form of json objects.
+2. The technology to provide execution information towards background services is OpenTelemetry.
+3. OpenTelemetry is added as direct dependency. The target is that the usage of the java agent is abandonned.
+4. The connector will not calculate domain metrics. Instead it will provide relevant domain events that are forwarded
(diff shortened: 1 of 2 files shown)


rafaelmag110 (approved):

rafaelmag110:
This is a nit.
I'm finding it hard to associate this decision with the _framework_ concept. The way I see it we are more deciding on a **_strategy_** to provide logging and eventing information - for uses cases to be built upon them - than a framework.

bmg13 (approved):

Issue #2083, opened 2025-07-08
Adapt handling of policies according to new policy constraint definitions


lgblaumeiser:
## WHAT

See https://github.com/eclipse-tractusx/sig-release/issues/1311 for details on the new constrains

## HOW

- Change policy validation
- Enhance testing to consider new policy constraints in, e.g., end-to-end tests

farhin23:
Hi, @lgblaumeiser 
Can you please assign this issue to me?

PR #2101, opened 2025-07-15
feat: add policy schema validators

farhin23:
## WHAT

This PR adds `CxPolicyDefinitionValidator `, a `JsonObjectValidator` to validate the structure of policies according to the schema defined in [cx-odrl-profile](https://github.com/catenax-eV/cx-odrl-profile). Adds support for validating ACCESS and USAGE policy types. 

## WHY

The ensure a correct structure and definition of the policies at their creation time. 

## FURTHER NOTES
* The `PolicyValidationConstants` class is intended to contain all relevant constants, including allowed constraints and left operands. Currently, it only includes the existing constraint functions and is not yet complete. Once the changes from PR(#2086) are merged, I will update this file with the new constraint left operand values. However, this PR does not depend on those changes.
* The `PolicyDefinitionEndToEndTest` class now includes additional tests to cover invalid policies with various structures.
* Some tests from certain E2E test classes (e.g., `CatalogTest`, `AbstractIatpConsumerPullTest`) have been removed, as they are no longer applicable under the newly defined policy schema. Keeping them would cause failures with the current policy validation implementation.
* In `PolicyHelperFunctions`, some method signatures have been updated, specifically to include the action type information in rules. The E2E tests using these methods have been updated accordingly.

Related to #2083


edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/validator/UsagePolicyValidator.java
```diff
@@ -0,0 +1,103 @@
...
+    private final JsonLdPath path;
+
+    public UsagePolicyValidator(JsonLdPath path) {
+        this.path = path;
+    }
+
+
+    @Override
+    public ValidationResult validate(JsonObject input) {
+        var structureValidator = validateBasicStructure(input);
```
rafaelmag110, line 49: The `validateBasicStructure` seems redundant. The check for AtLeastOneRuleExists is done twice.
Can't we have the TypedMandatoryArray verifications and the UsageXValidator in the same validator?

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/validator/ArrayIsEmpty.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        var value = input.get(path.last());
+
+        if (value.getValueType() == JsonValue.ValueType.NULL) {
+            return ValidationResult.success();
+        }
+
+        if (value.getValueType() != JsonValue.ValueType.ARRAY) {
+            return ValidationResult.failure(
+                    violation("Array must be of type array", path.toString())
```
rafaelmag110, line 50:
```suggestion
                    violation("Must be an empty array", path.toString())
```

Nitpick: Combine both this and the below if statement into one.

rafaelmag110:
I'm also labeling this PR as a breaking change since this might break existing interfaces with the connector where policy creations are not according to standard.

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/CxPolicyExtension.java
```diff
@@ -121,5 +133,6 @@ public String name() {
     public void initialize(ServiceExtensionContext context) {
         registerFunctions(policyEngine);
         registerBindings(bindingRegistry);
+        validatorRegistry.register(EDC_POLICY_DEFINITION_TYPE, CxPolicyDefinitionValidator.instance());
```
rafaelmag110, line 136: this should be changed into a different stage of the connector boot phase, otherwise we are dependent in the order which extensions are initialized.

rafaelmag110:
@farhin23 is this already ready for a re-review?

edc-tests/e2e/catalog-tests/src/test/java/org/eclipse/tractusx/edc/tests/catalog/CatalogTest.java
```diff
@@ -95,7 +95,7 @@ void requestCatalog_fulfillsPolicy_shouldReturnOffer() {
     @Test
     @DisplayName("Verify that the consumer receives only the offers he is permitted to (using the legacy BPN validation)")
     void requestCatalog_filteredByBpnLegacy_shouldReject() {
-        var onlyConsumerPolicy = bpnPolicy("BPN1", "BPN2", CONSUMER.getBpn());
```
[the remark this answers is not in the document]
farhin23: I have adapted these two tests to create a policy using the operator anyOf. This required modifying the `BusinessPartnerNumberPermissionFunction` evaluation logic for the AnyOf operator.

farhin23:
Hi @rafaelmag110 it's ready for a re-review.

farhin23:
Hi @lgblaumeiser ,
I have a question regarding the invalid argument provider (`InvalidContractPolicyProvider`) in the `AbstractIatpConsumerPullTest`: https://github.com/eclipse-tractusx/tractusx-edc/blob/main/edc-tests/e2e/iatp-tests/src/test/java/org/eclipse/tractusx/edc/tests/transfer/AbstractIatpConsumerPullTest.java . At the moment, the arguments only include dismantler credentials. Based on our last discussion, we cannot include these credentials in any of access or usage policy. However, if we remove them, the test arguments will be empty.
Could you clarify the purpose of this test and advise how we should handle it in this case? Its related to one of the failing tests.

lgblaumeiser:
Hi @farhin23 , can you ping me, when this PR is ready for review again, thanks!

farhin23:
> Do we have a problem with this setup?

Keeping the Dismantler credentials will fail the tests as it will fail the policy validation at creation time.

> If you have another idea, how we can force a situation, where a provider connector would refuse a contract because the consumer

I can not think of any such scenarios. For now I will get rid of this test then.

farhin23:
Hi @lgblaumeiser 
its ready for a re-review. Some E2E tests are failing for connection errors. I have tested those locally and works fine.

edc-extensions/cx-policy/src/test/java/org/eclipse/tractusx/edc/policy/cx/validator/PolicyTypeResolverTest.java
```diff
@@ -0,0 +1,186 @@
...
+                .build();
+        JsonObject policy = Json.createObjectBuilder()
+                .add(ODRL_PROHIBITION_ATTRIBUTE, Json.createArrayBuilder().add(prohibition))
+                .build();
+
+        String result = PolicyTypeResolver.resolve(policy);
+
+        assertThat(result).isEqualTo("odrl:distribute");
+    }
+
```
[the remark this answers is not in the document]
farhin23: The updated implementation throws exception with proper message in this case.

lgblaumeiser (approved):
lgtm, still we have to wait for the ip check

PR #2120, opened 2025-07-25
feat: update cx policy namespace

farhin23:
## WHAT

Updates the CX namespace to `https://w3id.org/catenax/2025/9/policy/`, in alignment with the latest version of the policy schema [context](https://github.com/catenax-eV/cx-odrl-profile/blob/33d62716335f567d6e3df718779bb7f1ae743779/schema/context/context.jsonld#L4), which now defines the CX namespace as `https://w3id.org/catenax/2025/9/policy/`.

## WHY

This change ensures consistency with the updated policy schema specification.

## FURTHER NOTES

Related to #2083


lgblaumeiser (commented):
Again my question here is, how is this concerning backward compatibility, with two questions:
1. When a provider updates his connector, are there migrations that need to be done?
2. Is there an interaction between a consumer and a provider, that we break with the change?

PR #2121, opened 2025-07-25
Deprecate the Mock Connector Runtime

### LINKED ISSUE #1918: Deprecate the Mock Connector Runtime
## WHAT

Deprecate the mock-connector runtime from the application.

## WHY

This runtime does not seem to be used by anyone and so the effort of maintaining it is unnecessary.
It was created aiming contract testing, however lacks usage and other solutions are available.

If interest in usage rises in the future, the inclusion of a mock server can be reconsidered then.

## HOW

Notify interested parties (in Matrix chat) and deprecate the runtime with eventually deletion of the [runtime module](https://github.com/eclipse-tractusx/tractusx-edc/tree/0bf683e7a799dc52f84a153cba19e70abd6e6ec7/edc-tests/runtime/mock-connector) from the repository.

bmg13:
## WHAT

Deprecates the mock connector runtime since the v0.11.0.

## WHY

This runtime does not seem to be used by anyone and so the effort of maintaining it is unnecessary.

Added javadoc due checkstyle's failing.

Closes #1918


lgblaumeiser (approved):
lgtm

PR #2131, opened 2025-07-29
feat: Allow to create dated SNAPSHOT manually

### LINKED ISSUE #2130: Make possible to create dated SNAPSHOT manually
## WHAT

The `Publish new snapshot` workflow currently creates a dated SNAPSHOT only when it’s triggered by the scheduled `Run-All-Tests` workflow (once per week). We need to add the ability to manually trigger daily SNAPSHOTs on demand.

## WHY
To provide greater flexibility in producing daily SNAPSHOT builds outside of the weekly schedule.

## HOW

Introduce a manual-trigger option with a checkbox offering two modes for `Run-All-Tests` workflow:
   **Regular (default):** Behaves as today, creating only undated SNAPSHOTs.
   **Dated:** When selected, generates a dated SNAPSHOT upon manual execution of the workflow.

AndrYurk:
## WHAT

This PR makes it possible to run the "Publish new snapshot" workflow manually with identification of whether it is dated or not.

## WHY

To provide flexibility in producing daily SNAPSHOT builds outside of the weekly schedule.

## FURTHER NOTES
As we do not use the `bugfix/*` in the new [release process](https://github.com/eclipse-tractusx/tractusx-edc/tree/main/docs/development/decision-records/2025-05-26-release-process), it will be changed to `release/*`

Closes #2130

### DIFF
--- a/.github/workflows/publish-new-snapshot.yaml
+++ b/.github/workflows/publish-new-snapshot.yaml
@@ -30,10 +30,16 @@ on:
     workflows: [ "Run-All-Tests" ]
     branches:
       - main
-      - bugfix/*
+      - release/*
     types:
       - completed
   workflow_dispatch:
+    inputs:
+      dated_snapshot:
+        description: 'Dated snapshot version.'
+        default: false
+        type: boolean
+        required: true
 
   # periodic build triggers are defined in run-all-tests.yml, which triggers this workflow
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-new-snapshot.yaml
```diff
@@ -30,10 +30,16 @@ on:
...
+      - release/*
     types:
       - completed
   workflow_dispatch:
+    inputs:
+      dated_snapshot:
+        description: 'Dated snapshot version.'
+        default: false
+        type: boolean
+        required: true
```
bmg13, line 42: why is it required?

.github/workflows/publish-new-snapshot.yaml
```diff
@@ -30,10 +30,16 @@ on:
     workflows: [ "Run-All-Tests" ]
     branches:
       - main
-      - bugfix/*
+      - release/*
```
bmg13, line 33: remind me please why we are removing the bugfix. is not stated in description nor issue.

bmg13 (commented):
a couple of small clarifications

.github/workflows/publish-new-snapshot.yaml
lgblaumeiser: We do not use bugfix branches anymore, everything is a release branch, but you are right, it is an additional aspect worth to be mentioned in the pr description.

AndrYurk: Additional information has been added to the PR

.github/workflows/publish-new-snapshot.yaml
AndrYurk: @bmg13, @lgblaumeiser, generally, the workflow may operate without specifying both `required` and `default` (for example, when triggered by workflow_run), so I will remove them.

bmg13 (approved):
Please check my comment for the default value.

PR #2134, opened 2025-07-31
fix: add missing test packages to compatibility tests

### LINKED ISSUE #2040: Ensure a proper standard implementation of .well-known/dspace-version
## WHAT

The DSP 2025-1 protocol defines that the .well-known/dspace-version endpoint can be implemented in the service hierarchy of a connector, the DID document has to point to that endpoint for an initial connect. The endpoint has to return for each supported DSP version a path to the implementation of that DSP versions root endpoint, so that the subpaths defined in the DSP spec can be applied according to the version.

This implementation is either done upstream or has to be done in Tractus-X, either way, it must support version 0.8 and 2025-1 and no other version with an appropriate path.

## WHY

To meet the standard requirements of the upcoming CX-0018 version of the Catena-X association

rafaelmag110:
## WHAT

adds missing tck packages to compatibility tests. 

## WHY

To verify metadata and catalog endpoints according to dsp 2025-1

Closes #2040


lgblaumeiser (approved):
lgtm

PR #2161, opened 2025-08-13
feat: bpn policy namespace migration

farhin23:
## WHAT
This PR introduces three changes:
1. The BPN policy constraints (`BusinessPartnerNumber`, `BusinessPartnerGroup`) previously used the `TX_NAMESPACE = "https://w3id.org/tractusx/v0.0.1/ns/"`. Since these constraints are now part of the new policy schema defined under the CX_POLICY_NAMESPACE, their namespace has been updated accordingly to use the cx namespace.

2. Introduces two Flyway Java migration classes, `V0_0_7__Bpn_Namespace_Migration` and `V0_1_0__Bpn_Namespace_Migration`, to handle the migration of existing policies containing `BusinessPartnerNumber` or `BusinessPartnerGroup` constraints which uses the old namespace `https://w3id.org/tractusx/v0.0.1/ns/`. The migrations update the relevant policy rules in the database to ensure proper namespace handling and compatibility with the latest policy model.

3. Updates the CX namespace to https://w3id.org/catenax/2025/9/policy/, in alignment with the latest version of the policy schema [context](https://github.com/catenax-eV/cx-odrl-profile/blob/33d62716335f567d6e3df718779bb7f1ae743779/schema/context/context.jsonld#L4).

## WHY
Policies with BPN/BPG constraints require namespace adjustments to align with updated policy standards and ensure consistent enforcement.

## FURTHER NOTES
Migrated Tables and Columns:

| Table Name | Column Name | 
|-----------------|-----------------|
| edc_policydefinitions     | permissions, prohibitions, duties     |
| edc_contract_agreement     | policy     |
| edc_contract_negotiation     | contract_offers     |

Related to https://github.com/eclipse-tractusx/tractusx-edc/issues/2083

PR #2172, opened 2025-08-19
feat: cx namespace update

farhin23:
## WHAT

Updates the CX namespace to `https://w3id.org/catenax/2025/9/policy/`, in alignment with the latest version of the policy schema [context](https://github.com/catenax-eV/cx-odrl-profile/blob/33d62716335f567d6e3df718779bb7f1ae743779/schema/context/context.jsonld#L4), which now defines the CX namespace as `https://w3id.org/catenax/2025/9/policy/`.

## WHY

This change ensures consistency with the updated policy schema specification.

## FURTHER NOTES


farhin23:
Hi @lgblaumeiser 
The failing tests were related to BPN policies. I’ve made some adjustments in the cx-policy-v1.jsonld file to get them working, but this should be updated once we have the BPN rule functions (with the new namespace) in place.

rafaelmag110:
@farhin23 does this replace #2161?

rafaelmag110:
This PR will be closed as it has been replicated by #2182 .

PR #2198, opened 2025-08-26
fix: make possible to disable migration

### LINKED ISSUE #2197: Exception when disabling any DB migration
## Describe the bug

It is not possible to disable the DB migration using configuration options `tx.edc.postgresql.migration.%s.enabled` because of an exception:
```
SEVERE 2025-08-26T16:18:27.153138967 Error booting runtime: Cannot invoke "java.util.function.Supplier.get()" because "this.migrationExecutor" is null
java.lang.NullPointerException: Cannot invoke "java.util.function.Supplier.get()" because "this.migrationExecutor" is null
	at org.eclipse.tractusx.edc.postgresql.migration.AbstractPostgresqlMigrationExtension.prepare(AbstractPostgresqlMigrationExtension.java:87)
	at org.eclipse.tractusx.edc.postgresql.migration.AssetPostgresqlMigrationExtension.prepare(AssetPostgresqlMigrationExtension.java:23)
	at org.eclipse.edc.boot.system.injection.lifecycle.ExtensionLifecycleManager.bootServiceExtensions(ExtensionLifecycleManager.java:67)
	at org.eclipse.edc.boot.system.runtime.BaseRuntime.bootExtensions(BaseRuntime.java:158)
	at org.eclipse.edc.boot.system.runtime.BaseRuntime.boot(BaseRuntime.java:102)
	at org.eclipse.edc.boot.system.runtime.BaseRuntime.main(BaseRuntime.java:80)
```

### To Reproduce

Disable any migration for `control-plane` or `data-plane`.

### Expected behavior

It should be possible to disable any DB migration without exception.

## Possible Implementation

In class `AbstractPostgresqlMigrationExtension`, check `migrationExecutor` for null during preparation.

AndrYurk:
## WHAT

This PR provides a fix for the migration extension
## WHY
To make it possible to disable migration without exception.

Closes #2197

### DIFF
--- a/edc-extensions/migrations/postgresql-migration-lib/src/main/java/org/eclipse/tractusx/edc/postgresql/migration/AbstractPostgresqlMigrationExtension.java
+++ b/edc-extensions/migrations/postgresql-migration-lib/src/main/java/org/eclipse/tractusx/edc/postgresql/migration/AbstractPostgresqlMigrationExtension.java
@@ -64,6 +64,7 @@ abstract class AbstractPostgresqlMigrationExtension implements ServiceExtension 
 
         ConfigUtil.propertyCompatibility(context, MIGRATION_ENABLED_TEMPLATE.formatted(subSystemName), MIGRATION_ENABLED_TEMPLATE_DEPRECATED.formatted(subSystemName), Boolean.valueOf(DEFAULT_MIGRATION_ENABLED_TEMPLATE));
         if (!enabled) {
+            context.getMonitor().info("Migration for subsystem %s disabled".formatted(subSystemName));
             return;
         }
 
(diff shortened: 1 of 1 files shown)


edc-extensions/migrations/postgresql-migration-lib/src/main/java/org/eclipse/tractusx/edc/postgresql/migration/AbstractPostgresqlMigrationExtension.java
```diff
@@ -84,6 +85,10 @@ public void initialize(final ServiceExtensionContext context) {
 
     @Override
     public void prepare() {
+        if (migrationExecutor == null) {
```
rafaelmag110, line 88: I would instead check for the value of the migration enabled config and return if its disabled.
This migration might get a null value for some other weird reason, the behavior of the extension become unexpected and hard to detect.

AndrYurk: corrected

rafaelmag110 (approved):

lgblaumeiser:
PR for RC2: https://github.com/eclipse-tractusx/tractusx-edc/issues/2210

PR #2199, opened 2025-08-27
feat - added otel appender for log4j2 [TAP 7.2]

kkotowiczz:
## WHAT
Added log4j2 opentelemetry appender dependency 

## WHY
In order to integrate OTEL with log4j2 and umbrella helm charts

## FURTHER NOTES
Blocking:
[[TAP 7.2] - feat(observability): Log4j2 integration with OTEL](https://github.com/eclipse-tractusx/tractus-x-umbrella/pull/316)


rafaelmag110 (changes requested):
This would fit better in the control and dataplane base BOM. Also, the version of the module should be moved to libs.toml, as with all other dependencies we declare.

Also, isn't there a tag for a stable version instead of an "-alpha"? Should be a better option.

PR #2211, opened 2025-09-01
fix: Switch DataExchangeGovernance right operand version requirement back to 1.0

### LINKED ISSUE #2208: Change validation of FrameworkAgreement policy constraint to support version 1.0
## Describe the bug

The current validation for the Framework Agreement checks hardcoded for the right value `DataExchangeGovernance:2.0`. The issue is, that this left operand refers to a credential, which only exists in version 1.0 and which will not be reissued for the ecosystem release.

Therefore, the validation has to be changed to check for a right operand `DataExchangeGovernance:1.0`

### To Reproduce

Check code in edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/framework/FrameworkAgreementCredentialConstraintFunction.java

### Expected behavior

The only valid value for a right operand is `DataExchangeGovernance:1.0`

lgblaumeiser:
## WHAT

Instead of validating the policy creation by ensuring "DataExchangeGovernance:2.0" as right operand, require "DataExchangeGovernance:1.0" to be the expected value.

## WHY

The latest decision is, that there is no new credential version which means, that in policies the reference has to be to the version of the real existing credential, which is 1.0.

Closes #2208


rafaelmag110:
@lgblaumeiser will the standard be fixed then?

Issue #2212, opened 2025-09-02
Connector configuration doesn't work on BPDM INT environment


SujitMBRDI:
## Description

We as BPDM updated the existing consumer and provider EDCs to version 0.11.0-RC1  (with existing offers in the provider EDC) and followed the migration guide to update the participant-ID configuration. 
We executed a catalog request from the consumer to the provider EDC and received an error which we could not further identify from the provider EDC logs. (some unexpected jersey exception we received).

Later we discovered, The migration guide mentions that the property `edc.dcp.v08.forced`  should be removed. We saw this environment variable to be set to true in the current 0.11.0 Charts by default which sounds contradictory to the migration guide `EDC_DCP_V08_FORCED` . When we removed it we get another error:
 ```
[
    {
        "message": "Unable to obtain credentials: Failure in DID resolution: Expected exactly 1 VP, but was empty",
        "type": "BadGateway",
        "path": null,
        "invalidValue": null
    }
]
```

As far as the EDC,  it seems to work in a controlled environment but not currently on BPDM INT environment. Therefore, it is a bit uncertain if there are more issues with the Chart configuration or just issues with dependent services (or still a problem of our specific BPDM configuration).

### Expected behavior

BPDM so far has always used the INT environment to deploy and configure the EDCs so we can currently only test with a working INT.

rafaelmag110:
@SujitMBRDI please let us know exactly what request is failing so we can replicate the issue.

SujitMBRDI:
> [#2225](https://github.com/eclipse-tractusx/tractusx-edc/pull/2225) Addresses the catalog issue you described.

@rafaelmag110 Thank you for providing PR details.

SujitMBRDI:
With release candidate - [0.11.0-RC1 ](https://github.com/eclipse-tractusx/tractusx-edc/releases/tag/0.11.0-RC2) catalog request working as expected.
On BPMD INT environment,  I've applied work around mentioned in issue https://github.com/eclipse-tractusx/tractusx-edc/issues/2223 for e2e test.
So, keeping this issue open.

nicoprow:
@rafaelmag110 @lgblaumeiser 

We tested the catalog request again with an updated deployment without the forced DCP 0.8 flag and it worked also as expected. We tested 0.8 and 2025.1 protocol and both worked for us in the catalog request.

For us this issue can be closed

Issue #2210, opened 2025-09-01
Cherry-pick list for 0.11.0-RC2


rafaelmag110:
This issue should help us keep track of what PRs shall be considered for RC2.

Releasing RC2 shall close this PR.

rafaelmag110:
@lgblaumeiser @bmg13 FYI

PR #2213, opened 2025-09-02
fix: Remove switch to use dcp 1.0 from charts

lgblaumeiser:
## WHAT

Remove the EDC_DCP_V08_FORCED environment setting from the chart templates

## WHY

In 0.10.0 this was needed to enforce the usage of the old DCP protocol, since version 0.11.0 uses the new protocol, this setting must not be set.

Related to: #2212


lgblaumeiser:
Relevant for RC2: https://github.com/eclipse-tractusx/tractusx-edc/issues/2210

bmg13 (approved):

Issue #2217, opened 2025-09-02
Catalog request in jupiter (and Saturn) version not working


matbmoser:
## Describe the bug

I have upgraded my two Connectors (which were working in 0.10.0 at the int env) to v0.11.0-RC1.

Without changing the helm values or something.

And have done a simple catalog request to here:

```json
{
    "@context": {
        "edc": "https://w3id.org/edc/v0.0.1/ns/",
        "odrl": "http://www.w3.org/ns/odrl/2/",
        "dct": "https://purl.org/dc/terms/"
    },
    "@type": "CatalogRequest",
    "counterPartyAddress": "https://edc-provider-ichub-control.int.catena-x.net/api/v1/dsp",
    "counterPartyId": "BPNL0000000093Q7",
    "protocol": "dataspace-protocol-http",
    "additionalScopes": [],
    "querySpec": {
        "@context": {
            "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
        },
        "@type": "QuerySpec",
        "filterExpression": {
            "operandLeft": "'http://purl.org/dc/terms/type'.'@id'",
            "operator": "=",
            "operandRight": "https://w3id.org/catenax/taxonomy#DigitalTwinRegistry"
        }
    }
}
```

```json
[
    {
        "message": "org.eclipse.edc.http.spi.EdcHttpClientException: Server response to [POST, https://edc-provider-ichub-control.int.catena-x.net/api/v1/dsp/catalog/request] was not one of [2xx, 4xx] but was 500: <html>\n<head>\n<meta http-equiv=\"Content-Type\" content=\"text/html;charset=ISO-8859-1\"/>\n<title>Error 500 Internal Server Error</title>\n</head>\n<body>\n<h2>HTTP ERROR 500 Internal Server Error</h2>\n<table>\n<tr><th>URI:</th><td>http://edc-provider-ichub-control.int.catena-x.net/api/v1/dsp/catalog/request</td></tr>\n<tr><th>STATUS:</th><td>500</td></tr>\n<tr><th>MESSAGE:</th><td>Internal Server Error</td></tr>\n</table>\n\n</body>\n</html>\n",
        "type": "BadGateway",
        "path": null,
        "invalidValue": null
    }
]
```

with the new protocol it is also not working:

```json
{
    "@context": {
        "edc": "https://w3id.org/edc/v0.0.1/ns/",
        "odrl": "http://www.w3.org/ns/odrl/2/",
[...]

PR #2225, opened 2025-09-04
fix: create JtiValidationPostgresqlMigrationExtension extension

MDSBarbosa:
## WHAT

Creates an extension to apply migrations to the JTI Validation table

## WHY

This extension is required to apply the schema expected as per the upstream Connector [JtiValidationStore](https://github.com/eclipse-edc/Connector/tree/e4eb7eceda2f141d408e116d49b24b00025ac49f/extensions/common/store/sql/jti-validation-store-sql), which is required by the [IdentityAndTrust](https://github.com/eclipse-edc/Connector/blob/e4eb7eceda2f141d408e116d49b24b00025ac49f/extensions/common/iam/identity-trust/identity-trust-core/src/main/java/org/eclipse/edc/iam/identitytrust/core/IdentityAndTrustExtension.java#L157) extension

## FURTHER NOTES

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/2217


rafaelmag110 (approved):
Please create an issue so we don't forget about the fact that the catalog e2e tests didn't catch this.

rafaelmag110:
#2210

PR #2264, opened 2025-09-18
feat: token interceptor

### LINKED ISSUE #2249: Impossible to communicate with earlier EDC versions
## Describe the bug

This issue was discovered while running compatibility tests between versions 0.11.0-RC2 and 0.10.0 when the consumer attempted to query the catalog from the provider. The consumer token could not be verified by the provider.

Starting from **eclipse-edc/Connector** v0.14.0, tokens are created with the prefix `Bearer` and validated accordingly (see [IdentityAndTrustService.java#L123-L135](https://github.com/eclipse-edc/Connector/blob/main/extensions/common/iam/identity-trust/identity-trust-service/src/main/java/org/eclipse/edc/iam/identitytrust/service/IdentityAndTrustService.java#L123-L135)).

This leads to two different exceptions with the same root cause, depending on which version acts as provider and which as consumer:

**Case 1**
**Consumer:** EDC 0.10.0
**Provider:** EDC 0.11.0-RC2

The consumer creates a token **without** the `Bearer` prefix, causing the provider to reject it during JWT validation:
```
[local-control-plane] DEBUG 2025-09-12T15:09:03.424122 Unauthorized: Token is not a Bearer token
[local-control-plane] DEBUG 2025-09-12T15:09:03.424733 DSP: Service call failed: Unauthorized
``` 

**Case 2**
**Consumer:** EDC 0.11.0-RC2
**Provider:** EDC 0.10.0

The consumer creates a token **with** the `Bearer` prefix. The older provider tries to parse it but throws an exception:
```
EVERE 2025-09-12T13:19:20.033458379 JerseyExtension: Unexpected exception caught
org.eclipse.edc.spi.EdcException: java.text.ParseException: Invalid JWS header: Invalid JSON object
	at org.eclipse.edc.iam.identitytrust.core.validation.SelfIssueIdTokenValidationAction.apply(SelfIssueIdTokenValidationAction.java:55)
	at org.eclipse.edc.iam.identitytrust.core.validation.SelfIssueIdTokenValidationAction.apply(SelfIssueIdTokenValidationAction.java:33)
	at org.eclipse.edc.iam.identitytrust.service.IdentityAndTrustService.verifyJwtToken(IdentityAndTrustService.java:126)
	at org.eclipse.edc.connector.controlplane.services.protocol.ProtocolTokenValidatorImpl.verify(ProtocolTokenValidatorImpl.java:62)
	at org.eclipse.edc.connector.controlplane.services.spi.protocol.ProtocolTokenValidator.verify(ProtocolTokenValidator.java:40)
	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.lambda$getDataset$4(CatalogProtocolServiceImpl.java:73)
	at org.eclipse.edc.transaction.local.LocalTransactionContext.execute(LocalTransactionContext.java:74)
	at org.eclipse.edc.connector.controlplane.services.catalog.CatalogProtocolServiceImpl.getDataset(CatalogProtocolServiceImpl.java:73)
```

### To Reproduce

1) Run two EDC instances, one with version 0.11.0-RC2 and one with 0.10.0.
2) Have each instance query the other’s catalog.

### Expected behavior

The provider should successfully parse and validate the consumer’s token.

AndrYurk:
## WHAT

This PR provides interceptors that remove or add the "Bearer " prefix to the authorization header depending on the used DSP protocol.

## WHY

To make communication possible between the old connector (<=v0.10.0) and the new one (>=v0.11.0).

Closes #2249


edc-extensions/token-interceptor/src/main/java/org/eclipse/tractusx/edc/interceptor/AuthRequestFilter.java
```diff
@@ -0,0 +1,45 @@
...
+public class AuthRequestFilter implements ContainerRequestFilter {
+    private static final String BEARER_PREFIX = "Bearer ";
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
+        var requestUri = requestContext.getUriInfo().getRequestUri().toString();
+
+        if (!requestUri.contains(V_2025_1_PATH)) {
+            var authHeader = requestContext.getHeaderString(HttpHeaders.AUTHORIZATION);
+            if (authHeader != null && !authHeader.startsWith(BEARER_PREFIX)) {
```
lgblaumeiser, line 40: That is a bit instable, as the authorization header might have another auth-schema but I think it is safe here. An alternative check would be that the trimmed header still has one space, because we want to catch the case where the header has no auth-schema but only a token, which is a string without any spaces, right? Still not perfect, though

edc-extensions/token-interceptor/src/main/java/org/eclipse/tractusx/edc/interceptor/OkHttpInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+    public OkHttpInterceptor(List<String> skipPaths) {
+        this.skipPaths = skipPaths;
+    }
+
+    @Override
+    public @NotNull Response intercept(Chain chain) throws IOException {
+        Request originalRequest = chain.request();
+        String url = originalRequest.url().toString();
+
+        if (skipPaths.stream().noneMatch(url::contains) && !url.contains(V_2025_1_PATH)) {
```
lgblaumeiser, line 47: This might have an issue with alternative implementations of the connector, I would recommend to at least document the issue, that if alternative implementations use a different mechanism for separating the version implementations, we run into trouble.

edc-extensions/token-interceptor/src/main/resources/META-INF/services/org.eclipse.edc.spi.system.ServiceExtension
```diff
@@ -0,0 +1,21 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations
+#  under the License.
+#
+#  SPDX-License-Identifier: Apache-2.0
+#################################################################################
+
+org.eclipse.tractusx.edc.interceptor.OkHttpInterceptorExtension
+org.eclipse.tractusx.edc.interceptor.ProtocolFilterExtension
```
lgblaumeiser, line 21:
```suggestion
org.eclipse.tractusx.edc.interceptor.ProtocolFilterExtension

```

lgblaumeiser (commented):
Minors, in general ok

Issue #2221, opened 2025-09-04
RightOperand value as a list in policy definition is returned as a quoted string of objects instead of list


akshayadsap:
# Description
When creating a PolicyDefinition where the right operand (odrl:rightOperand)  is a list of strings, the connector accepts the payload, but on retrieval (GET policy), the rightOperand is serialized as a single JSON string that contains the connector’s internal value representation, e.g. "[{@value={chars=BPNPROVIDER, string=BPNPROVIDER, valueType=STRING}}, …]" instead of a JSON array of string literals.
 
## To Reproduce
1. Create a policy definition (management API)
    Endpoint: POST /management/v3/policydefinitions (policydefinitions endpoint)
    Body: 

  [1]
  ```
  {
      "@context": [
          {
              "@vocab": "https://w3id.org/edc/v0.0.1/ns/"
          },
          {
              "edc": "https://w3id.org/edc/v0.0.1/ns/"
          },
          {
              "odrl": "http://www.w3.org/ns/odrl/2/"
          },
          {
              "tx": "https://w3id.org/tractusx/v0.0.1/ns/"
          },
          {
              "cx-policy": "https://w3id.org/catenax/policy/"
          }
      ],
      "@type": "edc:PolicyDefinition",
         "@id": "policy-dc90b627-667f-4bd3-bdb3-ec15737a23d5",
      "edc:policy": {
          "@type": "odrl:Set",
          "odrl:permission": [
              {
                  "@type": "odrl:Permission",
                  "odrl:action": {
                      "@type": "odrl:Action",
                      "@id": "odrl:use"
                  },
                  "odrl:constraint": [{
                      "@type": "odrl:Constraint",
                      "odrl:leftOperand": {
                          "@type": "odrl:LeftOperand",
                          "@id": "tx:BusinessPartnerNumber"
                      },
                      "odrl:operator": {
                          "@type": "odrl:Operator",
                          "@id": "odrl:isAnyOf"
                      },
                      "odrl:rightOperand": ["BPNPROVIDER", "BPNFOSS0000"]
                  },
                  {
                      "@type": "odrl:Constraint",
                      "odrl:leftOperand": {
                          "@type": "odrl:LeftOperand",
                          "@id": "cx-policy:UsagePurpose"
                      },
                      "odrl:operator": {
                          "@type": "odrl:Operator",
                          "@id": "odrl:isAnyOf"
                      },
[...]

PR #2283, opened 2025-09-24
fix: evaluate BPN group constraint with string or array

ronjaquensel:
## WHAT

Updates the evaluation in `BusinessPartnerGroupConstraintFunction` to consider that the right operand may have been created as either a string or an array. If the right operand is created as a string, the set of allowed groups contains the group names as-is, but if the right operand is created as an array, the set of allowed groups contains the group names in the following format: `{@value={chars=<group-name>, string=<group-name>, valueType=STRING}}`.

## WHY

Previously, the evaluation considered the right operand only as string, causing the evaluation to fail if the right operand was created as an array.

Relates to #2221


lgblaumeiser (approved):
lgtm

bmg13 (approved):

PR #2294, opened 2025-09-29
fix: remove problematic versionService from dspversionparams request

### LINKED ISSUE #2277: Connector Discovery throws `Unable to obtain credentials` exception
## Describe the bug

While requesting the Connector Discovery, I receive the following error.

```json
[
  {
    "message": "Error while discovering dsp parameters: org.eclipse.edc.spi.EdcException: Unable to obtain credentials: Scope string invalid: input string was null or empty",
    "type": "Unexpected Result",
    "path": null,
    "invalidValue": null
  }
]
```

### To Reproduce

Request the EDC requesting parameters as follows:

```http
POST https://{{hostname}}/management/v4alpha/connectordiscovery/dspversionparams
X-Api-Key: {{apiKey}}
Content-Type: {{contentType}}

{
    "@context": {
        "tx": "https://w3id.org/tractusx/v0.0.1/ns/",
        "edc": "https://w3id.org/edc/v0.0.1/ns/"
    },
    "@type": "tx:ConnectorParamsDiscoveryRequest",
    "tx:bpnl": "{{counterPartyBPN}}",
    "edc:counterPartyAddress": "{{counterPartyAddress}}"
}
```

### Expected behavior

It should return the corresponding EDC request parameters.

```json
{
    "@context": {
        "edc": "https://w3id.org/edc/v0.0.1/ns/",
        "tx": "https://w3id.org/tractusx/v0.0.1/ns/"
    },
    "edc:counterPartyId": "did:web:one-example.com",
    "edc:counterPartyAddress": "https://provider.domain.com/api/dsp/2025-1",
    "edc:protocol": "dataspace-protocol-http:2025-1"
}
```

## Context Information

- Used version: v0.11.0-RC2

rafaelmag110:
## WHAT

Replaces dsp version request delegation from `versionService` to direct http call.

## WHY

The `versionService` will dispatch the request through the dsp dispatcher, which requires the request to be decorated with a scope. In the case of a version request, no scope exists, hence the request will fail.

Also, the version service has been removed from the main branch on the upstream connector repo, so this its not good to rely on that.

## FURTHER NOTES

The API previously responded with 502 Bad Gateway in cases where the metadata endpoint failed to response. This is no longer the case as the API will now always respond with 500 - in case of errors. However, the status code and a message will be passed on to the client for troubleshooting.

Closes #2277

### DIFF
--- a/edc-extensions/connector-discovery/connector-discovery-api/src/main/java/org/eclipse/tractusx/edc/discovery/v4alpha/service/ConnectorDiscoveryServiceImpl.java
+++ b/edc-extensions/connector-discovery/connector-discovery-api/src/main/java/org/eclipse/tractusx/edc/discovery/v4alpha/service/ConnectorDiscoveryServiceImpl.java
@@ -41,7 +41,7 @@ import java.io.IOException;
 
 public class ConnectorDiscoveryServiceImpl implements ConnectorDiscoveryService {
 
-    private final static String DSP_DISCOVERY_PATH = "/.well-known/dspace-version";
+    private static final String DSP_DISCOVERY_PATH = "/.well-known/dspace-version";
 
     private final BdrsClient bdrsClient;
     private final EdcHttpClient httpClient;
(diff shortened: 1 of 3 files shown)


lgblaumeiser (approved):
lgtm

PR #2303, opened 2025-10-09
docs: adds connector discovery walkthrough

### LINKED ISSUE #2219: Management API walkthrough contains not up to date docs
## Describe the bug

With the introduction of the CX Policy standard, and the tractusx-edc policy validation mechanism, the management API walkthrough has documentation that is not up to date.

Example: https://github.com/eclipse-tractusx/tractusx-edc/blob/main/docs/usage/management-api-walkthrough/02_policies.md

These docs should be revisited and the contained policies adapted according to the standard.

rafaelmag110:
## WHAT

Adds a page for connector version params discovery.

## WHY

The feature was introduced but documentation was missing.

Closes #2219

### DIFF
--- a/docs/usage/management-api-walkthrough/04_catalog.md
+++ b/docs/usage/management-api-walkthrough/04_catalog.md
@@ -233,7 +233,7 @@ policies included.
 
 
 ## Reference
-- [Connector Discovery API](https://eclipse-tractusx.github.io/tractusx-edc/openapi/control-plane-api/0.11.0/#/Connector%20Discovery/discoverDspVersionParamsV4Alpha)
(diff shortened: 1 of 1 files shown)


rafaelmag110:
I'm fine either way and don't oppose to adding it to the Catalog request page as that is the entry point for the rest.

rafaelmag110:
@lgblaumeiser
Please review.
I moved the walkthrough to the catalog one and rewrote so the whole thing makes more sense.

AndrYurk (commented):
One note regarding the use of a strict API version reference.

lgblaumeiser (approved):
lgtm

PR #2355, opened 2025-10-29
feat: removing trailing slash for dsp 0.8

### LINKED ISSUE #2345: DSP version params returns trailing slash for counter parties who only implement the dataspace-protocol-http
## Describe the bug

The `counterPartyAddress` in the response of the DSP Version Params contains a trailing slash but only for the `dataspace-protocol-http` and not the `dataspace-protocol-http:2025-1`. When using this `counterPartyAddress` in other calls this leads to an `Ambiguous URI empty segment` error.

### To Reproduce

Make a DSP Version Params request to a Tractus-X EDC with version lower than 11.

### Expected behavior

The returned `counterPartyAddress` does not contain a trailing slash.

AndrYurk:
## WHAT

This PR removes the trailing slash for dsp 0.8, as it is redundant

## WHY

Make a DSP Version Params request to a Tractus-X EDC with a version lower than 11.

Closes #2345

### DIFF
--- a/edc-extensions/connector-discovery/connector-discovery-api/src/main/java/org/eclipse/tractusx/edc/discovery/v4alpha/service/ConnectorDiscoveryServiceImpl.java
+++ b/edc-extensions/connector-discovery/connector-discovery-api/src/main/java/org/eclipse/tractusx/edc/discovery/v4alpha/service/ConnectorDiscoveryServiceImpl.java
@@ -102,9 +102,8 @@ public class ConnectorDiscoveryServiceImpl implements ConnectorDiscoveryService 
                 "not support any of the expected protocol versions (" + Dsp08Constants.V_08_VERSION + ", " + Dsp2025Constants.V_2025_1_VERSION + ")");
     }
 
-    private String removeTrailingSlash(String path)
-    {
-        if(path.endsWith("/")) {
+    private String removeTrailingSlash(String path) {
+        if (path.endsWith("/")) {
             return path.substring(0, path.length() - 1);
         }
         return path;


arnoweiss:
The docs say:
https://github.com/eclipse-tractusx/tractusx-edc/blob/64a60cec3a8f276ab4d1df1068b663976b0f962a/docs/usage/management-api-walkthrough/04_catalog.md?plain=1#L58-L59

This simply isn't true currently - at least if the trailing slash is returned from connector discovery AND the management api does not heal potential ambiguous uri segments. I'm not sure however that the PR catches the case where EDC discovery returns an address with the trailing slash.

lgblaumeiser:
> The docs say:
> 
> https://github.com/eclipse-tractusx/tractusx-edc/blob/64a60cec3a8f276ab4d1df1068b663976b0f962a/docs/usage/management-api-walkthrough/04_catalog.md?plain=1#L58-L59
> 
> This simply isn't true currently - at least if the trailing slash is returned from connector discovery AND the management api does not heal potential ambiguous uri segments. I'm not sure however that the PR catches the case where EDC discovery returns an address with the trailing slash.

You are right, the issue is actually, that @ronjaquensel and I discovered some inconsistencies in the constants defined in upstream for the different version base paths. I absolutely agree that we have to make this more consistent, but I think, this topic needs some more analysis, there is a potential, that the current change make the situation more inconsistent instead of less.

AndrYurk:
@lgblaumeiser , the main problem here is that upstream `.well-known/dspace-version` for DSP version `0.8` returns `"path": "/"`.

```
{
    "protocolVersions": [
        {
            "version": "v0.8",
            "path": "/",
            "binding": "HTTPS"
        },
        {
            "version": "2025-1",
            "path": "/2025-1",
            "binding": "HTTPS"
        },
        {
            "version": "2024/1",
            "path": "/2024/1",
            "binding": "HTTPS"
        }
    ]
}
```

That is actually incorrect behaviour according to https://catenax-ev.github.io/docs/next/standards/CX-0018-DataspaceConnectivity#27-protocol-version-handling. 

```
{
  "protocolVersions": [
    {
      "version": "2025-1",
      "path": "/2025-1",
      "binding": "HTTPS"
    },
    {
      "version": "0.8",
      "path": "",
      "binding": "HTTPS"
    }
  ]
}
```

However, it is much easier and faster to have the fix on our side, as it does not make sense to add a path that we are expecting to be empty.

PR #2400, opened 2025-11-18
infra: Update to jre lts version 25

### LINKED ISSUE #2401: Update base image to use latest lts version
Update the docker base image used for multiple reasons:
- There are a number of security issues related to openssl which I assume is due to the base image
- JRE 24 is not a lts version, so the update of new images should end soon
- We wanted to make use of latest JDK technology to get along with improvements in the execution environment

lgblaumeiser:
## WHAT

Update of the base image in the dockerfile

## WHY

The previously used java 24 version is not lts, we have a multitude of security issues in openssl which I assume are coming with the base image. As it is not expectable that there will be many update to Java 24, it only makes sense to update to the latest lts version.

This follows also the former approach to make use of improvements in the java versions concerning jit compiling and performance in general.

Closes #2401

### DIFF
--- a/.github/actions/setup-java/action.yml
+++ b/.github/actions/setup-java/action.yml
@@ -20,15 +20,15 @@
 
 
 ---
-name: "Setup JDK 17"
-description: "Setup JDK 17"
+name: "Setup JDK"
+description: "Setup JDK"
 runs:
   using: "composite"
   steps:
-    - name: Setup JDK 17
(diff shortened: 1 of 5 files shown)


lgblaumeiser:
@AndrYurk I have done the updates, for build I switched to 21 as this is the last lts version that is supported by the gradle version we use.

PR #2423, opened 2025-11-24
fix: scope extraction for legacy policy constraints

### LINKED ISSUE #2421: 0.11.x connector cannot properly negotiate for contracts provided by a Jupiter connector
## Describe the bug

The issue is, that a consumer has to identify the scope for the STS token from the usage policy that is provided in the contract offer. As the left operands for the policies have changed completely, the extraction has to be aware that the legacy policies have to be evaluated as well. Obviously, this is not completely the case.

### To Reproduce

Setup a system with an older connector as provider and a newer connector as consumer. Create a Jupiter contract definition and execute the DSP dance between the two connectors. The negotiation will fail because the usage policy cannot be fulfilled.

### Expected behavior

The negotiation should be successful

- Used version: 0.10.x or earlier as provider, 0.11.1 as consumer

## Possible Implementation

Allow old policy left operands as well for scope extraction for the STS token

- Fix the issue -> @ronjaquensel 
- Cherrypick the fix for 0.11.2 -> @lgblaumeiser 
- Check the adaptation of the compatibility tests -> tbd

ronjaquensel:
## WHAT

Adds support for legacy policy consraints in the `CredentialScopeExtractor`, as previously they were missing. Thus, for Jupiter policies, no scopes were extracted from any policy and no VCs requested/provided.

## WHY

backward compatibility

## FURTHER NOTES

Cleaned up the legacy credential types. Quite some test cases were affected which were heavily using the `PcfCredential`. These now use `MembershipCredential` and `DataExchangeGovernanceCredential`. Also renamed the `FrameworkAgreementCredentialConstraintFunction` to `FrameworkAgreementConstraintFunction`, as there is no `FrameworkAgreementCredential`.

Closes #2421


edc-extensions/dcp/tx-dcp/src/main/java/org/eclipse/tractusx/edc/iam/iatp/scope/CredentialScopeExtractor.java
lgblaumeiser: This could be reduced to 
```suggestion
return (leftOperand.equals(FRAMEWORK_CREDENTIAL_PREFIX)) ?
  "DateExchangeGovernance" :
  leftOperand;
```

lgblaumeiser (changes requested):
As discussed, I persisted the outcomes

PR #2426, opened 2025-11-25
fix: Make a draft release with a failed dependency check

### LINKED ISSUE #2365: Release process blocks bugfix releases
## Describe the bug

The release process requires two conditions that in certain situations block a successful draft release run for creating a bugfix.

Condition 1: Bugfix releases have to be started from a tag
Condition 2: The draft release process requires a successful strict dependency check to finish with a prepared bugfix branch.

Situation: Between the creation of 0.10.1 and 0.10.2 the dependency io.opentelemetry.javaagent:opentelemetry-javaagent:2.16.0 changed status from approved to restricted. This leads to a unsuccessful dependency check which prevents the creation of a new branch to fix the issue. But the tag is the tag, so the failing condition cannot be corrected in the old branch as well.

### To Reproduce

Happened for the above described dependency

### Expected behavior

A bugfix release branch is created even if there is a dependency check with the old version. Still a release should be prevented with restricted dependencies

## Context Information

- Used version: 212c6ba6f539fa3719de567ded92480d763d8cf8 resp. 0.10.1 tag
- OS: Github action "Draft Release"

## Possible Implementation

Move the dependency check from the draft release to the release workflow. Ensure that this is done for older releases as well, as we might want to do a bugfix for 0.9 or 0.10 throughout this year.

AndrYurk:
## WHAT

This PR allows creating a branch for draft-release even if the dependency check fails.

## WHY
To make it possible to fix the dependency for a new release

## FURTHER NOTES

Workflow `generate-and-publish-dependencies.yaml` was extended to run also in a `release/*` branches.

Closes #2365

### DIFF
--- a/.github/workflows/draft-release.yaml
+++ b/.github/workflows/draft-release.yaml
@@ -110,12 +110,6 @@ jobs:
       - name: Create Release branch
         run: git checkout -b ${{ needs.validate-and-prepare.outputs.branch_name }}
       - uses: ./.github/actions/setup-java
-      - name: Check dependencies before release
-        uses: ./.github/actions/generate-and-check-dependencies
-        with:
-          run: ${{ needs.validate-and-prepare.outputs.is_official_release == true && 'strict' || 'standard' }}
-      - name: Replace published DEPENDENCIES file link in NOTICE with the one just created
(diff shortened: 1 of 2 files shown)


lgblaumeiser (approved):
lgtm

PR #2432, opened 2025-11-27
fix: remove asSequence call on downloadOpenapi

### LINKED ISSUE #2429: downloadOPenAPI does not work anymore
## Describe the bug

The downloadOpenAPI task used in the PublishNewSnapshot workflow stopped working with a ConcurrentModificationException. Could be, that this build lately used Gradle 9.21 instead of a 8.x version.

### To Reproduce

Check https://github.com/eclipse-tractusx/tractusx-edc/actions/runs/19710819677

### Expected behavior

The task executes properly

## Context Information

- Github Actions Workflow

ndr-brt:
## WHAT

removes the spurious `.asSequence` call in `downloadOpenapi` task

## WHY

it was causing concurrent exceptions since gradle 9

## FURTHER NOTES
- replaced `map/flatten` with `flatMap` calls. same behavior, less verbose

Closes #2429

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -207,10 +207,9 @@ subprojects {
                 .resolve("docs").resolve("openapi")
 
             configurations.asMap.values
-                .asSequence()
                 .filter { it.isCanBeResolved }
-                .map { it.resolvedConfiguration.firstLevelModuleDependencies }.flatten()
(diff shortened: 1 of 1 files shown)


lgblaumeiser (commented):
lgtm, thanks @ndr-brt

lgblaumeiser (approved):

Issue #2309, opened 2025-10-13
Auto-Register DataService in did-document on startup


arnoweiss:
## WHAT
It's a best practice to advertise the DSP-endpoint as `DataService` in the did document of a Participant. EDC should be able to automatically add that entry on startup.

## WHY
Ease of integration

## HOW
Since this is likely dependent on the [DID Service](https://eclipse-dataspace-dcp.github.io/decentralized-claims-protocol/v1.0/#dfn-dids) implementation (not standardized), it might be necessary to do have separate extensions for each implementation. However, if the `CredentialService` and DID Service are one endpoint, EDC already has all the relevant coordinates and credentials to implement this. 

## FURTHER NOTES
Should be possible to switch off.

arnoweiss:
We agreed in an earlier committer meeting that this issue does not require an explicit DR. I will create one throughout the day.

PR #2433, opened 2025-11-27
docs: add did registration DR

arnoweiss:
## WHAT

Introduces a DR

## WHY

Introduces a significant new feature

## FURTHER NOTES

Relates #2309

### DIFF
--- a/docs/development/decision-records/2025-11-27-did-service-registration/README.md
+++ b/docs/development/decision-records/2025-11-27-did-service-registration/README.md
@@ -22,10 +22,10 @@ to avoid creating duplicate `service` entries and manage itself.
 
 1. Introduce configuration options in application and helm chart.
 2. Create a new SPI including an interface that represents the feature in an abstract manner.
-3. Implement the interface as client for [SAP DIV's write endpoint to the did document](https://api.sap.com/api/DIV/path/CompanyIdentityV2HttpController_updateCompanyIdentity_v2.0.0).
-4. The extension will perform self-cleanup in case the EDC is deprovisioned.
+3. An extension that implements the SPI's interface as client for [SAP DIV's write endpoint to the did document](https://api.sap.com/api/DIV/path/CompanyIdentityV2HttpController_updateCompanyIdentity_v2.0.0).
+4. Another extension that will implement the lifecycle management logic.
 
-The resulting logic shall look like:
+The lifecycle management logic shall look like:
 
 ```mermaid
 flowchart TD


docs/development/decision-records/2025-11-27-did-service-registration/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Decision
+
+The controlplane will be enabled to register itself as `DataService` with the participant's did document. There will be
+configuration variables to enable the feature, set an id for the DSP endpoint and point to the DID service's write-APIs.
+There will not be an additional endpoint on the Management API - this logic is purely internal.
+
+## Rationale
+
+Standard CX-0001 currently describes the predominant method for discovering DSP endpoints. It is a centralized service
```
lgblaumeiser: That rationale is not right, that part of CX-0001 is already deprecated. The rationale is to support CX-0018 section 2.6.

docs/development/decision-records/2025-11-27-did-service-registration/README.md
```diff
@@ -0,0 +1,40 @@
...
+Managing these `service` entries for DSP endpoints can become a chore: hosts may change, deployments may be 
+deprovisioned. That's why there should be a solution that is extensible for each wallet implementation and smart enough 
+to avoid creating duplicate `service` entries and manage itself.
+
+## Approach
+
+1. Introduce configuration options in application and helm chart.
+2. Create a new SPI including an interface that represents the feature in an abstract manner.
+3. Implement the interface as client for [SAP DIV's write endpoint to the did document](https://api.sap.com/api/DIV/path/CompanyIdentityV2HttpController_updateCompanyIdentity_v2.0.0).
+4. The extension will perform self-cleanup in
```
lgblaumeiser: Seems to be something missing

docs/development/decision-records/2025-11-27-did-service-registration/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+```java
+
+public interface DidServiceClient {
+
+    void createService(String id, String urlOfWellKnown);
+    
+    void updateService(String id, String urlOfWellKnown);
+    
+    void deleteService(String id);
```
lgblaumeiser: I am still struggling with an automation for this one. Could you elaborate on the mechanism?

docs/development/decision-records/2025-11-27-did-service-registration/README.md
```diff
@@ -0,0 +1,40 @@
...
+configuration variables to enable the feature, set an id for the DSP endpoint and point to the DID service's write-APIs.
+There will not be an additional endpoint on the Management API - this logic is purely internal.
+
+## Rationale
+
+Standard CX-0001 currently describes the predominant method for discovering DSP endpoints. It is a centralized service
+that is assumed to be a singleton. DIDs provide all means to map from a business partner identifier to a DSP endpoint.
+How that can be achieved is described in [DSP section 4](https://eclipse-dataspace-protocol-base.github.io/DataspaceProtocol/2025-1-err1/#discovery-of-service-endpoints)
+
+Managing these `service` entries for DSP endpoints can become a chore: hosts may change, deployments may be
```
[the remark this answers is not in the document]
arnoweiss: added that part

pratapipatelbcone:
@lgblaumeiser, we need to resolve did first when connector starts to check if service entry with correct url exists or not (as shown in flow chart).
TO resolve did, we can use existing [DidResolverRegistryImpl.java](https://github.com/eclipse-edc/Connector/blob/main/extensions/common/iam/decentralized-identity/identity-did-core/src/main/java/org/eclipse/edc/iam/did/resolution/DidResolverRegistryImpl.java) but there is a caveat, since it caches did document, so we may not get a updated did document.

To get a fresh copy, we need to modify the [DidResolverRegistry.java](https://github.com/eclipse-edc/Connector/blob/main/spi/common/identity-did-spi/src/main/java/org/eclipse/edc/iam/did/spi/resolution/DidResolverRegistry.java) interface to have a method which returns the updated data instead of a cached one.
```
public interface DidResolverRegistry {
    
    // existing method to resolve did
    Result<DidDocument> resolve(String didKey);
    
    // new method to resolve updated did
    Result<DidDocument> resolve(String didKey, boolean forceRefresh);
}
```

SO, I have two question here.
1. Do you think it is the right approach?
2. Do you need to get this updated in Decision record as well?

PR #2446, opened 2025-12-05
[Fix] Update Contract Request Message with missing ODRL Context

gerbigf:
## WHAT
- Remove duplicate odrl permission and obligation section
- Added ODRL Context
- Made policy more generic. 

## WHY

Wrong information in the text. 

## FURTHER NOTES

Cleanup of docs.

### DIFF
--- a/docs/usage/management-api-walkthrough/05_contractnegotiations.md
+++ b/docs/usage/management-api-walkthrough/05_contractnegotiations.md
@@ -30,7 +30,6 @@ Content-Type: application/json
   "counterPartyAddress": "https://provider-control.plane/api/v1/dsp/2025-1",
   "protocol": "dataspace-protocol-http:2025-1",
   "policy": {
-    "@context": "http://www.w3.org/ns/odrl.jsonld",
     "@type": "Offer",
     "@id": "{{OFFER_ID}}",
     "target": "{{ASSET_ID}}",


docs/usage/management-api-walkthrough/05_contractnegotiations.md
```diff
@@ -30,22 +30,14 @@ Content-Type: application/json
   "counterPartyAddress": "https://provider-control.plane/api/v1/dsp/2025-1",
   "protocol": "dataspace-protocol-http:2025-1",
   "policy": {
+    "@context": "http://www.w3.org/ns/odrl.jsonld",
```
lgblaumeiser: Should be the new odrl context `https://w3id.org/catenax/2025/9/policy/odrl.jsonld`, right?

gerbigf: Actually, it works even without the @context, which is also according to the DSP spec: https://eclipse-dataspace-protocol-base.github.io/DataspaceProtocol/2025-1/#catalog-protocol:~:text=string-,MessageOffer,-%40id

gerbigf: > Actually, it works even without the @context, which is also according to the DSP spec: https://eclipse-dataspace-protocol-base.github.io/DataspaceProtocol/2025-1/#catalog-protocol:~:text=string-,MessageOffer,-%40id

This here works, too: 
image

The difference being that our example uses "ContractRequest" and DSP in it's examples uses "ContractRequestMessage"... Both generate a valid contract agreement for me.

PR #2456, opened 2025-12-10
feat: Did Document Service Client SPI and DIV Implementation

pratapipatelbcone:
## WHAT
- Did Document Service Client SPI
- SAP DIV Specific Implementation module for Did Document Service Client SPI

## WHY
To Advertise custom urls of any type via DID Document

## FURTHER NOTES
See this DR for more info: https://github.com/eclipse-tractusx/tractusx-edc/pull/2433
> Self Registration of `DataService` at connector control plane startup will be in next PR.

Relates #2309

### DIFF
--- a/edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
+++ b/edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
@@ -103,7 +103,8 @@ public class DidDocumentServiceDimClient implements DidDocumentServiceClient {
     @Override
     public ServiceResult<Void> update(Service service) {
         return validateService(service)
-                .compose(v -> deleteById(service.getId()))
+                .compose(v -> deleteServiceEntry(service.getId()))
+                .compose(v -> updatePatchStatus())
                 .compose(v -> createServiceEntry(service))
                 .compose(v -> updatePatchStatus())
                 .onSuccess(v -> monitor.info("Updated service entry %s in DID Document".formatted(asString(service))))
@@ -404,6 +405,6 @@ public class DidDocumentServiceDimClient implements DidDocumentServiceClient {
     }
 
     private boolean isBlank(String str) {
-        return str == null || str.trim().isEmpty();
+        return str == null || str.isBlank();
     }
 }
--- a/edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
+++ b/edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
@@ -27,10 +27,9 @@ import org.eclipse.edc.spi.monitor.Monitor;
 import org.eclipse.edc.spi.system.ServiceExtension;
 import org.eclipse.edc.spi.system.ServiceExtensionContext;
 import org.eclipse.edc.spi.types.TypeManager;
+import org.eclipse.tractusx.edc.core.utils.PathUtils;
 import org.eclipse.tractusx.edc.iam.dcp.sts.dim.oauth.DimOauth2Client;
 import org.eclipse.tractusx.edc.spi.did.document.service.DidDocumentServiceClient;
-
-import java.util.Optional;
 
 @Provides(DidDocumentServiceClient.class)
 public class DidDocumentServiceDimClientExtension implements ServiceExtension {
@@ -48,7 +47,7 @@ public class DidDocumentServiceDimClientExtension implements ServiceExtension {
     private Monitor monitor;
 
     @Setting(key = "tx.edc.iam.sts.dim.url", description = "STS Dim endpoint", required = false)
-    private String dimUrlConfig;
+    private String dimUrl;
 
     @Setting(key = "edc.participant.id", description = "EDC Participant Id")
     private String ownDid;
@@ -56,17 +55,17 @@ public class DidDocumentServiceDimClientExtension implements ServiceExtension {
     @Override
     public void initialize(ServiceExtensionContext context) {
 
-        Optional.ofNullable(dimUrlConfig)
-                .map(dimUrl -> new DidDocumentServiceDimClient(
-                        httpClient,
-                        dimOauth2Client,
-                        typeManager.getMapper(),
-                        dimUrl,
-                        ownDid,
-                        monitor)
-                ).ifPresentOrElse(client ->
-                                context.registerService(DidDocumentServiceClient.class, client),
-                        () -> monitor.info("DIM Url not configured, DidDocumentServiceDIMClient will not be registered")
-                );
+        if (dimUrl == null || dimUrl.isBlank() || dimOauth2Client == null) {
+            monitor.info("DidDocumentServiceDIMClient will not be registered because DIM URL not configured or an implementation of DimOauth2Client is missing");
+        } else {
+            var client = new DidDocumentServiceDimClient(
+                    httpClient,
+                    dimOauth2Client,
+                    typeManager.getMapper(),
+                    PathUtils.removeTrailingSlash(dimUrl),
+                    ownDid,
+                    monitor);
+            context.registerService(DidDocumentServiceClient.class, client);
+        }
     }
 }
--- a/edc-extensions/did-document/did-document-service-dim/README.md
+++ b/edc-extensions/did-document/did-document-service-dim/README.md
@@ -3,6 +3,7 @@
 ## Overview
 This extension provides a client for managing DID Document Service entries using the DIM (Decentralized Identity Management) API.
 It enables secure and programmatic updates to DID Documents for organizations using the TractusX ecosystem.
+The client's purpose is to be injected in the `did-document-service-self-registration` extension by implementing the `DidDocumentServiceClient` SPI.
 
 ## API Details
 
--- a/edc-extensions/did-document/did-document-service-dim/build.gradle.kts
+++ b/edc-extensions/did-document/did-document-service-dim/build.gradle.kts
@@ -25,6 +25,7 @@ plugins {
 dependencies {
     implementation(project(":spi:did-document-service-spi"))
     implementation(project(":edc-extensions:dcp:tx-dcp-sts-dim"))
+    implementation(project(":core:core-utils"))
     implementation(libs.edc.runtime.metamodel)
     implementation(libs.edc.spi.identity.did)
     implementation(libs.edc.spi.http)


edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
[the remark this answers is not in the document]
pratapipatelbcone: Done

arnoweiss (approved):

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,352 @@
...
+        }
+        return deleteServiceEntry(id)
+                .compose(v -> updatePatchStatus())
+                .onSuccess(v -> monitor.info("Deleted service entry %s in DID Document".formatted(id)))
+                .onFailure(f -> monitor.warning("Failed to delete service entry %s with failure %s".formatted(id, f.getFailureDetail())));
+    }
+
+    @Override
+    public ServiceResult<List<Service>> findAll() {
+        return ServiceResult.from(resolverRegistry.resolve(ownDid).map(DidDocument::getService));
```
lgblaumeiser: This might fail in the current setup, as this will map not to the DID document in the SAP DIM and might lag behind concerning updates. I think, we should be a bit careful here, if the DID does not point to a SAP DIM, it should stop, right?

arnoweiss (approved):
Thanks for the update - it's better to give the service's `id` at deployment time rather than letting the did service do it.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
```diff
@@ -0,0 +1,77 @@
...
+                        resolverRegistry,
+                        httpClient,
+                        dimOauth2Client,
+                        typeManager.getMapper(),
+                        dimUrl,
+                        ownDid,
+                        monitor)
+                ).ifPresentOrElse(client ->
+                                context.registerService(DidDocumentServiceClient.class, client),
+                        () -> monitor.warning("DIM Url not configured, DidDocumentServiceDIMClient will not be registered")
```
lgblaumeiser: That is either an info or an error. If the missing url is the indicator, that the service is not used, than it is an info, if an explicit setting defines that the service has to be used, the misconfiguration of this is an error.

lgblaumeiser (changes requested):
In general, this code is really hard to understand because the functionality is spread in so many different functions, so e.g, there is the create method which calls createServiceEntry, followed by an updatePatchStatus, which is doing nothing except calling. The responses are typically hidden, and there is no clue what happens, if the create succeeds but the updatePatchStatus fails.

createServiceEntry, then again calls patchRequest and a payload creation method. These methods are then scattered throughout the whole file and there is no place where I can easily understand what is going on.

Using the fluent metaphor does not help either to simplify readability.

I also think, that the whole thing is quite inefficient, because it does so many calls to the api, which might not be necessary, if the error handling on the DIV side works, e.g., if I understand what the api returns in case of a deletion of something that does not exist, I do not have to check, whether it exists in the first place, the reaction can be handled gracefully and I reduce the amount of calls to one. The same applies to create and update.

In addition, I suggest to implement the lifecycle management component first, as I fear a lot of redundancy, as in order to distinguish between create and update, you have to check whether the entry exists or not. But as update is not available and boils down to delete and create, there is no actual need to distinguish between the two, as you can always do the update. The deletion of a non-existent entry can easily be handled gracefully.

So in total, this code is way to complex and it is developed from the wrong end, as the life cycle component determines the needed calls which then can be implemented

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
pratapipatelbcone: This should map to SAP DIM, a DID provided by SAP DIM successfully resolves to a DID via `resolverRegistry`.
Also, if a `dimUrl` is not configured, this service registration will be skipped via Self Registration Extension.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
```diff
@@ -0,0 +1,77 @@
...
+    @Inject
+    private DimOauth2Client dimOauth2Client;
+
+    @Inject
+    private TypeManager typeManager;
+
+    @Inject
+    private Monitor monitor;
+
+    @Setting(key = "tx.edc.iam.sts.dim.url", description = "STS Dim endpoint", required = false)
```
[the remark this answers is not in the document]
pratapipatelbcone: No, updating did requires wallet endpoint, SAP DIM in this case.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,352 @@
...
+            return ServiceResult.conflict("%s already exists".formatted(asString(existingService.getContent())));
+        }
+        return createServiceEntry(service)
+                .compose(v -> updatePatchStatus())
+                .onSuccess(v -> monitor.info("Created service entry %s in DID Document".formatted(asString(service))))
+                .onFailure(f -> monitor.warning("Failed to create service entry %s with failure %s".formatted(asString(service), f.getFailureDetail())));
+    }
+
+    @Override
+    public ServiceResult<Void> update(Service service) {
```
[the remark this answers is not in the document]
pratapipatelbcone: As suggested by you.
1. SPI and self registration extension PR got merged.
2. Removed unused method from SPI
3. Removed unused implementation from DIM extension.

edc-controlplane/edc-controlplane-base/build.gradle.kts
```diff
@@ -62,6 +62,7 @@ dependencies {
     implementation(project(":edc-extensions:token-interceptor"))
     implementation(project(":edc-extensions:event-subscriber"))
     implementation(project(":edc-extensions:did-document:did-document-service-self-registration"))
+    implementation(project(":edc-extensions:did-document:tx-did-document-service-dim"))
```
lgblaumeiser: Is there a good reason to prefix the module name with tx-? I do not see any value in that, as the group name already implies that.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+ *         "id": "did:web:example.com:123#DataService",
+ *         "serviceEndpoint": "https://edc.com/edc/.well-known/dspace-version",
+ *         "type": "DataService"
+ *       }
+ *     ]
+ *   }
+ * }
+ * }
+ * </pre>
+ * 2. A subsequent PATCH request is sent to the /status endpoint to finalize the update.
```
lgblaumeiser: Can you please add a short note, what is happening, because a patch with an empty body is in need of an explanation.

lgblaumeiser (changes requested):
Some improvements, but the PR is in a much better shape.

In general I am not a big fan of this fluent, functional style, as I think it makes things harder to read. I would appreciate, if you could bring together the private helper methods to the corresponding calling functions, as this would improve readability. E.g., all the payloads are basically generated for exactly one call, so the payload creation method should not be at the end, but near the place where it is used.

Perhaps you can group the create methods and the delete methods as big blocks, and leave this idea of having the public methods at the top.

lgblaumeiser:
Hi @pratapipatelbcone, obviously there is something broken, as the runtime seems to not come up. Please check

pratapipatelbcone:
> Hi @pratapipatelbcone, obviously there is something broken, as the runtime seems to not come up. Please check

Yes, I am looking into it.
I was able to fix few stuff, but e2e is failing locally on `main` branch as well, which is making it hard for me to verify if everything looks good now.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+     * @return a Result containing the response body as a string if successful, or a failure message
+     */
+    Result<String> handlePatchStatusResponse(Response response) {
+        if (!response.isSuccessful()) {
+            return Result.failure("Failed to update patch status request with status code: %d, message %s".formatted(response.code(), response.message()));
+        }
+        try {
+            var body = Objects.requireNonNull(response.body()).string();
+            var parsedBody = mapper.readTree(body);
+            return Optional.ofNullable(parsedBody.get("status").asText())
```
AndrYurk: Possible NPE, use `parsedBody.path("status").asText(null)`

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+     */
+    Result<String> handleDidUpdateResponse(Response response) {
+        if (!response.isSuccessful()) {
+            return Result.failure("DID update request failed with status code: %d, message: %s".formatted(response.code(), response.message()));
+        }
+        try {
+            var body = Objects.requireNonNull(response.body()).string();
+            var parsedBody = mapper.readTree(body);
+            return Optional.ofNullable(parsedBody.get("updateDidRequest"))
+                    .map(updateDidRequest -> updateDidRequest.get("success").asBoolean())
```
AndrYurk: Possible NPE, use `.path()` instead of `.get()` and `asBoolean` with default value

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+                    .filter(Boolean.TRUE::equals)
+                    .map(success -> Result.success(body))
+                    .orElseGet(() -> Result.failure("Failed to Update Did Document, res: %s".formatted(body)));
+        } catch (IOException e) {
+            monitor.severe("Failed to parse did update response from DIM");
```
AndrYurk:
```suggestion
            monitor.severe("Failed to parse did update response from DIM", e);
```

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+        try {
+            var body = Objects.requireNonNull(response.body()).string();
+            var parsedBody = mapper.readTree(body);
+            return Optional.of(parsedBody)
+                    .filter(data -> data.get("count").asInt() == 1)
+                    .map(data -> data.get("data").get(0).get("id").asText())
+                    .map(Result::success)
+                    .orElseGet(() -> Result.failure("Failed to Resolve Company Identity Response, res: %s".formatted(body)));
+        } catch (IOException e) {
+            monitor.severe("Failed to parse company identity response from DIM");
```
AndrYurk:
```suggestion
            monitor.severe("Failed to parse company identity response from DIM", e);
```

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+     */
+    Result<String> handleCompanyIdentityResponse(Response response) {
+        if (!response.isSuccessful()) {
+            return Result.failure("Company identity resolution request failed with status code: %d, message: %s".formatted(response.code(), response.message()));
+        }
+        try {
+            var body = Objects.requireNonNull(response.body()).string();
+            var parsedBody = mapper.readTree(body);
+            return Optional.of(parsedBody)
+                    .filter(data -> data.get("count").asInt() == 1)
```
AndrYurk:
```suggestion
                    .filter(data -> data.path("count").asInt(0) == 1)
```

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,381 @@
...
+    private final AtomicReference<String> companyIdentity = new AtomicReference<>();
+
+    public DidDocumentServiceDimClient(EdcHttpClient httpClient,
+                                       DimOauth2Client dimOauth2Client, ObjectMapper mapper, String dimUrl, String ownDid, Monitor monitor) {
+        this.httpClient = httpClient;
+        this.dimOauth2Client = dimOauth2Client;
+        this.mapper = mapper;
+        this.ownDid = ownDid;
+        this.monitor = monitor.withPrefix(getClass().getSimpleName());
+        this.didDocApiUrl = "%s/%s".formatted(dimUrl, DID_DOC_API_PATH);
```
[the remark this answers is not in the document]
pratapipatelbcone: Thanks for catching it. Will update it.

edc-extensions/did-document/tx-did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
arnoweiss: Again - that's how the API is designed. Will be explained in README.

pratapipatelbcone: It's how the way API works, we need to make a request to update the did document.
Did resolver gets the updated did document but the operation remains in pending state. And any subsequent did document fails, until `/status` endpoint is executed to update the operation state as successful.

lgblaumeiser (changes requested):
Small change I propose

edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClientExtension.java
```diff
@@ -0,0 +1,72 @@
...
+    @Setting(key = "tx.edc.iam.sts.dim.url", description = "STS Dim endpoint", required = false)
+    private String dimUrlConfig;
+
+    @Setting(key = "edc.participant.id", description = "EDC Participant Id")
+    private String ownDid;
+
+    @Override
+    public void initialize(ServiceExtensionContext context) {
+
+        Optional.ofNullable(dimUrlConfig)
```
AndrYurk:
```suggestion
        Optional.ofNullable(dimUrlConfig)
        .map(PathUtils::removeTrailingSlash)
```

edc-extensions/did-document/did-document-service-dim/src/main/java/org/eclipse/tractusx/edc/did/document/service/DidDocumentServiceDimClient.java
```diff
@@ -0,0 +1,409 @@
...
+        this.mapper = mapper;
+        this.ownDid = ownDid;
+        this.monitor = monitor.withPrefix(getClass().getSimpleName());
+        this.didDocApiUrl = String.join("", dimUrl, DID_DOC_API_PATH);
+    }
+
+    @Override
+    public ServiceResult<Void> update(Service service) {
+        return validateService(service)
+                .compose(v -> deleteById(service.getId()))
```
AndrYurk: Why do we delete and recreate the entry during an update? When scaling the same EDC, this results in two unnecessary write operations (delete and create for the same entry). I propose that we first read the existing service by ID and, if it already exists and is unchanged, do nothing.

Issue #2435, opened 2025-11-28
Left Operand WarrantyDurationMonths might have issues


lgblaumeiser:
## Describe the bug

See picture in screenshot section, the red circled policy constraint led to a validation error and the assumption is, that it should not.

### To Reproduce

Create a policy according to the example below

### Expected behavior

The left operand should work as specified

### Screenshots/Error Messages

<img width="2204" height="1374" alt="Image" src="https://github.com/user-attachments/assets/e859669d-b308-4de4-afbd-b6b8237efdaa" />

## Context Information

- Used version: 0.11.0

PR #2460, opened 2025-12-10
Apply policy fix for #2435 to the 0.11.2 bugfix branch

lgblaumeiser:
## WHAT

Related PR: #2458

Compress changes to a bugfix version instead of applying the full pr changes

## WHY

Bug #2435 is also an issue in bugfix release, but refactoring of tests is out-of-scope for the bugfix branch


lgblaumeiser:
The removed tests have been removed, because they were wrong in the first place and approved something that was not according to the spec.

Issue #2287, opened 2025-09-26
Improve wallet handling


lgblaumeiser:
## WHAT

Wallet calls are done by the provider when downloading the credentials from the wallet and from the consumer when the STS token needs to be created. The idea of this feature to reduce the amount of wallet calls by applying multiple strategies.

1. Remember the availability of a credential on the provider side, by storing which credential has been checked when and for how long the validation is assumed to be valid (e.g., latest and the expiry date, the check becomes invalid, but typically much earlier, like after one or several days). Use this information in subsequent DSP requests instead of always retrieving the credentials.
2. To use the wallet to create the STS token is quite heavy weight, as it is a standard OAuth procedure, which could also be provided by some local thing, close to the connector or even within. Investigate and describe alternative STS token generation approaches which can easily applied in a setup.

## WHY

Today, every DSP call ends up in retrieving the credentials from the wallet, checking the credentials according to the relevant policy and move on from there. This results in an awful lot of wallet calls, which increase network traffic and costs.

Actually, the credentials are quite stable, they expire at a certain point in time and they are stable until then. It might be, that they are revoked or renewed prematurely but these are the only expected changes on the credentials. As a consequence it makes sense to think about strategies that reduce the amount of wallet calls.

## HOW

1. Store the result of the evaluation locally in the connector, e.g., in a cache. Provide mechanisms to configure the caching period. Adapt the check of the credentials to isolate the credential checking and to use the check result information in the policy engine.
2. Basically, this should already be possible. As the STS token retrieval just uses standard protocols. Still a concept should be developed, that defines steps like documentation needs or potentially things like an implementation of a token extension. An important issue is, that the public key to verify the signature on the STS token has to be part of the DID document.

lgblaumeiser:
@ronjaquensel @AndrYurk I missed some points we discussed yesterday, would appreciate, if you could add them.

PR #2464, opened 2025-12-11
docs: add DR for VP caching

ronjaquensel:
## WHAT

Adds a decision record for caching Verifiable Presentations within the DCP flow.

## WHY

To reduce the number of wallet calls

Relates to #2287

### DIFF
--- a/docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
+++ b/docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
@@ -1,0 +1,160 @@
+# Verifiable Presentation Caching
+
+## Decision
+
+We will implement a caching mechanism for Verifiable Presentations (VPs) within the DCP communication flow. Each
+requested VP will be cached and before any new presentation request is made, the cache is checked for a matching VP
+first.
+
+## Rationale
+
+For each DSP message exchanged, the receiving connector requests a VP of the sending participant. This includes multiple
+requests to the wallet (sending participant's STS, receiving participant's STS, sending participant's presentation API).
+This causes quite high network traffic, as e.g. during a contract negotiation at least 4 DSP messages are exchanged,
+i.e. the whole request sequence will be run at least 4 times during a single contract negotiation.
+
+As available Verifiable Credentials (VCs) do not frequently change, part of the request sequence can be omitted after
+the whole sequence has been executed once by introducing a cache for VPs. The initial call to the sending participant's
+STS always needs to be made, as the receiving participant may not have any VPs cached. But after initially requesting a
+VP for a participant, the requests to the receiving participant's STS as well as to the sending participant's
+presentation API can be skipped for subsequent DSP messages exchanged with the same participant, thus greatly reducing
+network traffic.
+
+## Approach
+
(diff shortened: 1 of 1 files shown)


docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+As runtimes are provided for both in-memory and persistent deployments, there will be two implementations of
+`VerifiablePresentationCacheStore`: the `InMemoryVerifiablePresentationCacheStore` and the
+`SqlVerifiablePresentationCacheStore`. The former will be the default implementation provided as part of the same module
+as the [CachePresentationRequestService](#cachepresentationrequestservice) and the cache implementation, while the
+latter will be made available in a separate module.
+
+Usually, a cache should be purely in-memory, i.e. not actually persisted, but as EDCs may be downscaled when no
+processes are running, there will also be an SQL implementation to not lose the benefits of caching VPs in scenarios
+where EDCs are frequently downscaled.
```
bmg13: hmm, this is the step that left me thinking. Would make more sense to me to have a "more-usual-cache" solution (like redis or anything) and having something persistable like sql seems unexpected.
with that said, would it make sense to only have the in-mem one? I understand the concern you mention on this paragraph, but does not seem to be a cache-like behaviour.
I am okay with both, no issue on my side, just felt the need to comment.

docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+Even though the VCs are checked also within the cache, there may be situations where an invalid VP is cached, e.g.
+when a VC defined in the requested scopes was initially missing and shortly after added to the wallet. To not block
+communication in these situations, there needs to be a way to trigger removal of cache entries. For this purpose,
+we'll add an API, which will comprise a single endpoint which takes a participant ID as parameter and removes all cache
+entries for that participant when called.
+
+### Configuration
+
+The cache will be enabled by default and have a default validity period of 24 hours. The validity period will be
```
bmg13: I would go with 2 different props: enabled cache + ttl.
But is a personal preference, this seems okay as well.

docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
ronjaquensel: Good point. I thought of SQL initially as that would not require setting up additional external services, but generally, an actual cache implementation would make more sense. I updated the section to be more vague, mentioning both options but not saying one will definitely be implemented. I would not include that part in the initial implementation PR anyway, so we can discuss this in-depth.

docs/development/decision-records/2025-12-11-verifiable-presentation-caching/README.md
ronjaquensel: Fine for me, I don't really have a preference here. Updated it to mention multiple settings.

lgblaumeiser (commented):
lgtm

Issue #2405, opened 2025-11-19
Log stack traces for exceptions


lgblaumeiser:
## WHAT

If code execution fails in an unexpected exception, standard behavior in requests is to handle the exception by returning a 500 or to handle the exception appropriately. This should be changed in a way, that the issue information should end in the log.

## WHY

Analyzing issues in the connector is a pain because it typically involves debugging a running instance and reproducing the error situation. In many cases, the full information on the stack trace could prevent intensive debugging sessions by simply analyzing the stack trace information. This would improve operational excellence.

## HOW

Unfortunately, this is donkeywork, i.e., it means to go through the code and identify catch clauses which hide the stack trace, add a log call at these places, or change it, that this is possible, e.g., if a logger is not available at the place where the changes is needed.

Motivate the necessary changes also upstream, so that we get a complete picture.

hemantsonu20:
Also to add, I (and many others like me) try to debug edc logs and try to find the origin of message but log entry doesn't contain the logger class name and line number.
There are two solution.
1. Use Common logging frameworks such as Slf4j / log4j / logback etc. 
> It is quite disruptive and requires lot of effort to change in the code.
2. Another option which we can use Monitor#withPrefix() to inject class name. We inject monitor via extension into concrete classes.
Concrete class constructors need to be updated like below.
```
public class SomeClientImpl implements SomeClient {
    
    private final Monitor monitor;
    public SomeClientImpl(Monitor monitor) {
        this.monitor = monitor.withPrefix(getClass().getSimpleName());
    }
}
```
> This is small change and works with existing framework seamlessly. Also it can be done incrementally as well as in one go.

AndrYurk:
Thank you, @hemantsonu20 , for your interest in the topic — I really appreciate it.
In my opinion, we should go with option 2 for now. Since you have already done some investigation, would you like to contribute?

PR #2482, opened 2025-12-18
feat: Use Prefix Monitor to inject class name in logs

pratapipatelbcone:
## WHAT

- Use Prefix Monitor to inject class name in logs

## WHY

- Logger name will be present in each log statement
- WIll be easy to identify the log source

## FURTHER NOTES

Relates #2405


lgblaumeiser (changes requested):
Unfortunately I have to withdraw my approval, as there are tests failing, please fix them, thank you!

PR #2484, opened 2025-12-19
Add all relevant versions of openapi specs to api-hub

lgblaumeiser:
## WHAT

Add the openapi spec of older versions of the connector to the api-hub.

## WHY

To support usage of older versions of the connector for development of solutions


stephanbcbauer:
and there we go :) https://eclipse-tractusx.github.io/api-hub/tractusx-edc/ -> thx

PR #2498, opened 2026-01-02
Update jakarta-json to newer version

lgblaumeiser:
## WHAT

As title says

## WHY

Compatibility tests were using the later version, do not know, how the discrepancy came into the game, so I thought, I switch here.


AndrYurk (approved):
lgtm

PR #2510, opened 2026-01-08
feat: Log failure messages

### LINKED ISSUE #2405: Log stack traces for exceptions

AndrYurk:
## WHAT

This PR adds logging for failure messages and exceptions.

## WHY

To improve log transparency and make troubleshooting easier.

Closes #2405

### DIFF
--- a/edc-extensions/dataspace-protocol/src/main/java/org/eclipse/tractusx/edc/protocol/identifier/MembershipCredentialIdExtractionFunction.java
+++ b/edc-extensions/dataspace-protocol/src/main/java/org/eclipse/tractusx/edc/protocol/identifier/MembershipCredentialIdExtractionFunction.java
@@ -59,7 +59,7 @@ public abstract class MembershipCredentialIdExtractionFunction implements Partic
                 .flatMap(this::getIdentifier)
                 .orElseThrow(() -> {
                     var msg = "Required credential type '%s' not present in ClaimToken, cannot extract property '%s'".formatted(IDENTITY_CREDENTIAL, identityProperty());
-                    monitor.severe(msg);
+                    monitor.warning(msg);
                     return new EdcException(msg);
                 });
     }
@@ -70,18 +70,18 @@ public abstract class MembershipCredentialIdExtractionFunction implements Partic
         
         if (vcListClaim == null) {
             var msg = "ClaimToken did not contain a '%s' claim.".formatted(VC_CLAIM);
-            monitor.severe(msg);
+            monitor.warning(msg);
             return Result.failure(msg);
         }
         if (!(vcListClaim instanceof List)) {
             var msg = "ClaimToken contains a '%s' claim, but the type is incorrect. Expected %s, got %s.".formatted(VC_CLAIM, List.class.getName(), vcListClaim.getClass().getName());
-            monitor.severe(msg);
+            monitor.warning(msg);
             return Result.failure(msg);
         }
         var vcList = (List<VerifiableCredential>) vcListClaim;
         if (vcList.isEmpty()) {
             var msg = "ClaimToken contains a '%s' claim but it did not contain any VerifiableCredentials.".formatted(VC_CLAIM);
-            monitor.severe(msg);
+            monitor.warning(msg);
             return Result.failure(msg);
         }
         return Result.success(vcList);
--- a/edc-extensions/dcp/tx-dcp-sts-dim/src/main/java/org/eclipse/tractusx/edc/iam/dcp/sts/dim/DimSecureTokenService.java
+++ b/edc-extensions/dcp/tx-dcp-sts-dim/src/main/java/org/eclipse/tractusx/edc/iam/dcp/sts/dim/DimSecureTokenService.java
@@ -162,7 +162,7 @@ public class DimSecureTokenService implements SecureTokenService {
     private Result<TokenRepresentation> executeRequest(Request request, String context) {
         return httpClient.execute(request, List.of(retryWhenStatusIsNotIn(200, 201)), this::handleResponse)
                 .recover(failure -> {
-                    monitor.warning("Request too %s failed: [%s] %s".formatted(request.url().url(), context, failure.getFailureDetail()));
+                    monitor.warning("Request to %s failed: [%s] %s".formatted(request.url().url(), context, failure.getFailureDetail()));
                     return Result.failure("[%s] %s".formatted(context, failure.getFailureDetail()));
                 });
     }
@@ -188,7 +188,7 @@ public class DimSecureTokenService implements SecureTokenService {
             return baseRequestWithToken()
                     .map(builder -> builder.post(requestBody));
         } catch (JsonProcessingException e) {
-            monitor.warning("Failed to serialize request body: " + e.getMessage(), e);
+            monitor.severe("Failed to serialize request body: " + e.getMessage(), e);
             return Result.failure(e.getMessage());
         }
 
--- a/edc-extensions/dcp/verifiable-presentation-cache/src/main/java/org/eclipse/tractusx/edc/iam/dcp/cache/VerifiablePresentationCacheImpl.java
+++ b/edc-extensions/dcp/verifiable-presentation-cache/src/main/java/org/eclipse/tractusx/edc/iam/dcp/cache/VerifiablePresentationCacheImpl.java
@@ -138,7 +138,7 @@ public class VerifiablePresentationCacheImpl implements VerifiablePresentationCa
                 .toList();
 
         if (requestedScopes.size() > allCreds.size()) {
-            var msg = "Number of requested credentials does not match the number of returned credentials";
+            var msg = "More credentials are requested than returned";
             monitor.debug(msg + ": requested { %s }, returned { %s }".formatted(String.join(",", requestedScopes),
                     String.join(",", types)));
             return Result.failure(msg);
@@ -167,7 +167,7 @@ public class VerifiablePresentationCacheImpl implements VerifiablePresentationCa
             return Result.success();
         } else {
             var msg = "Returned presentations contains invalid issuer. Expected %s found %s".formatted(expectedIssuer, issuers);
-            monitor.debug(msg);
+            monitor.warning(msg);
             return Result.failure(msg);
         }
     }


edc-extensions/agreements/retirement-evaluation-core/src/main/java/org/eclipse/tractusx/edc/agreements/retirement/service/AgreementsRetirementServiceImpl.java
```diff
@@ -77,7 +80,9 @@ public ServiceResult<Void> retireAgreement(AgreementsRetirementEntry entry) {
         return transactionContext.execute(() -> {
             var contractAgreement = contractAgreementService.findById(entry.getAgreementId());
             if (contractAgreement == null) {
-                return ServiceResult.notFound(NOT_FOUND_IN_CONTRACT_AGREEMENT_TEMPLATE.formatted(entry.getAgreementId()));
+                var msg = NOT_FOUND_IN_CONTRACT_AGREEMENT_TEMPLATE.formatted(entry.getAgreementId());
```
lgblaumeiser: Again, not found is really not worth a log entry and especially not on that level, debug could be ok.

edc-extensions/dataplane/dataplane-proxy/dataplane-proxy-http/src/main/java/org/eclipse/tractusx/edc/dataplane/http/pipeline/ProxyHttpDataSource.java
```diff
@@ -59,17 +59,17 @@ private ProxyHttpDataSource() {
...
-        monitor.debug(() -> "Executing HTTP request: " + request.url());
+        monitor.debug(() -> "Executing HTTP request: " + request.url().url());
         try {
             // NB: Do not close the response as the body input stream needs to be read after this method returns. The response closes the body stream.
             var response = httpClient.execute(request);
             var statusCode = (String.valueOf(response.code()));
 
             return handleResponse(response, statusCode);
         } catch (IOException e) {
+            monitor.severe("Exception during executing HTTP request: " + request.url().url(), e);
```
lgblaumeiser: Again, an IOException is something to expect, I would not see this as a severe issue, max a warning

edc-extensions/dataspace-protocol/src/main/java/org/eclipse/tractusx/edc/protocol/identifier/MembershipCredentialIdExtractionFunction.java
```diff
@@ -41,32 +42,51 @@ public abstract class MembershipCredentialIdExtractionFunction implements Partic
...
+    public MembershipCredentialIdExtractionFunction(Monitor monitor) {
+        this.monitor = monitor.withPrefix(getClass().getSimpleName());
+    }
     
     @Override
     public String apply(ClaimToken claimToken) {
         var credentials = getCredentialList(claimToken)
-                .orElseThrow(failure -> new EdcException("Failed to fetch credentials from the claim token: %s".formatted(failure.getFailureDetail())));
+                .orElseThrow(failure -> {
+                    var msg = "Failed to fetch credentials from the claim token: %s".formatted(failure.getFailureDetail());
```
lgblaumeiser: Double logging, all cases are already logged below

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
```diff
@@ -104,16 +107,24 @@ public ServiceResult<DataAddress> refreshToken(String tokenId, DataAddress edr)
         var refreshAudience = edr.getProperties().get(EDR_PROPERTY_REFRESH_AUDIENCE);
 
         if (isNullOrBlank(accessToken)) {
-            return ServiceResult.badRequest("Cannot perform token refresh: required property 'authorization' not found on EDR.");
+            var msg = "Cannot perform token refresh: required property 'authorization' not found on EDR.";
```
lgblaumeiser: The question is, how can this happen, this and the three next issuer. IN which situations, can we have an EDR without these properties. Is this really a bad request, because the user wants to refresh an EDR which has no token, but what is that. I would say, there are two options here:
1. This is really possible, that a user can reference an id, that maps to an EDR which does not have these properties, but then, the error message should be more in a direction that indicates the wrong choice of EDR, not messages that indicate a missing property in something the user cannot change. And then, logging this is questionable.
2. This is actually an issue of the EDR, so this is really a 5xx issue, but then severe would be the right choice.

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
```diff
@@ -152,17 +167,23 @@ private ServiceResult<TokenResponse> executeRequest(Request tokenRefreshRequest)
                     var tokenResponse = objectMapper.readValue(jsonBody, TokenResponse.class);
                     return ServiceResult.success(tokenResponse);
                 }
-                return ServiceResult.badRequest("Token refresh successful, but body was empty.");
+                var msg = "Token refresh successful, but body was empty.";
```
lgblaumeiser: What is the error, and why is this a bad request. This sounds like a programming issue here, please analyze prior to changing this here and adapt the logging to the reason of the issue, potentially changing the return status code.

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
```diff
@@ -152,17 +167,23 @@ private ServiceResult<TokenResponse> executeRequest(Request tokenRefreshRequest)
...
+                monitor.severe(msg);
+                return ServiceResult.badRequest(msg);
             }
+            var responseBody = StringUtils.toString(response.body());
+            monitor.severe(format("Received error response from %s: %s - %s: \"%s\".", tokenRefreshRequest.url().url(),
```
lgblaumeiser: Why do you log an severe entry here, obviously all status codes are 4xx.

lgblaumeiser (changes requested):
I have quite some issues with this PR:
1. You introduce redundant logging, which only clutters the log, some places are obvious, in those I did a remark, but in general, if we log every place, a result is handled and "rethrown", it might be a bad place for logging. You should really understand the flow before you log multiple times along the path.
2. You log issues related to bad user data. This should be max a debug message, so, in case of an issue that a user has, you can turn on debug logging and try to reproduce the issue from the user.
3. I see a tendency to use too high log levels. An error should be something, that indicates a severe issue in the setup, connection issues to external systems should be a warning only. And user induced issues should not be part of the standard log, debug is the max to allow checking the issue after turning on debug logging.

edc-extensions/dataplane/dataplane-proxy/dataplane-proxy-http/src/main/java/org/eclipse/tractusx/edc/dataplane/http/pipeline/ProxyHttpDataSourceFactory.java
```diff
@@ -62,10 +62,14 @@ public String supportedType() {
...
             } else {
-                return failure("Transfer type mismatch: ProxyHttpData data type is only allowed for PULL transfer types.");
+                var msg = "Transfer type mismatch: ProxyHttpData data type is only allowed for PULL transfer types.";
+                monitor.severe(msg);
+                return failure(msg);
             }
         } catch (Exception e) {
-            return failure("Failed to build ProxyHttpDataSource: " + e.getMessage());
+            var msg = "Failed to build ProxyHttpDataSource: " + e.getMessage();
+            monitor.severe(msg, e);
```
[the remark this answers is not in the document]
AndrYurk: will be deleted, failure should be logged upstream

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
```diff
@@ -178,7 +182,9 @@ public Result<TokenResponse> refreshToken(String refreshToken, String authentica
                 .map(accessTokenClaims -> accessTokenDataStore.getById(accessTokenClaims.getStringClaim(JwtRegisteredClaimNames.JWT_ID)));
 
         if (accessTokenDataResult.failed()) {
-            return Result.failure("Access token validation failed: %s".formatted(accessTokenDataResult.getFailureDetail()));
+            var msg = "Access token validation failed: %s".formatted(accessTokenDataResult.getFailureDetail());
+            monitor.severe(msg);
```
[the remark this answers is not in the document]
AndrYurk: will move to debug level.

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
```diff
@@ -202,10 +210,12 @@ public Result<TokenResponse> refreshToken(String refreshToken, String authentica
         var accessTokenData = new AccessTokenData(existingAccessTokenData.id(), claimToken, existingAccessTokenData.dataAddress(), existingAccessTokenData.additionalProperties());
 
         var storeResult = accessTokenDataStore.update(accessTokenData);
-        return storeResult.succeeded() ?
-                Result.success(new TokenResponse(newAccessToken.getContent(),
-                        newRefreshToken.getContent(), tokenExpirySeconds, "bearer")) :
-                Result.failure(storeResult.getFailureMessages());
+
+        if (storeResult.failed()) {
```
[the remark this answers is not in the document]
AndrYurk: Did not get it. It is log for unsuccessful stores in the DB or memory, nothing related to `refreshTokenResult` (line 230) or `accessTokenResult` (line 237)

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
```diff
@@ -129,7 +140,11 @@ public ServiceResult<DataAddress> refreshToken(String tokenId, DataAddress edr)
                         .flatMap(ServiceResult::from))
                 .compose(authToken -> createTokenRefreshRequest(refreshEndpoint.toString(), refreshToken.toString(), "Bearer %s".formatted(authToken.getToken()))
                         .flatMap(ServiceResult::from))
-                .recover(f -> ServiceResult.badRequest("Could not execute token refresh: " + f.getFailureDetail()))
+                .recover(f -> {
```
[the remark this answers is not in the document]
AndrYurk: Yes, it is, but it happens on the consumer side, while `DataPlaneTokenRefreshServiceImpl` happens on the provider side.

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
AndrYurk: As I know, we cannot create an EDR without these properties, so will move to sever.

edc-extensions/tokenrefresh-handler/src/main/java/org/eclipse/tractusx/edc/common/tokenrefresh/TokenRefreshHandlerImpl.java
AndrYurk: By the code base, the successful response body should never be empty, unless it was not consumed somewhere in the middle. I will move in to warning and change to an unexpected result.

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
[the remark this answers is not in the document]
lgblaumeiser: Oh sorry,

edc-extensions/dcp/tx-dcp-sts-dim/src/main/java/org/eclipse/tractusx/edc/iam/dcp/sts/dim/DimSecureTokenService.java
```diff
@@ -159,7 +161,10 @@ private Result<Map<String, Object>> signTokenPayload(Map<String, Object> claims)
 
     private Result<TokenRepresentation> executeRequest(Request request, String context) {
         return httpClient.execute(request, List.of(retryWhenStatusIsNotIn(200, 201)), this::handleResponse)
-                .recover(failure -> Result.failure("[%s] %s".formatted(context, failure.getFailureDetail())));
+                .recover(failure -> {
+                    monitor.warning("Request too %s failed: [%s] %s".formatted(request.url().url(), context, failure.getFailureDetail()));
```
lgblaumeiser:
```suggestion
                    monitor.warning("Request to %s failed: [%s] %s".formatted(request.url().url(), context, failure.getFailureDetail()));
```

edc-extensions/dcp/verifiable-presentation-cache/src/main/java/org/eclipse/tractusx/edc/iam/dcp/cache/VerifiablePresentationCacheImpl.java
```diff
@@ -153,7 +166,9 @@ private Result<Void> verifyPresentationIssuer(String expectedIssuer, List<Verifi
         if (issuers.stream().allMatch(expectedIssuer::equals)) {
             return Result.success();
         } else {
-            return Result.failure("Returned presentations contains invalid issuer. Expected %s found %s".formatted(expectedIssuer, issuers));
+            var msg = "Returned presentations contains invalid issuer. Expected %s found %s".formatted(expectedIssuer, issuers);
+            monitor.debug(msg);
```
lgblaumeiser: IN the logic of above, this is a warning, as it is actually more crititcal than many things logged as warning above.

PR #2552, opened 2026-02-03
feat: reduced mocking in the pg‑runtime test classes

### LINKED ISSUE #2226: e2e-test:runtime-postgresql does not pickup does not test JTI validation operations
## Describe the issue

When executing the E2E tests using the `runtime-postgresql` runtime, possible issues with identity and trust operations, such as JTI validations, are not picked up.

This is due to the fact that both the `dcp:tx-dcp` and the `dcp:tx-dcp-sts-dim` extensions are being excluded of the E2E tests for this runtime, each of which make use of the `IdentityAndTrust` extension, which is used by the `TokenValidationServiceImpl` to run JTI validation rules. On the E2E tests these operations are mocked.

Having the setup as such, may lead to issues such as https://github.com/eclipse-tractusx/tractusx-edc/issues/2217, which was caused by a missing migration. Such an issue would have been picked up, if the runtime did not entirely mock the IAM operations.

## Context Information

- Used version: 0.11.0-RC1
 
## Possible Improvement

A possible improvement could be to not mock these operations, and instead use real values for these configurations. The disadvantage of this implementation is that it couples the E2E tests to an external service.

A different, possible improvement would be to be more granular when mocking resources, so that only the connections to external services are mocked, and not the entire flow. The disadvantage of this implementation is that it will be harder and more complex to maintain than the current implementation.

BrunoMiguelNogueira:
## WHAT

Reduced mocking in the pg‑runtime test classes.

## WHY

Using logic closer to the main runtime rather than heavy mocking, this exposes bugs that mocks would otherwise fail to trigger.

## FURTHER NOTES

The driver for this improvement was issue #2226.
Because the tests now evaluate the same validation rules as the main runtime—specifically `JtiValidationRule`—removing all references to the `JtiValidationStore` schema (to accurately reproduce the conditions of issue #2217) results in failures with the error `relation "edc_jti_validation" does not exist`.

---
Closes #2226.


lgblaumeiser (commented):
lgtm, @bmg13 I leave it to you to have a second look and to merge

correiaafonso12 (commented):
LGTM :) comments are just small nits

PR #2579, opened 2026-02-11
Fix OpenApi spec for connector discovery

lgblaumeiser:
## WHAT

The OpenApi spec did not show the example properly, because it could not detect the structure. I hope, this fixes the issue

## WHY

Because it is ugly


AndrYurk (approved):
lgtm

Issue #2609, opened 2026-02-24
During release of 0.11.0-rcx, the modules have not been released on maven central


lgblaumeiser:
## Describe the bug

Neither 0.11.0-rc1 nor 0.11.0-rc2 have been uploaded to maven central, 0.11.0-rcX are there, so this is a new phenomenon since 0.11.2 which is also there.

The build was successful and the publish step did not report any issues.

### To Reproduce

Check release build for 0.11.0-rc2 and observe that the modules are not available on maven central

### Expected behavior

The modules are available on Maven Central

PR #2631, opened 2026-03-02
fix: remove snapshot dependency

AndrYurk:
## WHAT

This PR removes snapshot dependency

## WHY

Failed validations from the central Maven portal while using snapshot dependencies.

Part of #2609


lgblaumeiser (commented):
lgtm

lgblaumeiser (approved):
again lgtm, lets wait for the checks

PR #2632, opened 2026-03-03
chore: deprecate C-X ODRL context

ronjaquensel:
## WHAT

Marks the Catena-X ODRL context as deprecated and removes it from the registered contexts, i.e. it will not be actively used for compaction anymore. Replaces all references to the context in tests and documentation with `http://www.w3.org/ns/odrl.jsonld`. 

For now, keeps the registration of the context file as a cached document, so that it remains processable, as it is still used by older connector versions. It may be removed completely in a future version.

## WHY

As stated in [this issue](https://github.com/catenax-eV/cx-ex-data-sovereignty/issues/36#issuecomment-3932262566), it was decided by the data sovereignty expert group to remove the Catena-X ODRL context. 

Relates to https://github.com/catenax-eV/cx-ex-data-sovereignty/issues/36

### DIFF
--- a/core/json-ld-core/src/main/resources/document/edc-v1.jsonld
+++ b/core/json-ld-core/src/main/resources/document/edc-v1.jsonld
@@ -13,7 +13,7 @@
       "@id": "edc:policy",
       "@type": "@id",
       "@context": [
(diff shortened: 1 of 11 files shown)


lgblaumeiser (commented):
lgtm, I'll apply it for the release

PR #2633, opened 2026-03-03
fix: change return type on provider to interface

### LINKED ISSUE #2618: BPN-to-agreement mapping isn't persisted
## Describe the bug

The relevant `edc_contract_agreement_bpns` table isn't populated.

### To Reproduce

1. Provider: Setup tractusx-edc with Postgres persistence configured
2. Provider: setup Asset, Policy, Contract Definition
3. Consumer: negotiate successfully
4. Provider: SSH into the postgres pod
5. Provider: execute `psql "postgres://user:password@localhost:5432/edc" -c "SELECT * FROM edc_contract_agreement_bpns"`. Change the connection string if necessary
6. Provider: execute `psql "postgres://user:password@localhost:5432/edc" -c "SELECT * FROM edc_contract_agreement"`. Change the connection string if necessary

### Expected behavior

There shouldn't be a difference between the counts in step 5 and 6.

In reality, step 5 always shows that the table is empty

## Context Information

0.12.0-SNAPSHOT

## Possible Implementation

Didn't see anything in logs and didn't debug into the runtime yet

arnoweiss:
## WHAT

`SqlAgreementsBpnsStore` wasn't properly being injected because the method annotated with `@Provider` was retuning the implementation's class name and not that of the Interface.

## WHY

store bpn-agreementid-mappings in table, not memory

## FURTHER NOTES

I built the image, ran it on my cluster and found that the agreements negotiated based on it are persisted properly. I'd be unsure how to write unit tests for this.

image

Closes #2618


bmg13 (approved):
Agreed on to test but uncertain how. Either way is a needed change

PR #2698, opened 2026-03-26
fix: Handle older connectors <=0.9.0 for version metadata

lgblaumeiser:
## WHAT

The `.well-known/dspace-version` endpoint was access-controlled in previous versions <= 0.9.0, this workaround fixes that without adding too much complexity, as this is a temporary effect, that will be gone soon enough.  

## WHY

To support detection of connectors <= 0.9.0

## FURTHER NOTES

Closes #2696


AndrYurk (approved):
lgtm

PR #2714, opened 2026-03-30
feat/multi-tenant-bdrs: Enable multi-tenancy via `participantContextId` and restructure cache mechanism 

### LINKED ISSUE #2695: `edc-extensions/bdrs-client` multi-tenancy feature
## WHAT

| Modules                 | Description                           | Ready? | Relevance | Notes |
|-------------------|----------------------------|---------|-----------|-------|
| `edc-extensions/bdrs-client` | registers an `AudienceMapper` that will use the BDRS to map between BPNLs and DIDs | ❌/✅ | Standard Compliance | Is not participant aware, but potentially does not have to be, as there is no value in storing the mapping data redundantly and there is no record on which participant uses which mappings |

## WHY

As specified in #2683 we need to do some "changes" to the EDC extensions in order for it to be fully multi-tenancy aware.

## HOW

Replicating the behavior observed in the upstream EDC multi-tentancy modules which are aware of the `participantId` in the api context. Utilizing `UUIDs` similar to what is done in the `tractusx-identity-hub`.

matbmoser:
> [!CAUTION]
> UNIT TESTs not yet implemented. I wanted to get some feedback so I can continue with the implementation until end .

## WHAT

- Restructured the cache mechanism to a class, so it can have it own constructuctor and manage the different participants caches independently.
- Refactored the BRDS client to support the new cache mechanism.
- Enhanced error handling from BDRS client implementation. So if there is an exception it will print.

## WHY

The BDRS cache was not yet participant context aware.

I also believed that the way we manage the BDRS cache can be improved and we should have a class which manages this cache for us, having methods to execute the CRUD + supporting us on the maintenance.

## FURTHER NOTES

I have programmed it myself this time, the interfaces, implementation class and architecture was done my me.

Copilot (Claude OPUS) was used for support only on the documentation, and to find which references are bring broken by the refactor and it will be used to create unit tests.

### Questions that I have still (may require ADR):

- In order to continue with "backward compatibility" shall we enable a "default" participant id, which can be used if the methods are called without a participantContextId?

Closes #2695

### DIFF
--- a/edc-extensions/bdrs-client/src/main/java/org/eclipse/tractusx/edc/identity/mapper/BdrsCache.java
+++ b/edc-extensions/bdrs-client/src/main/java/org/eclipse/tractusx/edc/identity/mapper/BdrsCache.java
@@ -1,0 +1,134 @@
+/********************************************************************************
+ * Copyright (c) 2026 Catena-X Automotive Network e.V.
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License, Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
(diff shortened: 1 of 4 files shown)


lgblaumeiser:
Hi @matbmoser , I close this pr now, as the main functionality is not needed. We have discussed that. The proposal with the cache is not valuable, as there are existing solutions for a cache, it remains a good reason, why concurrent caches are not used in this implementation, but on the other hand, it is not really a cache, but a redundant copy of one database, so updates are atomic by replacing the current map with the newly built one, so concurrency is not really an issue.

PR #2716, opened 2026-03-30
fix: move from localstack to floci

AndrYurk:
## WHAT

This PR moves from localstack to floci

## WHY

To make end2end-transfer-cloud test runs successful

Closes #2715


bmg13 (approved):
Thank You @AndrYurk

PR #2730, opened 2026-04-08
Extend policy logs

### LINKED ISSUE #2729: Extend logs for policy evaluator
## WHAT

Logs all evaluation failures for policies.

## WHY
For policy evaluation, no logs are exposed for any failures, which requires debugging the codebase to identify the root cause.

## HOW

In case of failure, write it to the logs.

AndrYurk:
## WHAT

This PR logs all evaluation failures for policies.

## WHY

To simplify the identification of a problem's root causes.

Closes #2729

### DIFF
--- a/edc-extensions/cx-policy-legacy/src/main/java/org/eclipse/tractusx/edc/policy/cx/legacy/dismantler/DismantlerCredentialConstraintFunction.java
+++ b/edc-extensions/cx-policy-legacy/src/main/java/org/eclipse/tractusx/edc/policy/cx/legacy/dismantler/DismantlerCredentialConstraintFunction.java
@@ -25,6 +25,7 @@ import org.eclipse.edc.participant.spi.ParticipantAgentPolicyContext;
 import org.eclipse.edc.policy.engine.spi.PolicyContext;
 import org.eclipse.edc.policy.model.Operator;
 import org.eclipse.edc.policy.model.Permission;
+import org.eclipse.edc.spi.monitor.Monitor;
 import org.eclipse.tractusx.edc.core.utils.credentials.CredentialTypePredicate;
 import org.eclipse.tractusx.edc.policy.cx.legacy.common.AbstractDynamicCredentialConstraintFunction;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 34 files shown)


edc-extensions/cx-policy-legacy/src/main/java/org/eclipse/tractusx/edc/policy/cx/legacy/dismantler/DismantlerCredentialConstraintFunction.java
```diff
@@ -57,6 +58,12 @@ public class DismantlerCredentialConstraintFunction<C extends ParticipantAgentPo
     // allows to encode multiple values in a single string in the right-operand. Policies don't handle list-type right-operands well.
     public static final String RIGHT_OPERAND_LIST_SEPARATOR = ",";
     private static final String ALLOWED_ACTIVITIES_LITERAL = "activityType";
+    private final Monitor monitor;
```
lgblaumeiser, line 61: Wouldn't it make sense to make the monitor protected instead of creating the field multiple times?

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/datausage/DataUsageEndDefinitionConstraintFunction.java
```diff
@@ -33,7 +34,7 @@
 public class DataUsageEndDefinitionConstraintFunction<C extends ParticipantAgentPolicyContext> extends ValueValidatingConstraintFunction<Permission, C> {
     public static final String DATA_USAGE_END_DEFINITION = "DataUsageEndDefinition";
 
-    public DataUsageEndDefinitionConstraintFunction() {
+    public DataUsageEndDefinitionConstraintFunction(Monitor monitor) {
```
lgblaumeiser, line 37: Here seems to be something wrong, the parameter is not used, either you forgot it to add to the super call or it is not needed.

lgblaumeiser (changes requested):
Some general issues:

- Have you tried to understand how the context problems are used after the call? It seems odd, that there are messages created, but not used.
- The other thing is, that the failing evaluations are actually nothing for the logs, as these are business issues.
  - I could imagine, that all you need to do, is to get an event that is triggered with the context information and have all reported failures in that event.
  - This change makes the debug setting something to handle with care, we must not put business activities in the logs in general, this is something, an connector owner can decide.
- This solution will not survive the expected changes, as policy evaluations are replaced by a scripting language and in that, logging will be another story.

PR #2740, opened 2026-04-13
feat: resolve BPN from MembershipCredential in BusinessPartnerNumberConstraintFunction

### LINKED ISSUE #2462: BPN Credential based BPN retrieval instead of using reversed BDRS client
## WHAT

As a temporary solution, the retrieval of a BPNL, e.g., during contract negotiation is done based on the BDRS client and a reversed mapping of DID to BPN. The target solution is actually to use the BPN credential of the consumer to retrieve the BPN for a given DSP message. This would push the BDRS client out of the DSP handling and makes it only relevant in the discovery processes.

## WHY

The BPN credential is the more trustworthy source of a bpn compared to the BDRS

arnoweiss:
## Summary

- Removes the dependency on `BdrsClient` for BPN resolution in `BusinessPartnerNumberConstraintFunction`
- The BPN is now extracted directly from the `holderIdentifier` property of the participant's `MembershipCredential` (VC claim), eliminating the external BDRS lookup
- Updates `CxPolicyExtension` to construct `BusinessPartnerNumberConstraintFunction` without the `BdrsClient` argument

Closes #2462

### DIFF
--- a/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/CxPolicyExtension.java
+++ b/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/CxPolicyExtension.java
@@ -181,7 +181,7 @@ public class CxPolicyExtension implements ServiceExtension {
         engine.registerFunction(CatalogPolicyContext.class, Permission.class,
                 withCxPolicyNsPrefix(BUSINESS_PARTNER_GROUP), new BusinessPartnerGroupConstraintFunction<>(store, bdrsClient));
         engine.registerFunction(CatalogPolicyContext.class, Permission.class,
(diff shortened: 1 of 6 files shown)


lgblaumeiser:
Hello @arnoweiss , just a general remark, I would suggest to use the BpnCredential for extracting the information. Imho, the membership credential represents membership whereas the bpn credential associates the BPNL to the holder. The latter is the thing we want to check here.

arnoweiss:
Should it check against `bpn` or `holderIdentifier`? Both are mandatory. In the `MembershipCredential`, the `holderIdentifier` contains the BPNL.

lgblaumeiser:
> Should it check against `bpn` or `holderIdentifier`? Both are mandatory. In the `MembershipCredential`, the `holderIdentifier` contains the BPNL.

I would say the bpn field, as this is the purpose of the credential to be issued.

lgblaumeiser (approved):
lgtm

PR #2744, opened 2026-04-14
feat: Use participant context id as identifier for did document registration

### LINKED ISSUE #2743: Use participant context id as did document service identifier
## WHAT

Currently, an explicit configuration is needed to register the connector as service in the DID document. But with the participant context id, there is already an unique identifier for the connector. With this change, this identifier is used for the identifier in the did document.

## WHY

To remove redundant configurations and thereby configuration complexity

## HOW

Make the identifier to an uri by adding "urn:" as prefix and use it as identifier

lgblaumeiser:
## WHAT

_Briefly describe what your PR changes, which features it adds/modifies._

## WHY

## FURTHER NOTES

Closes #2743


AndrYurk (approved):
lgtm

PR #2745, opened 2026-04-15
infra: Make upgrade base fix temporarily

lgblaumeiser:
## WHAT

Refer to version 0.13.0-rc1 in upgradability tests instead of computing the last released version

## WHY

Due to the helm changes in the recent version, upgradability only works against this version due to the postgres helm chart changes.

### DIFF
--- a/.github/workflows/upgradeability-test.yaml
+++ b/.github/workflows/upgradeability-test.yaml
@@ -59,17 +59,14 @@ jobs:
       - name: "Get latest released version"
         id: get-version
         run: |
-          echo "RELEASE=0.13.0-rc1" >> $GITHUB_ENV # TODO: switch back once 0.13.0 is released
+          RELEASED_VERSION=$(helm search repo tractusx/tractusx-connector -l --devel -o json | jq -r 'first | .version')
+          echo "Last official release is $RELEASED_VERSION"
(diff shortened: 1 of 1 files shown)


lgblaumeiser:
Hi @AndrYurk @bmg13 , this pr allows us to temporarily fix the upgradability tests until we get a new release which is based on the changed helm charts. Please have a look with priority!

AndrYurk (changes requested):
I'm concerned about this solution because it requires us to update the workflow every time we decide to run an upgradeability test based on a release candidate. I propose pushing the workflow to take the RC automatically if it exists. So there should be enough to update the  `Get latest released version` step with `helm search repo tractusx/tractusx-connector -l --devel -o json | jq -r 'first | .version'`.

lgblaumeiser:
> I'm concerned about this solution because it requires us to update the workflow every time we decide to run an upgradeability test based on a release candidate. I propose pushing the workflow to take the RC automatically if it exists. So there should be enough to update the `Get latest released version` step with `helm search repo tractusx/tractusx-connector -l --devel -o json | jq -r 'first | .version'`.

Good point, thank you for the hint. Unfortunately the version has not been published, but I can solve that one.

AndrYurk (approved):
lgtm

PR #2775, opened 2026-04-24
fix: make dependabot ignore SNAPSHOT versions

### LINKED ISSUE #2774: Make dependabot ignore SNAPSHOT versions
## Describe the bug

Dependabot proposes SNAPSHOT dependency for the update. This is not correct because Maven then rejects such a library from publishing
See https://github.com/eclipse-tractusx/tractusx-edc/pull/2773

### Expected behavior

Dependabot should not propose SNAPSHOT dependencies

## Possible Implementation

Add SNAPSHOT to ignore

AndrYurk:
## WHAT

As the title says

## WHY

Because Maven rejects the library with  SNAPSHOT dependencies from publishing

Closes #2774

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,8 @@ updates:
     ignore:
       - dependency-name: "org.eclipse.dataspacetck.dsp:*"
       - dependency-name: "org.eclipse.dataspacetck.dcp:*"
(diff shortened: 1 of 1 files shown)


lgblaumeiser (approved):
lgtm

PR #2776, opened 2026-04-24
fix: remove snapshot maven repo

### LINKED ISSUE #2774: Make dependabot ignore SNAPSHOT versions

AndrYurk:
## WHAT

Revert changes from [PR](https://github.com/eclipse-tractusx/tractusx-edc/pull/2775) and reimplement by removing maven snapshot repository.

## WHY

Because Maven rejects the library with SNAPSHOT dependencies from publishing

## FURTHER NOTES

Closes #2774

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,8 +37,6 @@ updates:
     ignore:
       - dependency-name: "org.eclipse.dataspacetck.dsp:*"
       - dependency-name: "org.eclipse.dataspacetck.dcp:*"
-      - dependency-name: "*"
-        versions: [ "*SNAPSHOT" ]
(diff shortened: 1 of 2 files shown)


lgblaumeiser (commented):
lokishtm

lgblaumeiser (approved):

PR #2796, opened 2026-05-04
fix: add write permission for allure

AndrYurk:
## WHAT

Add write permission for allure

## WHY

To make `generate-allure-report` work


lgblaumeiser (approved):
lgtm

PR #2798, opened 2026-05-06
feat: remove validation from UsagePurposeConstraintFunction

### LINKED ISSUE #2758: Usagepurpose Individual is rejected during validation
## Describe the bug

we tried to create a policy with a usagepurposeindividual rightoperand. 
this means that the rightoperand is not a constant but a string based on standardization. 

see "UsagePurposeIndividualRightOperand" in the schema reference. 
https://github.com/catenax-eV/cx-odrl-profile/blob/main/schema/constraint/usage-purpose-constraint-schema.json

` {
              "leftOperand": "UsagePurpose",
              "operator": "isAnyOf",
              "rightOperand": "this is a specific text"
   }`

during the validation of the edc we receive the following error message 

`
call to external system https://api.bmw-internal.url.xx  failed with statuscode 400 - Message: Invalid right-operand: the following values are not allowed: [this is a specific text]
`
### To Reproduce

_Steps to reproduce the behavior:_

1. Create a policy using the leftoperand "usagepurpose" and define the "rightoperand" to be a string value 
2. try to register the policy in the edc

### Expected behavior

_A clear and concise description of what you expected to happen._
* the policy should be validated and created 

_If applicable, add screenshots and/or error messages to help explain your problem._

## Context Information

_Add any other context or information about the problem here._

- Used version: [e.g. Commit Hash]
- OS: [e.g. Mac OS (M1), Windows, Linux]
- Docker Version: [e.g. 20.10.12]
- `java --version`:

## Possible Implementation

_You already know the root cause of the erroneous state and how to fix it? Feel free to share your thoughts._

MDSBarbosa:
## WHAT

Remove validation from `UsagePurposeContraintFunction`

## WHY

According to the [schema reference](https://github.com/catenax-eV/cx-odrl-profile/blob/33d6271/schema/constraint/usage-purpose-constraint-schema.json), the `UsagePurposeContraint` can be defined at will by the data provider and the data consumer. Specifically, this is defined by the [UsagePurposeIndividualRightOperand](https://github.com/catenax-eV/cx-odrl-profile/blob/33d6271/schema/constraint/usage-purpose-constraint-schema.json#L208).

## FURTHER NOTES

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/2758

### DIFF
--- a/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/usage/UsagePurposeConstraintFunction.java
+++ b/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/usage/UsagePurposeConstraintFunction.java
@@ -20,19 +20,28 @@
 package org.eclipse.tractusx.edc.policy.cx.usage;
 
 import org.eclipse.edc.participant.spi.ParticipantAgentPolicyContext;
-import org.eclipse.edc.policy.engine.spi.AtomicConstraintRuleFunction;
 import org.eclipse.edc.policy.model.Operator;
 import org.eclipse.edc.policy.model.Permission;
+import org.eclipse.edc.spi.result.Result;
+import org.eclipse.tractusx.edc.policy.cx.common.BaseConstraintFunction;
+
+import java.util.Set;
 
 /**
  * This is a placeholder constraint function for UsagePurpose. It always returns true but allows
  * the validation of policies to be strictly enforced.
  */
-public class UsagePurposeConstraintFunction<C extends ParticipantAgentPolicyContext> implements AtomicConstraintRuleFunction<Permission, C> {
+public class UsagePurposeConstraintFunction<C extends ParticipantAgentPolicyContext> extends BaseConstraintFunction<Permission, C> {
     public static final String USAGE_PURPOSE = "UsagePurpose";
 
+    public UsagePurposeConstraintFunction() {
+        super(
+                Set.of(Operator.IS_ANY_OF)
+        );
+    }
+
     @Override
-    public boolean evaluate(Operator operator, Object rightValue, Permission rule, C context) {
-        return true;
+    protected Result<Void> validateRightOperand(Object rightValue) {
+        return Result.success();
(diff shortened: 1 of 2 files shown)


edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/usage/UsagePurposeConstraintFunction.java
```diff
@@ -20,54 +20,19 @@
...
-                        "cx.bpdm.vas.nps.download:1",
-                        "cx.ccm.base:1",
-                        "cx.bpdm.poolAll:1",
-                        "cx.logistics.base:1"
-                ),
-                true
-        );
+    @Override
+    public boolean evaluate(Operator operator, Object rightValue, Permission rule, C context) {
+        return true;
```
lgblaumeiser: I suggest, that we check for the operator, as for the operator, only "isAnyOf" is allowed.

lgblaumeiser (changes requested):

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/usage/UsagePurposeConstraintFunction.java
MDSBarbosa: Hey @lgblaumeiser! I've implemented it at [fabf5d5](https://github.com/eclipse-tractusx/tractusx-edc/pull/2798/commits/fabf5d52678dc48e70e32829bf0ed50a341b210b).

Kept the `evaluate` as-is (still returning `true`, now inherited from the `BaseConstraintFunction` class), and instead validated that the `IS_ANY_OF` operator is the only one that can be used on the `validate` method (also inherited from the `BaseConstraintFunction`).

bmg13 (approved):

correiaafonso12 (approved):

lgblaumeiser (approved):
lgtm

PR #2821, opened 2026-05-20
fix: add headers on the source address before storage

ndr-brt:
## WHAT

Move AgreementId and Bpn headers set just before the storage of the DataAddress that then will be used as source in the proxy calls

## WHY
- edc-data plane is bugged (provisioners with consumer pull transfer is not working correctly), plus the refresh token feature is an additional piece that adds complexity. this can be considered a patch that could eventually be refactored in the moment (if that'll happen) the EDC-DataPlane code will be inherited

## FURTHER NOTES
- this PR is not targetin `main`, but the `upstream016` PR, that's the base for PR #2811 
- this PR is NOT fixing the 0.8 tests


lgblaumeiser (approved):
lgtm

PR #2822, opened 2026-05-20
fix: 0.8 tests

ndr-brt:
## WHAT

Registers cached jsonld documents also in tests, otherwise they won't work.

## WHY

## FURTHER NOTES
- fixed a subtle bug for which the selected protocol version was always 0.8


lgblaumeiser (approved):
lgtm

PR #2824, opened 2026-05-21
fix: move federate catalog migration to baseline

AndrYurk:
## WHAT

This PR moves federated catalog migration to baseline, as `control-plane-migration` is deprecated and excluded.

## WHY

To have the federated catalog migration script run.


lgblaumeiser (approved):
lgtm

PR #2825, opened 2026-05-21
chore: skip run "publish snapshot" during release

AndrYurk:
## WHAT

Because "Publish Snapshot" runs now after "Run All Tests", we need to skip it during the release.

## WHY

Redundant call of "Publish Snapshot"  workflow


lgblaumeiser (approved):
lgtm

PR #2845, opened 2026-05-27
feat(tractusx-connector): add postgres custom user and secrets

tom-rm-meyer-ISST:
## WHAT

Adds the the following possibilities to the persistent (postgresql and vault) chart:

- differentiate admin and custom non-root user
- use secrets for controlplane and dataplane if postgres is enabled

## WHY

The cloudpirates chart works differently and explicitly changes an admin user instead of creating a non-admin user. Added the capability to differentiate during deployment (external only).

## FURTHER NOTES

The postgresql names will be `{name}-postgresql`.

Did a few things that might not be correct:

- [ ] default custom user is `edc` and not `user` (user is also a reserved term in psql) but might break deployments not specifying users
- [ ] added license header to helm templates with contributors to EF with start of existence
- [ ] did not bump version but consider this ideally a 0.13.0 change
- [ ] I did not fully evaluate the vault way of pulling data; I expect this to be not an issue as currently user and password have also been required.

> [!note]
> Currently there is no cc-by 4.0 note in the end of the `charts/tractusx-connector/README.md`. I propose to add one to the template of both charts.


tom-rm-meyer-ISST:
I noticed the missing values (see failure of `Lint helm charts / helm-lint (pull_request)`) also during my tests. These have been failing before. Is there any preferred way of fixing them? I can also set them to `CHANGEME` (similar as in new `context` field).

Further Note: Did a clean install on int and flyaway went through with customUser. Further I conducted my whole upgrade of an existing postgres instance using this chart version.

lgblaumeiser:
> I noticed the missing values (see failure of `Lint helm charts / helm-lint (pull_request)`) also during my tests. These have been failing before. Is there any preferred way of fixing them? I can also set them to `CHANGEME` (similar as in new `context` field).
> 
> Further Note: Did a clean install on int and flyaway went through with customUser. Further I conducted my whole upgrade of an existing postgres instance using this chart version.

Interesting effect, no idea, why they pop up now, as I would have expected, that some of them are in the values file. Strange!

Please them with CHANGEME for now, we will consider them afterwards, thank you!

tom-rm-meyer-ISST:
> > I noticed the missing values (see failure of `Lint helm charts / helm-lint (pull_request)`) also during my tests. These have been failing before. Is there any preferred way of fixing them? I can also set them to `CHANGEME` (similar as in new `context` field).
> > Further Note: Did a clean install on int and flyaway went through with customUser. Further I conducted my whole upgrade of an existing postgres instance using this chart version.
> 
> Interesting effect, no idea, why they pop up now, as I would have expected, that some of them are in the values file. Strange!
> 
> Please them with CHANGEME for now, we will consider them afterwards, thank you!

Updated!

charts/tractusx-connector/README.md.gotmpl
```diff
@@ -36,6 +36,65 @@ In addition, in order to map BPNs to DIDs, a new service is required, called the
...
+```yaml
+postgresql:
+  auth:
+    existingSecret: "my-existing-secret-name"
+    secretKeys:
+      adminPasswordKey: "postgres-password"
+  customUser:
+    existingSecret: "my-existing-secret-name"
+    secretKeys:
+      password: "CUSTOM_PASSWORD"
```
lgblaumeiser, line 92: Is there a reason, why these defaults are deviating from the overall convention to use minor letters and '-' as separator?

tom-rm-meyer-ISST: Indeed, I used the defaults used by the [cloudpirates chart](https://github.com/CloudPirates-io/helm-charts/blob/postgres-0.18.3/charts/postgres/templates/custom-user-secret.yaml#L21). I considered these keys for all of the secrets.

If you would like to use meaningful dummies (e.g. `custom-password-key`), please let me know.

In case you would like to use the office hour to discuss further details, I'll make it today to the meeting!

lgblaumeiser: So, to understand everything right, we talk about a property name in a secret, that is used by both Postgres to set up the user and the connector to make use of the secret for accessing the database, right? And the cloud pirates helm charts expect exactly this property name when setting up postgres, right? That would mean, in the vault or whereever the secret is stored, this value is used for storing the secret.

From my side, I could live with that default name, what do you say @AndrYurk , @bmg13 ?

lgblaumeiser: So, to understand everything right, we talk about a property name in a secret, that is used by both Postgres to set up the user and the connector to make use of the secret for accessing the database, right? And the cloud pirates helm charts expect exactly this property name when setting up postgres, right? That would mean, in the vault or whereever the secret is stored, this value is used for storing the secret.

From my side, I could live with that default name, what do you say @AndrYurk , @bmg13 ?

tom-rm-meyer-ISST: It's about the property name in the secret. By default the `CUSTOM_x` are used by cloudpirates. A user may overwrite them.

The respective secret value of that property will be used. This value is the default value used by cloudpirates:

- [secret creation](https://github.com/CloudPirates-io/helm-charts/blob/postgres-0.18.3/charts/postgres/templates/custom-user-secret.yaml)
- [custom user creation](https://github.com/CloudPirates-io/helm-charts/blob/postgres-0.18.3/charts/postgres/templates/init-scripts-configmap.yaml#L17)

Usage examples below.

## Postgres Vault Extension and Chart

BUT I might not get the following - _this might also indicate, that we should not introduce the pull from secrets feature?_

> That would mean, in the vault or whereever the secret is stored, this value is used for storing the secret.

I don't use the feature but I noticed that the postgres credentials are first tried to get pulled from the vault during startup for some edc property value. The current proposal for the chart update **DOES NOT** update any values in the vault. This would mean, that the helm chart would need to get the vault values feeded via AVP. We can add such an example scenario to the gotmpl so that it's in the `README.md`. Honestly I never used the feature as I didn't find documentation easily (this would be a way of promoting it). So it would be something like the following before the heading:

To use the same values for customUser and postgres user as set in the vault, do the following:

1. Configure the vault postgres values, picked up as prioritized value in the vault

```yaml
#... some value structure for the admin user (not pulled by EDC)
#... some value structure for the custom user (pulled by the EDC)
```

2. Configure the helm chart to pick up the correct values

```yaml
postgresql:
  auth: # use e.g. AVP to pull the correct values for admin user
    database: "<path:kv-engine/data/path/value#content"
    #...
  customUser: # use e.g. AVP to pull the correct values for admin user
    database: "<path:kv-engine/data/path/value#content"
    #...
```

## Usage of secrets

### Use of existing secret

Admin creates the secrets in any other way (e.g. manually, secrets operator). If the Admin used other keys, he can overwrite that.

Example: Admin created a secret `operator-created-secret-postgresql` with the keys

- `CUSTOM_DB`
- `CUSTOM_USER`
- `password`

```yaml
postgresql:
  customUser:
    existingSecret: "operator-created-secret-postgresql"
    secretKeys:
      password: "password"
```

### Let the chart handle the secret

The chart uses the default values for the secret names.

Example: Chart updates / creates a secret `{Release-Name}-postgresql-custom-user-credentials` with the keys

- `CUSTOM_DB`
- `CUSTOM_USER`
- `CUSTOM_PASSWORD`

```yaml
postgresql:
  customUser:
    name: "edc"
    database: "edc"
    password: "password"
```

lgblaumeiser: Sorry, @tom-rm-meyer-ISST, but you are out of my context with your explanations. It already starts with the image you are talking about.

Here my information state:
- The connector has a property family edc.datasource.<source_x>.* with * being replaced by e.g. user or password.
- If user and password is stored in the vault, each of them is a own secret with a content property that contains  the corresponding value.
- Obviously, the cloud pirates setup requires a different setup of the secrets, using the defaults we are talking about, right?
- How are these values transportet so that both the postgres as well as the edc image understand them?

Sorry, if my questions sound stupid, but helm is still something I am learning slowly and the mechanism for configuration used in the connector are also not an area, I was deeply involved with so far, especially the interaction with the vault, I have not looked into.

tom-rm-meyer-ISST: You're welcome.

Currently EDC does the following:

1. Search vault for `edc.datasource.{source_x}.*` property in vault, take if found.
2. Pickup value from ENV variable `EDC_DATASOURCE_{SOURCE_X}_*`

The changes made with this pr only affect the step 2.

AS IS / Problem:

1. The postgresql chart creates a secret for `auth` and `customUser`. The postgresql statefulset takes the values from there. It creates the postgres users.
2. The controlplane and dataplane deployment use the `postgresql.auth` variables to set the postgres user to the ENV variables. **The `customUser` is ignored**.

TARGET / Solution:

1. The EDC sets the ENV vars based on the `customUser` secrets from postgres (preferred). If not given, set them to the auth variables

As a result, an operator does the the following:

1. Store the values in the vault (if you use the variables picked up by the edc in step 1, the step 2 functionality won't be taken).
2. Configure your chart to use the same values as stored in the vault to configure the postgresql.

Maybe it's easier to have a 15 - 30 minutes call between 10 & 10.30 / 11 & 11.30 so that we can discuss the details and how to proceed?

charts/tractusx-connector/README.md
```diff
@@ -36,6 +36,64 @@ In addition, in order to map BPNs to DIDs, a new service is required, called the
...
+  auth:
+    username: "postgres"
+    password: "password"
+```
+
+When installing postgres with the chart, it is **RECOMMENDED** to use a custom user beside the admin user:
+
+> [!note]
+> The value `postgresql.jdbUrl` is composed in the deployments using a dependent variable of the database name and the name of the postgress deployment.
+
```
tom-rm-meyer-ISST, line 55: Secrets are updated

charts/tractusx-connector/README.md
```diff
@@ -36,6 +36,64 @@ In addition, in order to map BPNs to DIDs, a new service is required, called the
...
+    username: "postgres"
+    password: "password" # generated if empty
+  customUser: # name or exisitng secret must be set
+    name: "edc"
+    database: "edc" # must ge different than postgresql.auth.database
+    password: "password" # generated if empty
+```
+
+When installing postgres with the chart, you can alteratively not create and use a custom user (**NOT RECOMMENDED**).
+
```
tom-rm-meyer-ISST, line 69: secrets are updated

charts/tractusx-connector/README.md
```diff
@@ -36,6 +36,64 @@ In addition, in order to map BPNs to DIDs, a new service is required, called the
...
+    database: "postgres"
+    username: "postgres"
+    password: "password" # generated if empty
+  customUser:
+    existingSecret: "" # explicitly set to empty
+    name: "" # explicitly set to empty
+```
+
+Further you can also reuse existing secrets instead:
+
```
tom-rm-meyer-ISST, line 82: secrets are not updated

lgblaumeiser (approved):
lgtm

PR #2897, opened 2026-06-23
Fix: Make refresh use body as specified in rfc6749

### LINKED ISSUE #2896: Refresh Token mechanism not according spec
## Describe the bug

CX-0018 states, that refresh follows [RFC6749](https://www.rfc-editor.org/rfc/rfc6749#section-6) which states that the two parameters `grant_type` and `refresh_token` have to be provided as body in the content type `application/x-www-form-urlencoded`, the current implementation does add the two parameters to the query string, although it is a POST call.

### To Reproduce

Observe the tests https://github.com/eclipse-tractusx/tractusx-edc/blob/597867349f28a9c52953a14fee9b4f6d04e47835/edc-extensions/dataplane/dataplane-token-refresh/token-refresh-api/src/test/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/api/v1/TokenRefreshApiControllerTest.java to see the current setup

### Expected behavior

RFC is followed as expected

## Possible Implementation

As this is done today in existing connectors, the wrong way cannot be removed, instead the right way should be implemented and prioritized, i.e., if a body is given, the data is used, if not, the query params are used

lgblaumeiser:
## WHAT

The refresh mechanism used by Catena-X references to rfc6749 which states, that the parameter grant_type and refresh_token have to provided as formencoded values in the body. In the current implementation, this has been done using query parameters. The fix adds the use of the body and prioritizes the body if both is available. IN a later step, we can remove the query param based mechanism

## WHY

To have the implementation be consistent with the Catena-X spec and common practises in the internet. And because query parameter in a post request are criminal. The old way is kept because we need to interact with previous versions, so in both directions, we cannot assume that the counter party can handle the body. This can be removed in a later version.

Closes #2896

### DIFF
--- a/edc-extensions/dataplane/dataplane-token-refresh/token-refresh-api/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/api/v1/TokenRefreshApi.java
+++ b/edc-extensions/dataplane/dataplane-token-refresh/token-refresh-api/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/api/v1/TokenRefreshApi.java
@@ -21,7 +21,6 @@ package org.eclipse.tractusx.edc.dataplane.tokenrefresh.api.v1;
 
 import io.swagger.v3.oas.annotations.OpenAPIDefinition;
 import io.swagger.v3.oas.annotations.Operation;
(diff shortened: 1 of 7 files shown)


AndrYurk (approved):
lgtm

Issue #2899, opened 2026-06-24
Create Bugfix Release 0.12.2


lgblaumeiser:
This is for now an item to gather potential issues to be fixed in a bugfix release for version 0.12.0, Release branch is here: release/0.12.2

Items of relevance:

- Refresh Token Spec Issue
  - Issue: #2896
  - Main PR: #2897 
  - Bugfix PR: #2900 
  - Status: merged 

- Refresh Token Response Issue
  - Issue: #2936 
  - Main PR: #2937 
  - Bugfix PR: #2938 
  - Status: merged

- Null issue BDRS Client
  - Issue: #2915 
  - Main PR: n/a
  - Bugfix PR: #2934 
  - Status: merged 

- Prevent usage policies to be used as access policy
  - Issue: #2261 
  - Main PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3013
  - Bugfix PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/2993
  - Status: merged

- GET/POST /v3/edrs/request returns 500 when a cached EDR has no contractNegotiationId
  - Issue: #3001 
  - Main PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3003
  - Bugfix PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3027
  - Status: merged

- Make token refresh recoverable when the refresh response does not reach the consumer
  - Issue: #3002 
  - Main PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3004
  - Bugfix PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3026
  - Status: merged

- Allure test reporting in bugfix branch 0.12.x is broken
  - Issue: n/a
  - Main PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/2979
  - Bugfix PR: https://github.com/eclipse-tractusx/tractusx-edc/pull/3028
  - Status: merged

lgblaumeiser:
Release succesfully created

PR #2900, opened 2026-06-24
Fix: Refresh Token has to use request body as specified in rfc 6749

lgblaumeiser:
## WHAT

The refresh token endpoint follows RFC 6749, which states, that two properties called `grant_type` and `refresh_token` are transfered as form based parameters in the body of the post request. The current implementation makes use of query parameters instead of a body. The fix adds the body but keeps the query parameter as secondary mechanism to ensure interoperability with existing edc implementations.

## WHY

Several reasons:
- Adherence to the spec (in this case CX-0018) for future use of connectors as primary and at some time only mechanism
- Prepare to overcome non-conformant use of query parameters in a post request
- Bugfix, because it allows to remove the query parameter mechanism for future versions as soon as 0.12.x becomes the minimum used connector versions

Relates to: #2899


AndrYurk (approved):
lgtm

PR #2906, opened 2026-06-25
chore: Enable dependabot for bugfixes

lgblaumeiser:
## WHAT

A manual workflow to run dependabot. As the release branches change, the solution allows to run dependabot on a selected branch. It does not create a pull request, but the results can be easily used by Copilot to fix the issues and do a single pull request. This is sufficient and generally applicable to branches.

## WHY

To ensure updates of dependencies in the bugfix

### DIFF
--- a/.github/workflows/dependabot-on-demand.yaml
+++ b/.github/workflows/dependabot-on-demand.yaml
@@ -19,7 +19,7 @@
 
 ---
 name: "Dependabot (on-demand)"
-run-name: "Dependabot (gradle ecosystem) on ${{ inputs.branch }}"
+run-name: "Dependabot ${{ matrix.ecosystem }} (${{ matrix.directory }})"
 
 on:
   workflow_dispatch:
@@ -34,12 +34,20 @@ permissions:
 jobs:
   run:
     runs-on: ubuntu-latest
-    name: "Gradle dependency"
+    name: "Dependabot ${{ matrix.ecosystem }} (${{ matrix.directory }})"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - ecosystem: gradle
+            directory: /
+          - ecosystem: docker
+            directory: resources
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       LOCAL_GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      ECOSYSTEM: gradle
-      DIRECTORY: /
+      ECOSYSTEM: ${{ matrix.ecosystem }}
+      DIRECTORY: ${{ matrix.directory }}
       BRANCH: ${{ inputs.branch }}
       REPO: ${{ github.repository }}
     steps:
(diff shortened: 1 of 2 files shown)


bmg13 (approved):
In the future we could defined as target "release/*".

AndrYurk:
@lgblaumeiser, this solution requires making changes every time we need to run on a different release branch. I propose creating a separate workflow to run Dependabot checks for the input version. As an example, you can see [here](https://github.com/AndrYurk/tractusx-edc/actions/workflows/dependabot-on-demand.yml)
And the [output](https://github.com/AndrYurk/tractusx-edc/actions/runs/28363867043/job/84024823039)

lgblaumeiser:
> In the future we could defined as target "release/*".

As we keep all the release branches in place, this would be an issue. We would need to remove all branches after release. Actually, we do not need them anyway, because we use the tags for any relevant purpose.

lgblaumeiser:
@AndrYurk Is this closer to your suggestion 😃

.github/workflows/dependabot-on-demand.yaml
```diff
@@ -0,0 +1,50 @@
...
+#  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#  License for the specific language governing permissions and limitations
+#  under the License.
+#
+#  SPDX-License-Identifier: Apache-2.0
+#################################################################################
+
+---
+name: "Dependabot (on-demand)"
+run-name: "Dependabot (gradle ecosystem) on ${{ inputs.branch }}"
```
AndrYurk:
```suggestion
run-name: "Dependabot ${{ matrix.ecosystem }} (${{ matrix.directory }})"
```

.github/workflows/dependabot-on-demand.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    name: "Gradle dependency"
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      LOCAL_GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      ECOSYSTEM: gradle
+      DIRECTORY: /
```
AndrYurk:
```suggestion
      ECOSYSTEM: ${{ matrix.ecosystem }}
      DIRECTORY: ${{ matrix.directory }}
```

.github/workflows/dependabot-on-demand.yaml
```diff
@@ -0,0 +1,50 @@
...
+        description: "Branch to check (e.g. release/0.10.0)."
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    name: "Gradle dependency"
```
AndrYurk:
```suggestion
    name: "Dependabot ${{ matrix.ecosystem }} (${{ matrix.directory }})"
    strategy:
      fail-fast: false
      matrix:
        include:
          - ecosystem: gradle
            directory: /
          - ecosystem: docker
            directory: resources
```

AndrYurk (changes requested):
I think we need to add docker ecosystem check as well, please see my comments.

AndrYurk (approved):
lgtm

PR #2917, opened 2026-07-02
fix(agreements-bpns): null-check BdrsClient.resolveBpn result

### LINKED ISSUE #2915: agreements-bpns: missing null check on BdrsClient.resolveBpn()
## Summary

`EventContractNegotiationSubscriber` in the `agreements-bpns` extension does not null-check the return value of `BdrsClient.resolveBpn(did)`. When BDRS has no mapping for a provider or consumer DID, `resolveBpn` returns `null`, which is then passed straight into `AgreementsBpnsEntry.Builder`. It _should_ throw on `.build()`, however I haven't observed it does.

Downstream, `TxDataFlowPropertiesProvider` looks up the (missing) row on the DSP 2025 code path and fails the transfer with:

```
FATAL_ERROR - No BPN entry found for agreement <agreementId>
```

…which surfaces at transfer time, far from the actual root cause at negotiation-finalized time.

## Reproduction

1. Finalize a contract negotiation on the DSP 2025-1 (DID) code path where the counterparty's DID is not yet resolvable via BDRS (e.g. BDRS cache miss or truly missing entry).
2. Observe: no row in `edc_contract_agreement_bpns` for the agreement, and typically an unhandled exception from the subscriber in the control-plane log (not a `monitor.severe(...)` line).
3. Attempt a transfer against that agreement → `No BPN entry found for agreement <id>` FATAL_ERROR from `TxDataFlowPropertiesProvider`.

## Expected behavior

The subscriber should detect that a DID could not be resolved to a BPN and log a clear, actionable SEVERE message at the subscriber (naming the agreement id and the unresolvable DID), and

## Suggested fix

In `EventContractNegotiationSubscriber.extractBpn` (or its caller), check for `null` after `bdrsClient.resolveBpn(id)` and `monitor.severe(...)` including the agreement id and unresolvable DID. Do not attempt `AgreementsBpnsEntry.Builder.build()` with null fields.

arnoweiss:
## What
add null check on bdrs resolution

Closes #2915

### DIFF
--- a/edc-extensions/agreements-bpns/bpns-evaluation-core/src/main/java/org/eclipse/tractusx/edc/agreements/bpns/EventContractNegotiationSubscriber.java
+++ b/edc-extensions/agreements-bpns/bpns-evaluation-core/src/main/java/org/eclipse/tractusx/edc/agreements/bpns/EventContractNegotiationSubscriber.java
@@ -47,8 +47,12 @@ public class EventContractNegotiationSubscriber implements EventSubscriber {
         var agreement = payload.getContractAgreement();
         var agreementId = agreement.getId();
 
-        var providerBpn = extractBpn(agreement.getProviderId());
-        var consumerBpn = extractBpn(agreement.getConsumerId());
+        var providerBpn = extractBpn(agreementId, agreement.getProviderId());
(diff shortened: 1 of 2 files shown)


AndrYurk (approved):
lgtm

PR #2932, opened 2026-07-03
refactoring: Provide endpoints currently available under api version v4alpha to management api v3

### LINKED ISSUE #2914: Move all existing endpoints with a v4alpha prefix to management api v3
Situation, over the past months, we introduced additional management api endpoints and added then according to the upstream strategy as `/v4alpha/something`. In preparation of the adoption of management api v4 we should move the endpoints to `/v3/something` as all of them were designed according to the management api v3 logic and we should not invite these endpoints to v4 without a proper check, whether the semantics are right.

Therefore, this item proposes to move them to v3 as this was always the intention and to keep the v4alpha endpoints just for convenience to not break existing business apps. But the support of v4alpha should become a hidden feature, i.e., documentation should only present `/v3/something` as relevant.

We should do this in 0.13.0 version, because I want to prevent people from associating these endpoints with the real existing management api v4 from that version. How we adopt v4 should be a general discussion including how we move existing v3 addon endpoints from Tractus-X to the v4 api idea.

AndrYurk:
## WHAT

Move all existing endpoints with a v4alpha prefix to management api v3

## WHY

To prevent people from associating these endpoints with the real existing management api v4 from that version and to prepare the introduction of management version v4 in the upcoming connector release.

Closes #2914


edc-extensions/connector-discovery/connector-discovery-api/src/main/java/org/eclipse/tractusx/edc/discovery/api/v3/ConnectorDiscoveryV3Controller.java
```diff
@@ -0,0 +1,112 @@
+/*
```
lgblaumeiser, line 1: I suggest to bring the two controller into one implementation and have this class very thin by just pushing the request to the general controller?

edc-extensions/dataplane/dataflow/dataflow-api/src/main/java/org/eclipse/tractusx/edc/dataflow/api/v4alpha/DataFlowV4AlphaApiController.java
```diff
@@ -0,0 +1,60 @@
+/********************************************************************************
```
lgblaumeiser, line 1: Here as, well, could we remove the redundancy, by only providing one controller implementation and the version specific path controller just delegate to the common implementation?

lgblaumeiser (changes requested):
Just one thing about redundancy in code. Would appreciate to get the implementation reduced to one time

lgblaumeiser (approved):
lgtm

PR #2937, opened 2026-07-09
fix: Make token response rfc6749 compliant

### LINKED ISSUE #2936: Adapt Token Response to spec
[RFC6749](https://www.rfc-editor.org/rfc/rfc6749#section-6)

> I think there is a bit still missing which is the `TokenResponse`:
> 
> https://github.com/eclipse-tractusx/tractusx-edc/blame/72755ae7c3a38f38a483ff9b67b4caeafb86aa69/spi/tokenrefresh-spi/src/main/java/org/eclipse/tractusx/edc/spi/tokenrefresh/dataplane/model/TokenResponse.java#L24-L27
> 
> The there is a property called `expires` while according to the spec it should be called `expires_in` 

 _Originally posted by @wolf4ood in [#2896](https://github.com/eclipse-tractusx/tractusx-edc/issues/2896#issuecomment-4916361475)_

lgblaumeiser:
## WHAT

RFC 6749 (references in CX-0018) defines the content in a token refresh response to contain a field `expires_in`, in the current implementation, this field is called `expires`. The change is in a way, that the response now contains both fields to stay backward compatible. 

## WHY

The idea is, that hopefully, older implementations ignore additional fields in the token and make use of the legacy field, whereas the implementation returns the new field as default. I have no other idea on how to change that, as it is the old connector versions that might have an issue, this solution should be save, as it basically can handle both fields to be used.

Closes #2936


AndrYurk (approved):
lgtm

PR #2938, opened 2026-07-09
fix: Make token response rfc6749 compliant

lgblaumeiser:
## WHAT

RFC 6749 (references in CX-0018) defines the content in a token refresh response to contain a field expires_in, in the current implementation, this field is called expires. The change is in a way, that the response now contains both fields to stay backward compatible.

## WHY

The idea is, that hopefully, older implementations ignore additional fields in the token and make use of the legacy field, whereas the implementation returns the new field as default. I have no other idea on how to change that, as it is the old connector versions that might have an issue, this solution should be save, as it basically can handle both fields to be used.

## FURTHER NOTES

Relates #2936 // Fix for bugfix release 0.12.2


AndrYurk (approved):
lgtm

PR #2958, opened 2026-07-10
Feat/update the compatability tests

### LINKED ISSUE #2504: Compatibility tests for 0.11.x connector and later
## WHAT

Today, the compatibility tests are checking against the connector version 0.9.0, which is important, as this proves that we did not lose backward compatibility. But with the introduction of DSP version 2025-1 we actually also need a compatibility check on that dsp level, i.e., with an existing connector 0.11.0 or later in parallel.

## WHY

To ensure backward compatibility on DSP level 0.8 and 2025-1, as both DSP versions are relevant for the dataspace.

## HOW

This issue is on hold until the compatibility tests are actually merged into the edc repository, because this step is more important. Than, we have to come up with an own test suite that ensures the usage patterns of the new connector world (e.g. connector param discovery, ...) which is used for both DSP levels.

bmg13:
## WHAT

Updates the compatability tests to deprecate the 0.8 version of DSP, using the recent supported one of 2015-1.

## WHY

New 0.13 version is not meant to have support for the "old" 0.11 version.

## FURTHER NOTES

Legacy Participant is needed since the /v3 api path is not maintained in the most recent version, so this api versioning must be supported.

Hosting was slightly updated since we seem to be unable to run them in the runner with them.

Ran the compatibility tests successfully in [here](https://github.com/bmg13/tractusx-edc/actions/runs/29088686062/job/86348522042).

Closes #2504


edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/LegacyRemoteParticipant.java
```diff
@@ -0,0 +1,162 @@
+/*******************************************************************************
```
lgblaumeiser: Why did you create another inherited class, is there anything speaking against adding this to the RemoteParticipant?

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -173,12 +205,19 @@ void httpPullTransfer(TractusxDcpParticipantBase consumer, TractusxDcpParticipan
         provider.setProtocol(protocol);
         providerDataSource.stubFor(any(anyUrl()).willReturn(ok("data")));
         var assetId = UUID.randomUUID().toString();
-        var usagePolicy = inForceDatePolicyLegacy("gteq", "contractAgreement+0s", "lteq", "contractAgreement+5s");
+        var now = Instant.now().truncatedTo(ChronoUnit.SECONDS);
+        var usagePolicy = inForceDatePolicy("gteq", now.minusSeconds(60).toString(), "lteq", now.plusSeconds(20).toString());
```
lgblaumeiser: We should not use inForceDataPolicy anymore, as this is actually not part of the Catena-X Saturn based policy framework

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -207,11 +246,16 @@ void suspendAndResume_httpPull_dataTransfer(TractusxDcpParticipantBase consumer,
...
         var assetId = UUID.randomUUID().toString();
-        createResourcesOnProvider(provider, assetId, PolicyFixtures.noConstraintPolicy(), httpSourceDataAddress());
-
-        var transferProcessId = consumer.requestAssetFrom(assetId, provider)
-                .withTransferType("HttpData-PULL")
-                .execute();
+        createResourcesOnProvider(provider, assetId, noConstraintPolicy(), httpSourceDataAddress());
+
+        String transferProcessId;
+        if (consumer instanceof LegacyRemoteParticipant legacyConsumer) {
```
lgblaumeiser: How do they differ? We have a 0.12.1 and a current connector, why is their a difference depending on which of the two is provider?

edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/helpers/PolicyHelperFunctions.java
```diff
@@ -305,6 +305,21 @@ public static JsonObject frameworkConstraint(Map<String, Object> operandMappings
                 .build();
     }
 
+    public static JsonObject inForceDatePolicy(String operatorStart, Object startDate, String operatorEnd, Object endDate) {
```
lgblaumeiser: That was gone, because inForceDate is not relevant anymore as a policy constraint!

lgblaumeiser (commented):
Some questions

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
bmg13: I did think about that and decided to go with the legacy constraint.

My plan was that without check the time, the assertion
```java
 await().atMost(consumer.getTimeout())
                .untilAsserted(() -> assertThatThrownBy(() -> consumer.edrs().getEdr(transferProcessId)));
```

related with edr contract ending, would not run.

And now the way I see it, I removed the inForceDate, but had to reacess these assertions.

I thought on keeping as is to ensure assertion passing, but maybe I am missing something.

edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/helpers/PolicyHelperFunctions.java
bmg13: Replied this above.

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/LegacyRemoteParticipant.java
bmg13: I need to enforce a "/v3" and the most recent version, the "/v4" is enforce in the testing participant.

.github/workflows/verify.yaml
```diff
@@ -259,7 +259,7 @@ jobs:
         run: ./gradlew test -DincludeTags="PostgresqlIntegrationTest" -PverboseTest=true
 
   compatibility-tests:
-    if: false # Disabled while compatibility tests are being migrated to the updated EDC runtime.
+    if: true
```
lgblaumeiser: I think, we can simply run the action unconditionally, so I suggest to remove the line

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/LegacyRemoteParticipant.java
lgblaumeiser: But RemoteParticipant is a class in the same package, so I agree you have to switch the management api version, but why don't you add the overwritten baseManagementRequest there?

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -243,31 +277,176 @@ void suspendAndResume_httpPull_dataTransfer(TractusxDcpParticipantBase consumer,
...
+        createAssetManagementContext(provider, assetId, Map.of("description", "description"), dataAddressProperties);
+        var contractPolicyId = createPolicyDefinitionManagementContext(provider, contractPolicy);
+        var noConstraintPolicyId = createPolicyDefinitionManagementContext(provider, noConstraintPolicy());
 
-        createContractDefinitionLegacyManagementContext(provider, assetId, UUID.randomUUID().toString(), noConstraintPolicyId, contractPolicyId);
+        createContractDefinitionManagementContext(provider, assetId, UUID.randomUUID().toString(), noConstraintPolicyId, contractPolicyId);
     }
 
-    public String createContractDefinitionLegacyManagementContext(TractusxDcpParticipantBase participant, String assetId, String definitionId, String accessPolicyId, String contractPolicyId) {
+    private String executeLegacyTransfer(LegacyRemoteParticipant legacyConsumer, TractusxDcpParticipantBase provider, String assetId) {
```
lgblaumeiser: I do not think, that you need this method. If you would overwrite baseManagementRequest in RemoteParticipant, the polymorphism of Java ensures that this method is called. So, what is it, you try to get around with this method. Is it really, only the base path management?

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/LegacyRemoteParticipant.java
lgblaumeiser: The question is, whether we need this as well for the LocalParticipant, as we do not support v4 out of the box for the current state as well?

lgblaumeiser (changes requested):
There seems to be a misconception between the two of us. Let us discuss that in the weekly tomorrow

bmg13:
> There seems to be a misconception between the two of us. Let us discuss that in the weekly tomorrow

Discussed in the weekly meeting regarding needed approaches.
Now, they are addressed.

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/LegacyRemoteParticipant.java
bmg13: I understand. My original idea was to "prepare" for future compatibility api version deprecations, but I understand that it would not be the best approach atm. Changed.

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -234,41 +248,215 @@ void suspendAndResume_httpPull_dataTransfer(TractusxDcpParticipantBase consumer,
...
+                .jsonPath()
+                .getString("'@id'");
+    }
+
+    private String startTransferProcess(TractusxDcpParticipantBase consumer, TractusxDcpParticipantBase provider, String assetId) {
+        if (consumer instanceof RemoteParticipant legacyConsumer) {
+            return prepareAndExecuteTransferProcess(legacyConsumer, provider, assetId);
+        }
+
+        return consumer.requestAssetFrom(assetId, provider)
```
lgblaumeiser, line 245: Hi @bmg13. It is not exactly, what I proposed. The idea I was talking about, was, that instead of creating a prepareAndExecuteTransferProcess method in this file, you should "overwrite" the "requestAssetFrom" method in the class RemoteParticipant. This way, we would have a clean implementation here without any case separations because a call to "consumer.requestAssetFrom" would automatically call the original method for the LocalParticipant and the overwritten method for the RemoteParticipant.

It would still be two implementations, but both could be found in the Participant inheritance hierarchy, instead of looking at two very different places.

lgblaumeiser (changes requested):

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
bmg13: Okay, I understand this point, however looking into it, I do not think it would "look" good.

If I try to override the Participant 
```java
public RequestAsset requestAssetFrom(String assetId, Participant provider) {
    return new RequestAsset(assetId, provider);
}
```
into the RemoteParticipant, a ```RequestAsset``` must be instantiated, which would be impossible since it is a closed instation. 
Even so, that would not make sense since I "only" would have to ```override``` the ```negotiateContract``` and the ```initiateTransfer``` methods, based on Participant's "execute()" method.

```java
        public String execute() {
            var offer = getOfferForAsset(counterPart, assetId);
            var contractAgreementId = negotiateContract(counterPart, offer);
            var transferProcessId = initiateTransfer(counterPart, contractAgreementId, privateProperties, destination, transferType, callbacks);
            assertThat(transferProcessId).isNotNull();
            return transferProcessId;
        }
```

So, I can override those in ```RemoteParticipant``` and it would work.
But, in order to work, on those overriden methods the class check (```instanceof```) is still needed. 

The test class would invoke what you want, however the need for type validation would not disappear and relying only on polymorphism would be possible in the ```TransferEndToEndTest``` but not in the ```RemoteParticipant``` I think.

Do you think this is still better than the currently proposed solution?

lgblaumeiser: Ok, this is one of the original tests:
```
        consumer.setProtocol(protocol);
        provider.setProtocol(protocol);
        providerDataSource.stubFor(any(anyUrl()).willReturn(ok("data")));
        var assetId = UUID.randomUUID().toString();
        var usagePolicy = inForceDatePolicyLegacy("gteq", "contractAgreement+0s", "lteq", "contractAgreement+5s");
        createResourcesOnProvider(provider, assetId, usagePolicy, httpSourceDataAddress());

        var transferProcessId = consumer.requestAssetFrom(assetId, provider)
                .withTransferType("HttpData-PULL")
                .execute();

        consumer.awaitTransferToBeInState(transferProcessId, STARTED);

        var edr = await().atMost(consumer.getTimeout())
                .until(() -> consumer.edrs().getEdr(transferProcessId), Objects::nonNull);

        // Do the transfer
        var msg = UUID.randomUUID().toString();
        var data = consumer.data().pullData(edr, Map.of("message", msg));
        assertThat(data).isNotNull().isEqualTo("data");

        // checks that the EDR is gone once the contract expires
        await().atMost(consumer.getTimeout())
                .untilAsserted(() -> assertThatThrownBy(() -> consumer.edrs().getEdr(transferProcessId)));

        // checks that transfer fails
        await().atMost(consumer.getTimeout()).untilAsserted(() -> assertThatThrownBy(() -> consumer.data().pullData(edr, Map.of("message", msg))));

        providerDataSource.verify(getRequestedFor(urlPathEqualTo("/source")));
```

It is basically a script calling methods in the participants that do some dsp action. It does not care, which participant acts in which role. It worked in the old setup with DSP 0.8 and a consumer and provider being a 0.9.0 participant.

Now, the method implementations like `requestAssetFrom` have changed to use management api v4. That is the change, that is critical, am I right?

So, what I ask is, that you basically implement `requestAssetFrom` in RemoteParticipant, so that the Java polymorphism would simply use for the old connector your implementation from RemoteParticipant instead of the default one from Participant and everything should work fine. You could even switch to an older version of the upstream Participant class and just copy and paste the implementation from there into RemoteParticipant.

Sorry, but I just do not get your point, what do I miss here?

bmg13: So maybe I am missing the point, and tried to illustrate with a concrete change.
I updated the code in different branch to illustrate this and you can see some of those changes in [here](https://github.com/bmg13/tractusx-edc/commit/239a628aadd543dac86a260d549636ae7eefc005) and confirm passing tests in [here](https://github.com/bmg13/tractusx-edc/actions/runs/29748490093/job/88372477374).

In here, I added the 
```
    @Override
    public RequestAsset requestAssetFrom(String assetId, Participant provider) {
        return super.requestAssetFrom(assetId, provider);
    }
```

override to illustrate here, however it is not needed, since implicitly will be a RemoteParticipant (by java's polymorphism).

The other overrides I think are needed although the need to the inclusion of the typecasting is also there.
This solution encapsulates quite well the logic in the RemoteParticipant so I can see it and can persist in this pr, but just don't think it gives the upside you propose.

Again, I may be missing something.

lgblaumeiser: Why is the typecasting needed? I do not understand, why you have to touch the two tests at all. They are just calling a set of methods that do the actual management api call. All you have to ensure is, that they call the right implementation, for RemoteParticipant an implementation with management api v3 and for the LocalParticipant an implementation with the current Participant implementations of calling the management api.

Can you give me any evidence, that this is not possible, I do not understand the need for the type check you introduced. And sorry for the requestAssetFrom example, I did not get into the methods, so I did not see, that the requestAssetFrom example is a bad one.

bmg13: Oh, sorry @lgblaumeiser I missunderstood this. I was discussing a different thing (the methods) and not the correct type usage. Nevermind.
I updated the code to fully use polymorphism by overriding the specific needed-to-be-changed methods and works as expected.
Changed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/2958/changes/8b3b999e9fd2a762d1f4c2ef0a515a0f4078bfc7).

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/IdentityHubParticipant.java
```diff
@@ -70,12 +70,12 @@ public LazySupplier<URI> getSts() {
...
 
     public URI getResolutionApi() {
-        return credentialsApi.get();
+        return URI.create(credentialsApi.get().toString());
     }
 
     public String didFor(String participantId) {
         var didUri = didApi.get();
-        return "did:web:" + URLEncoder.encode(didUri.getHost() + ":" + didUri.getPort(), StandardCharsets.UTF_8) + ":" + participantId;
+        return "did:web:" + URLEncoder.encode("localhost" + ":" + didUri.getPort(), StandardCharsets.UTF_8) + ":" + participantId;
```
AndrYurk: What is the reason for this change?

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -234,41 +244,104 @@ void suspendAndResume_httpPull_dataTransfer(TractusxDcpParticipantBase consumer,
...
+                .body(requestBody)
+                .when()
+                .post("/policydefinitions")
+                .then()
+                .log().ifValidationFails()
+                .statusCode(200)
+                .extract().jsonPath().getString(ID);
+    }
+
+    private void createContractDefinitionLegacyManagementContext(TractusxDcpParticipantBase participant, String assetId, String definitionId, String accessPolicyId, String contractPolicyId) {
```
lgblaumeiser: With getting rid of the legacy stuff here, I think we are good to go. I approve, but please fix the naming

lgblaumeiser (approved):
lgtm, please remove Legacy prior to merging

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
bmg13: good point
removed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/2958/changes/7b4f0b2f149f47244d9789f594702fb0d3e46e15).

AndrYurk (commented):
Just small nit from my side.

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -164,64 +179,59 @@ static void beforeAll() {
...
+    }
+
+    private static void addAudienceMapping(TractusxDcpParticipantBase target) {
+        var bpnKey = "testing.edc.bdrs." + UUID.randomUUID().toString().substring(0, 8);
+        System.setProperty(bpnKey + ".key", target.getId());
+        System.setProperty(bpnKey + ".value", target.getDid());
+
+        var didKey = "testing.edc.bdrs." + UUID.randomUUID().toString().substring(0, 8);
+        System.setProperty(didKey + ".key", target.getDid());
+        System.setProperty(didKey + ".value", target.getDid());
```
AndrYurk: What is the idea behind this?

bmg13: Residual changes needed before, but not after updated changes.
Removed in [here](https://github.com/eclipse-tractusx/tractusx-edc/actions/runs/29933167665/job/88967839884?pr=2958).

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/IdentityHubParticipant.java
bmg13: Same as the other, a change assumed as needed, but the resolve is the same and the previous one is the correct one.
Reverted in [here](github.com/eclipse-tractusx/tractusx-edc/pull/2958/commits/301e77b1400fcff7876784cc0533dbf9a773a9b5)

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/RemoteParticipant.java
```diff
@@ -90,6 +158,7 @@ private Map<String, String> datasourceConfig(Config postgresqlConfig) {
             config.putAll(datasourceEnvironmentVariables(ds, postgresqlConfig));
         });
         config.put("org.eclipse.tractusx.edc.postgresql.migration.schema", postgresqlConfig.getString("tx.edc.postgresql.migration.schema"));
```
AndrYurk: This also looks redundant.

AndrYurk (commented):
Thanks @bmg13, two more small nits found.

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
```diff
@@ -164,64 +176,49 @@ static void beforeAll() {
         vault.storeSecret(LOCAL_PARTICIPANT.getPrivateKeyAlias(), LOCAL_PARTICIPANT.getPrivateKeyAsString());
         vault.storeSecret(LOCAL_PARTICIPANT.getFullKeyId(), LOCAL_PARTICIPANT.getPublicKeyAsString());
         vault.storeSecret("client_secret_alias", "clientSecret");
+
+        var remoteKey = "testing.edc.bdrs.remote-" + UUID.randomUUID().toString().substring(0, 8);
+        System.setProperty(remoteKey + ".key", LOCAL_PARTICIPANT.getId());
+        System.setProperty(remoteKey + ".value", LOCAL_PARTICIPANT.getDid());
```
AndrYurk: Is this also redundant?

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/fixtures/RemoteParticipant.java
bmg13: You are right, thank you!
Removed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/2958/changes/56393276fa06dd312365674606d4e6b9a180bf09).

edc-tests/compatibility-tests/src/test/java/org/eclipse/tractusx/edc/compatibility/tests/transfer/TransferEndToEndTest.java
bmg13: You are right, thank you!
Removed in [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/2958/changes/56393276fa06dd312365674606d4e6b9a180bf09).

AndrYurk (approved):
lgtm

PR #2960, opened 2026-07-15
feat: JSON-schema-based policy validation

### LINKED ISSUE #2980: Validation of policies based on JSON Schemas
# Overview

Instead of hardcoded policy validations, the validation of policy rules should be based on JSON Schemas which are generally used to define the conditions that have to be checked.

## What's the benefit?

Better configurability of policies

## Limitation:

- As this feature uses an upstream functionality, the replacement is only usable with management api v4 and above
- As the validation requires following the JSON Schema, the usage of expanded JSON-LD is not possible anymore, as it would fail to be detected as valid operand in the validation

ronjaquensel:
## WHAT

Adds JSON-schema-based policy validation for management API v4. All schemas (Catena-X policy schemas and DSP contract schema) are located in the policy module under `resources/schema`. All existing validators are left in place to keep support for management API v3. The JSON schema validator is registered additionally.

As the API object on which validation is performed is the `PolicyDefinition`, two validators are created: the first one for `PolicyDefinition`, which simply calls the second validator for the `PolicyDefinition's` `policy`, and the second one, which validates a `Policy` object against the JSON schemas. Validation on other fields of `PolicyDefinition` is not necessary here, as these are already covered by the standard `PolicyDefinition` validator from upstream.

## WHY

To make policy validation more flexible and maintainable.

## FURTHER NOTES

In the `policy-tests` module, a different version of the `json-schema-validator` library is transitively included through the imported test-fixtures. As exclusion of this transitive dependency does not seem to work (probably either due to `testFixtures` or because it's a nested transitive dependency), a `resolutionStrategy` to force the version from `libs.versions.toml` has been added. As it reads the version-to-force from `libs.versions.toml`, it should not create any additional maintenance effort when doing version updates.

## Issues

Relates to [cx-ex-data-sovereignty/#27](https://github.com/catenax-eV/cx-ex-data-sovereignty/issues/27)
Relates to [sig-release/#1564](https://github.com/eclipse-tractusx/sig-release/issues/1564)
Fixes: #2980


edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/CxPolicyExtension.java
```diff
@@ -374,6 +375,7 @@ public String name() {
     @Override
     public void prepare() {
         validatorRegistry.register(EDC_POLICY_DEFINITION_TYPE, CxPolicyDefinitionValidator.instance());
+        validatorRegistry.register("v4:PolicyDefinition", new CxJsonSchemaPolicyDefinitionValidator());
```
lgblaumeiser: Is there no constant. If not, please define one, this hard coded string is ugly.

edc-extensions/cx-policy/src/main/resources/jsonschema/cx-policy/affiliates-bpnl-constraint-schema.json
```diff
@@ -0,0 +1,40 @@
+{
```
lgblaumeiser, line 1: Concerning all those files, wouldn't it make sense to download them during build time from a link instead of having them in here as copy?

edc-extensions/cx-policy/src/testFixtures/java/org/eclipse/tractusx/edc/policy/cx/validator/PolicyBuilderFixturesV4.java
```diff
@@ -0,0 +1,115 @@
+/********************************************************************************
```
lgblaumeiser: What is different in the policy build for v3 and for v4? I thought, it is the endpoint called, but the policy payload, I would have expected to be the same

edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/helpers/PolicyHelperFunctionsV4.java
```diff
@@ -0,0 +1,295 @@
+/********************************************************************************
```
lgblaumeiser: Same question here, why do the policies differ between v3 and v4?

edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/participant/TransferParticipant.java
```diff
@@ -66,6 +73,28 @@ public Config getConfig() {
         )));
     }
 
+    public void createPolicyDefinitionAndExpectValidationFailure(JsonObject policy) {
```
lgblaumeiser: I would say, this is not a Participant method, a participant method could be
- a successful policy creation
- a raw result for a policy creation call
But failures must not be hidden in helpers, as this is test specific, only in policy tests needed and hiding the error could lead to false positives

edc-tests/e2e/policy-tests/src/test/java/org/eclipse/tractusx/edc/tests/policy/PolicyDefinitionV4EndToEndTest.java
```diff
@@ -0,0 +1,235 @@
+/********************************************************************************
```
lgblaumeiser: I am not too happy about this duplication, the tests seem to be the same as for v3, could that be integrated into one test implementation

lgblaumeiser (changes requested):
Some questions and some improvement potential

edc-extensions/cx-policy/src/main/resources/jsonschema/cx-policy/affiliates-bpnl-constraint-schema.json
ronjaquensel: I assume that could lead to issues if the link isn't reachable at the respective time. Also, like this it's in line with how it's done for other files (JSON-LD contexts, JSON schema files) both upstream and in the Tractus-X EDC.

lgblaumeiser: You are right, partially, the mechanism is there, that we do not have issues at runtime. Build time is a different story, we could even cache them somewhere. The issue with the other files, is, that the source is not always easily receivable, some of the files, e.g., are not even provided with the url they represent. But perhaps this is another issue.

lgblaumeiser (approved):
Lgtm, but @bmg13 or @AndrYurk I would appreciate a second opinion

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/validator/jsonschema/CxJsonSchemaPolicyValidator.java
```diff
@@ -0,0 +1,79 @@
...
+    public CxJsonSchemaPolicyValidator() {
+        this.objectMapper = JacksonJsonLd.createObjectMapper();
+        this.schemaRegistry = SchemaRegistry.withDialect(Dialects.getDraft201909(), builder -> builder
+                .schemaIdResolvers(schemaIdResolvers -> prefixMappings.forEach(schemaIdResolvers::mapPrefix))
+                .resourceLoaders(resourceLoaders -> resourceLoaders.add(IriResourceLoader.getInstance())));
+    }
+
+    @Override
+    public ValidationResult validate(JsonObject input) {
+        var schemaValidator = schemaRegistry.getSchema(SchemaLocation.of(CX_POLICY_SCHEMA));
```
AndrYurk: I think it would be better to do the creation of `schemaValidator` once by moving it to the constructor.

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/validator/jsonschema/CxJsonSchemaPolicyValidator.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static final String CX_POLICY_SCHEMA = CX_POLICY_SCHEMA_PREFIX + "/policy-schema.json";
+
+    private final SchemaRegistry schemaRegistry;
+    private final ObjectMapper objectMapper;
+
+    private final Map<String, String> prefixMappings = new HashMap<>() {
+        {
+            put(CX_POLICY_SCHEMA_PREFIX + "/schema", CX_POLICY_SCHEMA_LOCATION);
+            put(CX_POLICY_SCHEMA_PREFIX, CX_POLICY_SCHEMA_LOCATION);
```
AndrYurk:
```suggestion
    private static final Map<String, String> PREFIX_MAPPINGS = Map.of(
            CX_POLICY_SCHEMA_PREFIX + "/schema", CX_POLICY_SCHEMA_LOCATION,
            CX_POLICY_SCHEMA_PREFIX, CX_POLICY_SCHEMA_LOCATION,
            DSPACE_2025_SCHEMA_PREFIX, DSPACE_2025_SCHEMA_LOCATION);
```

edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/validator/jsonschema/CxJsonSchemaPolicyDefinitionValidator.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final CxJsonSchemaPolicyValidator policyValidator;
+
+    public CxJsonSchemaPolicyDefinitionValidator() {
+        this.policyValidator = new CxJsonSchemaPolicyValidator();
+    }
+
+    @Override
+    public ValidationResult validate(JsonObject input) {
+        var policy = input.getJsonObject(POLICY_ATTRIBUTE_NAME);
```
AndrYurk: `input.getJsonObject("policy")` could return null for an absent key and throw a `ClassCastException` when the value is an array, string, or number. Both end in an HTTP 500. Can you add a validation here, please?

AndrYurk (commented):
Please review my comments and resolve the merge conflicts.

AndrYurk (changes requested):

AndrYurk (approved):
lgtm

PR #3003, opened 2026-08-13
fix: Add the contractNegotiationId conditionally

### LINKED ISSUE #3001: GET/POST /v3/edrs/request returns 500 when a cached EDR has no contractNegotiationId
## Describe the bug

`JsonObjectFromEndpointDataReferenceEntryTransformer` adds `contractNegotiationId` unconditionally:
  ```
  // edc-extensions/edr/edr-api-v2/.../JsonObjectFromEndpointDataReferenceEntryTransformer.java:59
  .add(EDR_ENTRY_CONTRACT_NEGOTIATION_ID, entry.getContractNegotiationId())                                                                                                                                         
   ```                                                                                                                                                                                                                 
  That field is optional, but jakarta.json rejects null values:                                                                                                                                                     
                                                                                                                                                                                                                    
  `java.lang.NullPointerException: Value in JsonObjects name/value pair cannot be null`                                                                                                                               
                                                                                                                                                                                                                    
  Because the NPE is thrown rather than reported through `TransformerContext#reportProblem`, it bypasses the per-entry error handling in `BaseEdrCacheApiController#requestEdrEntries`:                                 
  ```                                                                                                                                                                                                                  
  .map(it -> transformerRegistry.transform(it, JsonObject.class))                                                                                                                                                   
  .peek(r -> r.onFailure(f -> monitor.warning(f.getFailureDetail())))   // never reached                                                                                                                            
  .filter(Result::succeeded)                                                                                                                                                                                        
   ```                                                                                                                                                                                                                 
  So it escapes the resource method as an HTTP 500 — and it fails the entire EDR listing, not just the affected entry.                                                                                              
                                                                                                                                                                                                                    
  **Why the field can be null:** 
The value is written by EDC's `EndpointDataReferenceStoreReceiver#handleTransferStarted`(edr-store-receiver), which resolves the negotiation indirectly — from the agreement, at transfer-start time — and         
  explicitly tolerates a miss:                                                                                                                                                                                      
                                                                                                                                                                                                                    
  ```
var contractNegotiationId = Optional.ofNullable(contractAgreementService.findNegotiation(transferStarted.getContractId()))                                                                                        
          .map(ContractNegotiation::getId)                                                                                                                                                                          
          .orElse(null);                                                                                                                                                                                            
  if (contractNegotiationId == null) {                                                                                                                                                                              
      monitor.debug("Contract Negotiation for transfer process %s not found. The EDR cached entry will not have an associated contract negotiation id");                                                            
  }                                                                                                                                                                                                                 
  // entry is saved anyway 
```                                                                                                                                                                                         
                                                                                                                                                                                                                    
  Nothing else objects either: `edc_edr_entry.contract_negotiation_id` is nullable (V0_0_1__Init_EdrIndex_Database_Schema.sql), `EndpointDataReferenceEntry.Builder#build()` requireNonNulls only `assetId`, `agreementId`, 
  `transferProcessId`, `providerId`, and `SqlEndpointDataReferenceEntryIndex#mapResultSet` reads the null back without complaint. The only trace is one DEBUG line at write time.  

### To Reproduce

  1. Insert (or produce) an edc_edr_entry row with contract_negotiation_id IS NULL.                                                                                                                                 
  2. Call POST /management/v3/edrs/request.                                                                                                                                                                         
  3. → 500, with the NPE above in the control-plane log.   
  
Because the query uses the default `QuerySpec` limit of 50 and generates no `ORDER BY` unless a sort field is supplied `SqlQueryStatement#parseSortField`, the returned window is heap-ordered — the endpoint fails intermittently at first and permanently once the row settles inside the window. In production this reads as random EDR-lookup failures that progressively become total.  

### Expected behavior

  An entry without a contract negotiation id is serialized with the property omitted; the endpoint returns 200. A single incomplete row must never fail the whole listing.          

## Possible Implementation

Add the property conditionally — omitting it is the correct JSON-LD representation of an absent optional value:                                                                                                   
```                                                                                                                                                                                                                    
 var builder = jsonFactory.createObjectBuilder()                                                                                                                                                                   
          .add(ID, entry.getId())                                                                                                                                                                                   
          .add(TYPE, EDR_ENTRY_TYPE)                                                                                                                                                                                
          .add(EDR_ENTRY_PROVIDER_ID, entry.getProviderId())                                                                                                                                                        
          .add(EDR_ENTRY_ASSET_ID, entry.getAssetId())                                                                                                                                                              
          .add(EDR_ENTRY_AGREEMENT_ID, entry.getAgreementId())                                                                                                                                                      
          .add(EDR_ENTRY_TRANSFER_PROCESS_ID, entry.getTransferProcessId())                                                                                                                                         
          .add(EDR_ENTRY_CREATED_AT, entry.getCreatedAt());                                                                                                                                                         
                                                                                                                                                                                                                    
  if (entry.getContractNegotiationId() != null) {                                                                                                                                                                   
      builder.add(EDR_ENTRY_CONTRACT_NEGOTIATION_ID, entry.getContractNegotiationId());                                                                                                                             
  }                                                                                                                                                                                                                 
                                                                                                                                                                                                                    
  return builder.build();         
```

AndrYurk:
## WHAT

Add the property conditionally — omitting it is the correct JSON-LD representation of an absent optional value

## WHY

Fix possible NPE

Closes #3001

### DIFF
--- a/edc-extensions/edr/edr-api-v2/src/main/java/org/eclipse/tractusx/edc/api/edr/transform/JsonObjectFromEndpointDataReferenceEntryTransformer.java
+++ b/edc-extensions/edr/edr-api-v2/src/main/java/org/eclipse/tractusx/edc/api/edr/transform/JsonObjectFromEndpointDataReferenceEntryTransformer.java
@@ -48,15 +48,21 @@ public class JsonObjectFromEndpointDataReferenceEntryTransformer extends Abstrac
 
     @Override
     public @Nullable JsonObject transform(@NotNull EndpointDataReferenceEntry entry, @NotNull TransformerContext context) {
-        return jsonFactory.createObjectBuilder()
+        var builder = jsonFactory.createObjectBuilder()
                 .add(ID, entry.getId())
                 .add(TYPE, EDR_ENTRY_TYPE)
                 .add(EDR_ENTRY_PROVIDER_ID, entry.getProviderId())
                 .add(EDR_ENTRY_ASSET_ID, entry.getAssetId())
                 .add(EDR_ENTRY_AGREEMENT_ID, entry.getAgreementId())
                 .add(EDR_ENTRY_TRANSFER_PROCESS_ID, entry.getTransferProcessId())
-                .add(EDR_ENTRY_CREATED_AT, entry.getCreatedAt())
-                .add(EDR_ENTRY_CONTRACT_NEGOTIATION_ID, entry.getContractNegotiationId())
-                .build();
+                .add(EDR_ENTRY_CREATED_AT, entry.getCreatedAt());
+
+        // the contract negotiation id is optional: it is not set when the negotiation cannot be
+        // resolved from the agreement when the transfer starts. jakarta.json rejects null values.
+        if (entry.getContractNegotiationId() != null) {
+            builder.add(EDR_ENTRY_CONTRACT_NEGOTIATION_ID, entry.getContractNegotiationId());
+        }
+
+        return builder.build();
     }
 }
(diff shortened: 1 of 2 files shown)


lgblaumeiser (approved):
lgtm

Issue #3002, opened 2026-08-10
Make token refresh recoverable when the refresh response does not reach the consumer


AndrYurk:
## WHAT

The provider data plane should retire a rotated refresh token only once the consumer has proven that it received the replacement, instead of the moment the replacement is issued. The proof of receipt is the consumer presenting the new refresh token — nothing but the consumer can produce it — so no additional message and no change to the [Refresh Token Grant Profile](https://github.com/eclipse-tractusx/tractusx-profiles/blob/main/tx/refresh/refresh.token.grant.profile.md) are required.

Behaviour, per call to `POST <refreshEndpoint>/token` with `grant_type=refresh_token`:

| Token presented by the consumer | Provider behaviour |
| --- | --- |
| the **current** refresh token | rotate as today, and record the presented token as the predecessor of the new one |
| the **superseded** refresh token, i.e. the one the last rotation replaced | return the current refresh token, the one the consumer failed to receive; do not rotate, do not change stored state |
| anything else | reject with HTTP 401, unchanged |

Requirements:

1. Alongside the current refresh token, the provider persists the token it replaced.
2. A repeated request with the superseded token changes no stored state and can be made any number of times.
3. Presenting the current refresh token retires the predecessor in the same request. At most two generations are live at any time.
4. The access token is issued fresh on both paths, so a consumer always receives a usable one however late it retries. Only the refresh token is held back.
5. No new configuration setting. The behaviour is not time-bounded and cannot be tuned or disabled — see WHY.
6. Existing vault entries remain readable and are treated as having no predecessor. No migration.
7. No change to the request or response format, so interoperability with connectors implementing the profile as published is preserved.

## WHY
Today a refresh is committed on the provider before the consumer can possibly have received it:
`DataPlaneTokenRefreshServiceImpl.refreshToken()` overwrites the vault entry with the new token and only then returns the response, while `RefreshTokenValidationRule` requires an exact match against that stored value. There is no state in which the previous token is still acceptable.

That is only correct if the response always reaches the consumer, which is not a property an HTTP call has. When it does not, the two sides diverge permanently: the provider considers the old token dead, and the consumer never learned the new one. Every subsequent refresh is answered with HTTP 401 (`Provided refresh token does not match the stored refresh token.`) for the remaining life of the transfer, and the only recovery is renegotiating the contract.

[...]

PR #3004, opened 2026-08-13
feat: make refresh token recoverable

AndrYurk:
## WHAT

Make token refresh recoverable when the refresh response does not reach the consumer

## WHY

The provider rotates the refresh token before the consumer can possibly have received it: the vault entry is overwritten and the previous token is rejected from then on. That is only safe if the response always arrives.  

The provider now retires a refresh token only once the consumer proves it received the replacement, by presenting it. Until then the superseded token is still accepted and returns the current one, so a lost    
  response costs a retry instead of the transfer.

Part of #3002

### DIFF
--- a/edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
+++ b/edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
@@ -82,6 +82,7 @@ import static org.eclipse.tractusx.edc.edr.spi.CoreConstants.EDR_PROPERTY_REFRES
 public class DataPlaneTokenRefreshServiceImpl implements DataPlaneTokenRefreshService, DataPlaneAccessTokenService {
     public static final String ACCESS_TOKEN_CLAIM = "token";
     public static final String TOKEN_ID_CLAIM = "jti";
+    private static final long SLOW_PHASE_THRESHOLD_MS = 5000;
     private final long tokenExpirySeconds;
     private final List<TokenValidationRule> authenticationTokenValidationRules;
     private final ParticipantContextSupplier participantContextSupplier;
@@ -150,7 +151,8 @@ public class DataPlaneTokenRefreshServiceImpl implements DataPlaneTokenRefreshSe
      *     <li>verify the token's signature</li>
      *     <li>assert {@code iss} and {@code sub} claims are identical</li>
      *     <li>assert the the token contains an {@code token} claim, and that the value is identical to the access token we have on record</li>
-     *     <li>assert that the {@code refreshToken} parameter is identical to the refresh token we have on record</li>
+     *     <li>assert that the {@code refreshToken} parameter is identical to the refresh token we have on record, or to
+     *     the one that record superseded</li>
      * </ul>
      *
      * @param refreshToken        The refresh token that was issued in the original/previous token request.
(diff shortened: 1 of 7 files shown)


edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/rules/RefreshTokenValidationRule.java
```diff
@@ -62,9 +68,20 @@ public Result<Void> checkRule(@NotNull ClaimToken accessToken, @Nullable Map<Str
...
-                .compose(rt -> incomingRefreshToken.equals(rt.refreshToken()) ?
-                        success() :
-                        failure("Provided refresh token does not match the stored refresh token."));
+                .compose(rt -> {
+                    if (incomingRefreshToken.equals(rt.refreshToken())) {
+                        return success();
+                    }
+                    if (incomingRefreshToken.equals(rt.previousRefreshToken())) {
+                        replayedToken = rt;
+                        return success();
```
bmg13, line 77: I would still log this scenario

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/rules/RefreshTokenValidationRule.java
bmg13, line 1: maybe update the RefreshTokenValidationRuleTest class as well

bmg13 (commented):
just small nits

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/rules/RefreshTokenValidationRule.java
AndrYurk: We have it [here](https://github.com/eclipse-tractusx/tractusx-edc/pull/3004/changes#diff-30da409d7cc35ddf35081dc060af5a736fd9d97ee154a940895e72482e6d118dR218-R220). Is it not enough?

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/rules/RefreshTokenValidationRule.java
AndrYurk: Good catch, will add.

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/rules/RefreshTokenValidationRule.java
bmg13: Yes

bmg13 (approved):

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
```diff
@@ -395,6 +412,24 @@ private Result<String> toJson(Object object) {
         }
     }
 
+    /**
+     * Executes the given action and records how long it took. Phases exceeding {@link #SLOW_PHASE_THRESHOLD_MS} are
+     * logged at WARNING so that a slow/stalled external dependency (DID resolution, Vault, database) can be identified
```
lgblaumeiser: Mismatch between documentation and implementation:
- Documentation "logged as WARNING" -> please also see the spelling issue at -> as
- Implementation: "monitor.debug"

lgblaumeiser (approved):
I already approve, but there is a small inconsistency, I would appreciate to be solved @AndrYurk

edc-extensions/dataplane/dataplane-token-refresh/token-refresh-core/src/main/java/org/eclipse/tractusx/edc/dataplane/tokenrefresh/core/DataPlaneTokenRefreshServiceImpl.java
AndrYurk: @lgblaumeiser resolved.

PR #2993, opened 2026-08-05
feat: add validation to Contract Definition Policies

brunoneiva02:
## WHAT

Add new validation step to Contract Definitions Policies

## WHY

Avoid invalid Contract Definitions

## FURTHER NOTES

Added two validators: one to Contract Definitions API and another to Policy Definitions API according to the decision record

Closes #2261

PR #2971, opened 2026-07-20
docs: add validate contract definition policies decision record

correiaafonso12:
## WHAT

Adds the Decision Record for the "Validate Contract Definitions Policies" feature

## WHY

Persist the decision and decide the approach

## FURTHER NOTES

Closes https://github.com/eclipse-tractusx/tractusx-edc/issues/2261

Issue #2261, opened 2025-09-18
Add validation to `contractdefinitions` API


arnoweiss:
## WHAT

There's rules as to which constraints are allowed in access policies and usage policies. This is enforced on the `policydefinitions` API by checking against the `action` property. It is however not forbidden to bind a valid access policy as `usagePolicy` in a contract definition. This should throw an error.

## WHY
tractusx-edc 0.11-RC2 fails quite ungracefully after a provider successfully bound an access policy (containing a `BusinessPartnerNumber` leftoperand) as `usagePolicy`. When the consumer triggers a negotiation, the provider-edc inevitably fails returning http 500 and the logs printing a high-level jetty error. This is because there's no handler registered [for the given scope](https://github.com/eclipse-tractusx/tractusx-edc/blob/4f57fd28f7cc8dc8c0384780185e83b65b4a8201/edc-extensions/cx-policy/src/main/java/org/eclipse/tractusx/edc/policy/cx/CxPolicyExtension.java#L307).

## HOW

Should fit in nicely with the [other validator for that API](https://github.com/eclipse-tractusx/tractusx-edc/tree/main/edc-extensions/validators).

lgblaumeiser:
Hi @correiaafonso12, do we really have a problem here? We have the following types of policies:

- Old usage policies: indicator "action:use" and old context
- Old access policies: indicator "action:use" and old context. Note: not distinguishable from an old usage policy
- New usage policies: indicator "action:use" and new context
- New access policies: indicator "action:access" and new context

But, we do not support old access policies anymore, i.e., we can easily prevent the use of an access policy which does not have "action:access". The only thing that might happen is, that we accidentally use an old "access policy" as contract/usage policy. This is nothing we can prevent, as earlier policies did not have anything to distinguish their purpose, but for me, having this general check would be good enough for now, as the old stuff will be gone at some point anyway.

correiaafonso12:
The introduction of this validation declares that a valid Contract Definition MUST have an access policy (`action : access`) as accessPolicy, and an usage policy (`action : use`) as contractPolicy. Therefore, Contract Definitions that do not follow this rule are in an invalid state.

Valid Contract Definitions can become invalid if the referenced Policy Definitions change (e.g. via PUT request, changing the action of the policy), and, IMO, as part of this feature this should be properly handled (e.g. by restricting updates to Policy Definitions that are being referenced by a Contract Definition).

[...]

PR #3013, opened 2026-08-18
feat: include validation to Contract Definition Policies to version 0.13.0

bmg13:
## WHAT

Add new validation step to Contract Definitions Policies

## WHY

Avoid invalid Contract Definitions

## FURTHER NOTES

Added two validators: one to Contract Definitions API and another to Policy Definitions API according to the decision record:https://github.com/eclipse-tractusx/tractusx-edc/pull/2971.

This is the update of main with existing approved logic present in upcoming version 0.12.2: https://github.com/eclipse-tractusx/tractusx-edc/pull/2993

Relates to issue #2261


lgblaumeiser (approved):
lgtm

PR #3023, opened 2026-08-21
fix: kafka and dcp flaky tests

AndrYurk:
## WHAT

Fix Kafka and DCP flaky tests

## WHY

It breaks the e2e tests after merging into main, so we can't publish a new snapshot to Maven.


lgblaumeiser (approved):
lgtm

PR #3024, opened 2026-08-24
feat: update openApiSpecs

AndrYurk:
## WHAT

Update openApiSpecs

## WHY

TRG 1.08

Part of #2978


lgblaumeiser:
There is a hen and egg issue here, as long as we have not done the release, this link points into void. Actually, it is a config file for tooling within Tractus-X to copy the open API specs from our publication place to the OpenAPI Hub. It always takes main, so we need to release first to merge this PR to main, we do not need the change in the release branch.

lgblaumeiser (approved):
lgtm

PR #3026, opened 2026-08-24
feat: make refresh token recoverable

### LINKED ISSUE #3002: Make token refresh recoverable when the refresh response does not reach the consumer

AndrYurk:
## WHAT

Make token refresh recoverable when the refresh response does not reach the consumer

## WHY

The provider rotates the refresh token before the consumer can possibly have received it: the vault entry is overwritten and the previous token is rejected from then on. That is only safe if the response always arrives.

The provider now retires a refresh token only once the consumer proves it received the replacement, by presenting it. Until then the superseded token is still accepted and returns the current one, so a lost
response costs a retry instead of the transfer.

Closes #3002

### DIFF
--- a/docs/development/decision-records/2026-08-10-refresh-token-retire-on-acknowledgement/README.md
+++ b/docs/development/decision-records/2026-08-10-refresh-token-retire-on-acknowledgement/README.md
@@ -1,0 +1,106 @@
+# Retire Refresh Tokens on Acknowledgement
+
+## Decision
+
+The provider data plane will retire a rotated refresh token only once the consumer has proven that it received the
+replacement. The proof of receipt is the consumer presenting the new refresh token. Until that happens, the token it
+replaced remains acceptable, and presenting it again returns the current refresh token — the one the consumer failed to
+receive — instead of rotating a second time. The access token is issued fresh on both paths.
+
+This is a behavioural change with no configuration surface and no way to opt out. There is nothing worth switching off:
+the old behaviour is the one that breaks transfers on a network timeout, and the new one costs one extra string in the
+vault record. The request and response formats are unchanged, so conforming consumers see no difference either way.
+
+## Rationale
+
+Token refresh as specified in the
+[Tractus-X Refresh Token Grant Profile](https://github.com/eclipse-tractusx/tractusx-profiles/blob/main/tx/refresh/refresh.token.grant.profile.md)
+and implemented per [2024-03-05_token_refresh](../2024-03-05_token_refresh/README.md) rotates the refresh token on every
+successful call. The implementation treated the moment of *issuing* a new token as the moment the old one dies: the
+vault entry was overwritten and the previous token was rejected from then on.
+
+That is only correct if the response always reaches the consumer, which is not a property an HTTP call has. Observed in
+production: a refresh takes longer than a proxy or ingress in front of the public API tolerates, the request is answered
+with HTTP 504, or the consumer cancels and the access log shows 499. The provider has rotated; the consumer still holds
+the old refresh token; it never saw the new one. Every subsequent refresh is answered with HTTP 401 because the token
+presented no longer matches the stored one. The transfer is permanently broken and can only be recovered by
+restarting the transfer process — for a failure mode that is a plain network timeout.
+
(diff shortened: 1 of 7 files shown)


lgblaumeiser (approved):
lgtm

PR #3028, opened 2026-08-24
feat: update allure action to avoid simple-elf action

AndrYurk:
## WHAT

Update allure action to avoid simple-elf action

## WHY

The Docker image that simple-elf/allure-report-action runs in no longer provides xargs


lgblaumeiser (approved):
lgtm

PR #3029, opened 2026-08-25
fix : enforcedPlatform is not allowed for published component

AndrYurk:
## WHAT

Fix the problem with publishing artifacts

## WHY

To make publishing possible


lgblaumeiser (approved):
lgtm

PR #3035, opened 2026-08-27
feat: update migration guide for 0.13.0

AndrYurk:
## WHAT

Update migration guide for 0.13.0

## WHY

Bring more clarity for end users


lgblaumeiser (changes requested):
One thing, I suggest to add is the refactoring: Provide endpoints currently available under api version v4alpha to management api v3 issue, that is not needed, but I would like to see this recommended as change.

Thanks, btw, great work so far!

lgblaumeiser (approved):
lgtm

PR #3045, opened 2026-09-02
fix(control-plane-transform): render multi-valued right operands as a list

pathaniaaviX:
## WHAT

Carries a copy of `org.eclipse.edc:control-plane-transform` as `edc-extensions/control-plane-transform`, in which `JsonObjectFromPolicyTransformer` renders a multi-valued ODRL right operand as a JSON-LD array instead of the Java `toString()` of the connector's internal literals.

Every other class in the module is verbatim upstream (EDC 0.17.0, matching the current `edc` pin). Only the license headers were rewritten to the Tractus-X format, as was done for the `transform-lib` copy in 0.12.1.

Draft, because the route is still open for discussion — see the last section.

## WHY

A `BusinessPartnerNumber` access policy with two or more BPNs is accepted by the management API (`200`) but cannot be read back:

```
"odrl:rightOperand": "[{@value={chars=BPNL00000000AAAA, string=BPNL00000000AAAA, valueType=STRING}}, {@value={chars=BPNL00000000BBBB, ...}}]"
```

CX-0152's [BusinessPartnerNumber constraint schema](https://w3id.org/catenax/2025/9/policy/business-partner-number-constraint-schema.json) permits only `isAnyOf`/`isNoneOf` over an array for this left operand, and the access-policy schema permits only `and` as a logical constraint — `odrl:or` of two `eq` constraints is rejected by the connector with `400 Policy includes not allowed logical constraints`. So the array form is the only conformant way to grant access to several partners, and it is the form that cannot be read back.

Beyond display, this is a round-trip hazard: a client that reads a policy, adds a BPN and writes it back stores the blob as a single string literal, at which point evaluation genuinely fails.

### Why not just wait for the upstream fix

eclipse-edc/Connector#5934 ("fix: rightOperand serialization for list types", merged 4 Aug 2026) addresses the same method, and is not yet in any release — it is in neither 0.17.0 nor 0.18.0.

It is also not sufficient on its own. Its tests cover `List<String>` and `List<Integer>`, but `JsonObjectToConstraintTransformer.extractComplexValue` does not expand a multi-element array, so the collection holds Json-Ld literals rather than strings. Running the upstream fix against the values a connector actually stores:

| collection holds | upstream #5934 renders |
|---|---|
| `List.of("BPNL…", "BPNL…")` | `[{"@value":"BPNL…"}, {"@value":"BPNL…"}]` — correct |
| Json-Ld literals (what `extractComplexValue` keeps) | `[{"@value":"{\"@value\":\"BPNL…\"}"}, …]` |
| the map those deserialize to once persisted | `[{"@value":"{@value={string=BPNL…, chars=BPNL…}}"}, …]` |

So the released fix will turn one large blob into several small ones. Its end-to-end test asserts only `hasSize(2)`, which passes either way. The change here adds the unwrapping that is missing.

## FURTHER NOTES

The to-transformer is deliberately **not** touched. `BusinessPartnerNumberPermissionLegacyFunction.checkListContains` reads exactly the shape currently stored (`entry -> get("@value") -> get("string")`), so unwrapping on the write side would break BPN evaluation — turning a read-back defect into an access-control outage. Stored policies are unchanged and no migration is needed; this only affects rendering.

Scalar right operands render exactly as before, pinned by a test.

`edc-runtime-memory` needed the exclusion too. A `configurations.all { exclude }` in `edc-controlplane-base` does not propagate to its consumers, so without it that runtime resolves both the copy and the upstream module. The `transform-lib` copy in 0.12.1 has the same gap.

### Verification

- 18 tests in the new module: the 14 upstream `JsonObjectFromPolicyTransformerTest` cases unchanged, plus 4 covering the shapes above. The three behavioural ones fail without the patch.
- Whole project: `compileJava compileTestJava` and `checkstyleMain` clean; `test` green — 967 tests across 49 modules.
- Runtime classpaths of all three control-plane runtimes resolve the copy with no duplicate of the upstream module.
- E2E `policy-tests` green. `management-tests` and `catalog-tests` each have one failure — `DelegatedAuthEndToEndTest` (Keycloak container) and `CatalogTest` (BPN group validation) — both of which reproduce identically on an unmodified tree, so they are unrelated to this change. Worth a look independently: `CatalogTest` is red on `main` locally.

### On the route

The defect is upstream, and this is a copied module — not a small change to carry. If you would rather wait for an EDC release, the note above still matters: #5934 alone will not fix it, and a follow-up would be needed upstream first. Happy to take it that way instead, or to close this if you would prefer to track it purely on #2221.

Relates to #2221


lgblaumeiser:
HI @pathaniaaviX , thank you for the proposal. As you know, Andrii is gone for some more days. It will take a bit of time, until we can discuss this proposal in detail. So please be patient.

PR #3053, opened 2026-09-07
feat(helm): add Gateway API HTTPRoute

### LINKED ISSUE #2901: Switch from Nginx Ingress to Gateway API
## WHAT

Nginx is not maintained as ingress tooling anymore. The helm charts should be adapted to use the new Gateway API mechanism.

## WHY

Solution favored by Tractus-X

## HOW

Do it as already done in other Tractus-X Repos

InRiPa:
## WHAT

Adds Gateway API HTTPRoute support for the EDC control plane and data plane to the tractusx-connector Helm charts.

- `templates/httproute-controlplane.yaml`: new HTTPRoute template, functioning like the existing ingress-controlplane.yaml
- `templates/httproute-dataplane.yaml`: new HTTPRoute template, functioning like the existing ingress-dataplane.yaml
- `values.yaml`: new controlplane.httpRoutes and dataplane.httpRoutes blocks, disabled by default, mirroring the layout of the existing ingress blocks
- `README.md`: regenerated with helm-docs to document the new parameters

## WHY

Nginx is not maintained as ingress tooling anymore. The helm charts should therefore be adapted to use the new Gateway API mechanism. The main benefit would be, that with Gateway API, any modern Ingress Controller is supported and could be chosen based on the user and environment. 

**Note** that in this version, TLS and wildcard certificates must be configured and maintained directly on the Gateway — the HTTPRoute templates do not yet cover TLS sections. ListenerSet support will follow as soon as Traefik implements it, see https://github.com/traefik/traefik/issues/12626. Kubernetes and cert-manager already support it.

## Example - Controlplane:

```yaml
ingresses:
    ## Public / Internet facing Ingress
  - enabled: false
  [...]
## HTTPRoute declaration to expose the network service via Gateway API.                                              
## Mirrors the ingresses list above -- each entry creates an HTTPRoute                                               
## attached to the specified parentRefs (Gateway resources).                                                         
httpRoutes:                                                                                                          
  ## Public / Internet facing HTTPRoute                                                                              
- enabled: true                                                                                                      
    # -- The hostname to be used to precisely map incoming traffic onto the underlying network service               
  hostname: "edc-1-dsp.example.org"                                                    
    # -- Additional HTTPRoute annotations to add                                                                     
  annotations: {}                                                                                                    
    # -- EDC endpoints exposed by this HTTPRoute resource                                                            
  endpoints:                                                                                                         
  - protocol                                                                                                         
    # -- [parentRefs](https://gateway-api.sigs.k8s.io/api-types/httproute/#parentrefs) referencing Gateway resources 
  parentRefs: &id001                                                                                                 
  - name: traefik-gateway                                                                                            
    namespace: traefik                                                                                               
    sectionName: web                                                                                                 
  - name: traefik-https                                                                                              
    namespace: traefik                                                                                               
    sectionName: websecure-test                                                                                      
- enabled: true                                                                                                      
  hostname: "edc-1-mgt.example.org"                                                    
  annotations: {}                                                                                                    
  endpoints:                                                                                                         
  - management                                                                                                       
  - control                                                                                                          
  parentRefs: *id001            
    ## alternatively write it out for each rule                                                                                     
    # - name: gateway                                                                                                
    #   namespace: traefik                                                                                           
    #   sectionName: https                                                                                           

```

## Example - Dataplane:

```yaml
ingresses:                            
  ## Public / Internet facing Ingress 
- enabled: false                      
[...]
## HTTPRoute declaration to expose the network service via Gateway API.                                              
## Mirrors the ingresses list above -- each entry creates an HTTPRoute                                               
## attached to the specified parentRefs (Gateway resources).                                                         
httpRoutes:                                                                                                          
  ## Public / Internet facing HTTPRoute                                                                              
- enabled: true                                                                                                      
    # -- The hostname to be used to precisely map incoming traffic onto the underlying network service               
  hostname: "edc-1-dataplane.example.org"                                              
    # -- Additional HTTPRoute annotations to add                                                                     
  annotations: {}                                                                                                    
    # -- EDC endpoints exposed by this HTTPRoute resource                                                            
  endpoints:                                                                                                         
  - public                                                                                                           
    # -- [parentRefs](https://gateway-api.sigs.k8s.io/api-types/httproute/#parentrefs) referencing Gateway resources 
  parentRefs: *id001
    ## alternatively write it out for each rule                                                                                                 
    # - name: gateway                                                                                                
    #   namespace: traefik                                                                                           
    #   sectionName: https                                                                                           

```

## Just for completion - example for Gateway config

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: traefik-https
  namespace: traefik
  annotations:
    cert-manager.io/cluster-issuer: <my_cluster_issuer_via_cert_manager>
spec:
  gatewayClassName: traefik
  listeners:
    # Wildcard listener
    - name: websecure
      protocol: HTTPS
      port: 8443
      hostname: "*.example.org"
      allowedRoutes:
        namespaces:
          from: All
      tls:
        mode: Terminate
        certificateRefs:
          - name: example-org-gateway-tls
```

## FURTHER NOTES

- The change is confined to charts/tractusx-connector; no other runtime code paths changed.
- The new blocks are fully optional and disabled by default — existing Ingress-based deployments are unaffected.

Closes #2901


charts/tractusx-connector/templates/httproute-controlplane.yaml
```diff
@@ -0,0 +1,64 @@
+#################################################################################
+#  Copyright (c) 2025 Contributors to the Eclipse Foundation
```
lgblaumeiser: I would appreciate if you would not use that in EDC. It is actually a shortcut that adds more issues than it solves. Please state the correct owner of IP created by you so that we can trace that back if we need that in the future, e.g. in case of questions concerning legal relevant changes that need an approval by all IP holders. Thank you!

charts/tractusx-connector/templates/httproute-dataplane.yaml
```diff
@@ -0,0 +1,64 @@
+#################################################################################
+#  Copyright (c) 2025 Contributors to the Eclipse Foundation
```
lgblaumeiser: Same here, please use the correct IP owner

lgblaumeiser (changes requested):
As I am not a helm hero, actually more a helm loser, I cannot really review the content, still I have two formal change proposals which I would appreciate to be considered.

lgblaumeiser (commented):
Thanks for working on my comments. We will look into this and the BDRS PR in the next days.

MDSBarbosa (approved):

lgblaumeiser (approved):
lgtm

Merge pr, because test failure is not due to this change, but a change in availability of dependencies

Issue #3046, opened 2026-09-03
EDC build fails because of unused imports


InRiPa:
## Describe the bug

`./gradlew build` fails on a on a fresh clone (confirmed on main@77a64e803) with two Checkstyle `UnusedImports` errors in the e2e-fixtures module — imports that no CI task ever checks.

To Reproduce

1. Clone upstream main (defect present since at least the 0.13.0 release).
2. Run `./gradlew build --console=plain` 
3. See `:edc-tests:e2e-fixtures:checkstyleTestFixtures FAILED` with 2 UnusedImports errors.

### Expected behavior

`./gradlew build` / `check` on main  succeeds, and CI's checkstyle step should cover **all source sets** it can reach.

### Screenshots/Error Messages

```bash
> Task :edc-extensions:edr:edr-callback:checkstyleTest

[Incubating] Problems report is available at: file:///path/to/tractusx-edc/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':edc-tests:e2e-fixtures:checkstyleTestFixtures' (registered by plugin class 'org.gradle.api.plugins.quality.CheckstylePlugin').
> A failure occurred while executing org.gradle.api.plugins.quality.internal.CheckstyleInvoker
   > Checkstyle rule violations were found. See the report at: file:///path/to/tractusx-edc/edc-tests/e2e-fixtures/build/reports/checkstyle/testFixtures.xml
     Checkstyle files with violations: 1
     Checkstyle violations by severity: [error:2]
```

## Context Information

- Used version: main, 2026-09-03
- OS: Linux
- java --version: OpenJDK 21.0.12.1
- Gradle wrapper 9.7.1

## Possible Implementation

1. Drop the two unused imports in edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/transfer/ProviderPushBaseTest.java (lines 50/53).
```diff
--- a/edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/transfer/ProviderPushBaseTest.java
+++ b/edc-tests/e2e-fixtures/src/testFixtures/java/org/eclipse/tractusx/edc/tests/transfer/ProviderPushBaseTest.java
@@ -47,10 +47,8 @@ import static org.awaitility.Awaitility.await;
[...]

PR #3054, opened 2026-09-07
ci: run checkstyleTestFixtures in verify workflow

InRiPa:
## WHAT

- Closes the Checkstyle blind spot for the testFixtures source set: "Run Checkstyle"
  invoked only `checkstyleMain` `checkstyleTest`.

## WHY

- The source set exists repo-wide (added in EDC 0.12.0 via java-test-fixtures) and is
  executed by every local `./gradlew build` through the static check lifecycle — but was
  unreachable in CI, whose hardcoded task list predates it. 
 That results in new clones using `./gradlew build` failing while CI stayed green.

## FURTHER NOTES

An alternative would be to switch this step to `./gradlew check -x test`, which skips test execution
  but covers all source sets. This way future sets can't be forgotten. 
  However, I'm not fully sure and it might produce other anomalies. Maybe a discussion for the future. 

Complements #3046

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -69,7 +69,7 @@ jobs:
       - uses: ./.github/actions/setup-java
       - name: Run Checkstyle
         run: |
-          ./gradlew checkstyleMain checkstyleTest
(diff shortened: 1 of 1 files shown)


lgblaumeiser (approved):
lgtm

